PR #45, opened 2021-02-19
Fix zip file name in "Getting Started" example

jayschwa:


jart (approved):

jart:
Thanks for spotting this!

PR #66, opened 2021-02-27
Support %lf and %lF in format string

### LINKED ISSUE #58: printf doubles with %lf gives nan
Hi, I tried out a bunch of portable executables in Linux and Windows, and they all worked perfectly :).

However, when I compiled and ran the below code:

```c
int main() {
    double x = 0.25;
    double y = sin(x);
    printf("sin(%f) = %f\n", x, y);
    printf("sin(%lf) = %lf\n", x, y);
    return 0;
}
```

The output is:

```
sin(0.250000) = 0.247404
sin(0.000000) = nan
```

When I compile using `gcc -lm` the output is same for both.
Do I have to add a compilation flag to print doubles via `%lf`?

alisonatwork:
Per C99 this should resolve to double.

This should resolve #58

### DIFF
--- a/libc/fmt/palandprintf.c
+++ b/libc/fmt/palandprintf.c
@@ -215,6 +215,10 @@ hidden int palandprintf(void *fn, void *arg, const char *format, va_list va) {
         signbit = sizeof(intmax_t) * 8 - 1;
         break;
       case 'l':
+        if (format[1] == 'f' || format[1] == 'F') {
+          format++;
+          break;
+        }
         if (format[1] == 'l') format++;
         /* fallthrough */
       case 't': /* ptrdiff_t */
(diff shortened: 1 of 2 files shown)


jart:
Thanks for your contribution!

alisonatwork:
Thanks for merging, although I think your (vanished?) commit in 207db7d54f6967223b6378fdedbf814d209fa39b was a better fix than this one, since it also avoids strange behavior if someone erroneously uses `%tf` or `%zf` or something.

jart:
Yeah I pushed that hastily before I noticed this change. Here's what I say: porque no los dos? https://www.youtube.com/watch?v=OawrlVoQqSs

PR #67, opened 2021-02-27
redbean.c: remove duplicate png content-types

miniupnp:
Hello. The duplicate `{"png", "image/png"}` entries look useless. That won't change much regarding performance because a binary search is used on this array, but it eats 48 bytes for nothing ; unless you noticed a performance or code compression improvement with the duplication.


jart:
That certainly wasn't intended. Thank you for noticing and sending this change!

PR #89, opened 2021-03-01
Make it easy to build in RAM

matheusmoreira:
I understand the build tree is rooted at the `o` directory. Is this correct?

If so, that would make it easy to build in RAM:

    ln -s /tmp/cosmopolitan/o o

The source code will remain on disk while all build artifacts will be written to `tmpfs`. Due to the Linux page cache, all source code will eventually end up being cached in memory as well. Very fast!

I added some GNU Make logic to compute whether `o` is a symbolic link and ensure the link target directory exists. Fixed the build errors when I tried this in my own projects.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,6 +70,15 @@ GNUMAKEFLAGS += --output-sync
 .FEATURES: output-sync
 .PHONY: all o bins check test depend tags
 
+# If the output directory is a symbolic link, make sure the target directory exists.
+# This makes it easy to build on tmpfs:
+#
+#     ln -s /tmp/cosmopolitan/o o
+#     ln -s $XDG_RUNTIME_DIR/cosmopolitan/o o
+#
+output_directory_link := $(shell readlink o)
+$(if $(output_directory_link),$(shell mkdir -p $(output_directory_link)))
+
 all:	o
 o:	o/$(MODE)/ape		\
 	o/$(MODE)/dsp		\


jart:
Assuming the file system is fully cached in memory, how much does this improve build latency?

lourkeur:
This assumes that /tmp is a tmpfs, which isn't always the case. The [FHS standard](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s18.html) does not specify it has to be, it't all up to how the system is configured by the admin. (especially on Linux) In my case, it's usually not a tmpfs.

jart:
I tried these commands. They brought my build latency down from ~18.4 seconds to ~17.9 seconds.

```
sudo mount -t tmpfs -o size=8G,noatime,nodiratime /dev/shm o
time make -j18
```

It was a good idea for an experiment. Now we know that TmpFS requires root permissions and it offers a 3% performance boost. So I feel like it's not worth using. Builds already run in RAM since the kernel stores hard disk pages in memory. Thank you for sending this!

matheusmoreira:
My measurements:

```bash
$ git checkout tmpfs-build

$ time make

real	3m47.331s
user	2m37.935s
sys	0m59.821s

$ rm -rf ./o
$ ln -s /tmp/cosmopolitan/o o
$ time make

real	3m29.201s
user	2m33.335s
sys	0m56.464s

$ rm -rf ./o/*
$ rm -rf ./o
$ time make -j4

real	0m57.591s
user	2m36.238s
sys	0m51.658s

$ rm -rf ./o
$ ln -s /tmp/cosmopolitan/o o
$ time make -j4

real	0m53.347s
user	2m32.769s
sys	0m48.975s
```

Similar results for `make -j$N` for any `N`.

matheusmoreira:
> This assumes that /tmp is a tmpfs, which isn't always the case. The [FHS standard](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s18.html) does not specify it has to be, it't all up to how the system is configured by the admin. (especially on Linux) In my case, it's usually not a tmpfs.

All this does is ensure the link target directory exists if it is a symbolic link. It does not depend on `tmpfs` and will not fail in that case. The developer must manually symlink the `o` directory to a tmpfs location. Each individual developer or packager will know whether this applies to their machine.

matheusmoreira:
> It was a good idea for an experiment. Now we know that TmpFS requires root permissions and it offers a 3% performance boost. So I feel like it's not worth using. Builds already run in RAM since the kernel stores hard disk pages in memory. Thank you for sending this!

It made a bigger difference in single threaded build but still not very significant. Had a bigger impact on my own projects... Maybe your Makefile is just too fast already! :)

Might still be valuable for SSD users since it reduces writes to their hardware.

PR #96, opened 2021-03-01
error if unable to modify self

tomberek:
Avoid an infinite loop when there are no permissions to modify the file.


jart:
Thank you for your contribution!

Issue #105, opened 2021-03-06
Missing static_assert, c32rtomb, and mbrtoc32


dosisod:
I was trying to compile my programming language [Skull](https://github.com/dosisod/skull) (uses C and LLVM), and came across a few issues:

* `static_assert` is not available, though `_Static_assert` is. Submitting a PR for this soon.
* `c32rtomb` and `mbrtoc32` are not defined (new to C11 I believe). These might be a little bit more work, since they deal with encoding/decoding Unicode strings.
* A minor issue, but I use `-Wundef` in my Makefile, which causes a whole bunch of warnings (ie, `__LINKER__` not defined).

Let me know I should split some of these into their own issues. I though it would be best to put them in as one for the time being.

jart:
I'm happy to add `c32rtomb` and `mbrtoc32` for you. I'll go review your PR for static assert now.

`-Wundef` yes we can accommodate that. The public header is intended to be ANSI C89 compatible. I want it to be more warnings agnostic. Low-level libraries should respect a broad array of choices. See also ABSL's policy file.

jart:
I'm closing this issue out since it's pretty much resolved. Making the amalgamation header warnings agnostic is something I intend to happen and hope to happen sometime soon. Thanks again for the PRs!

PR #106, opened 2021-03-06
Add macro for static_assert

dosisod:
See #105


jart:
Thanks for your contribution!

Issue #104, opened 2021-03-05
Compiling Wren: some math, strtol and %g format string issues


alisonatwork:
Inspired by @ahgamut's work with Lua on #61 I thought I'd try compile [Wren](https://github.com/wren-lang/wren) just for fun. It worked pretty well with just a few minor changes. I cloned here with my changes on the `cosmopolitan` branch: https://github.com/alisonatwork/wren/tree/cosmopolitan

Changes required were:
- some small Makefile hacks to compile the static lib and test suite with the Cosmopolitan flags
- bunch of dummy header files to fake out `#include <stdio.h>` and so on
  - Since this was also required for Lua and everything else I have tried to compile, I wonder if it might be nice to provide these stubs as a package somewhere? Or is there another solution?
- missing math function `modf`
  - I inlined the one from musl libc
- a few function name clashes in the test suites
- workarounds for the `%g` format string problem
  - I used `if (trunc(f) == f) sprintf(s, "%ld", (long) f)` to at least get the int-ish doubles printing as expected and `%f` for best-effort on the floaty doubles

With these changes, 797 tests pass and 42 fail. Almost all of these are format string related like this:

```
      Expected output "0.123" on line 5 and got "0.123000".
      Expected output "0.3" on line 6 and got "0.300000".
      Expected output "-0.123" on line 7 and got "-0.123000".
```

But there are a few suspicious ones that I still need to investigate.

For now, the great news is that this language almost completely works!

The bug reports:

1. `modf` missing
2. `%g` format string

Requests for input:

1. Anyone have a good idea how we can hook the compiler or somehow fake out angle-bracket includes on-demand? Is there a better alternative to providing stub headers or pushing in `#ifdef __COSMOPOLITAN__` everywhere?

jart:
That's great news! You now have float formatting and modf. Name clashes are in scope since there's still a lot of technical debt to pay off in terms of removing functions I whipped up on the fly for the sake of moving quickly. 

In terms of all the posix headers, what I'm hoping we can get away with is a shell script example code snippet that looks something like this:

```sh
mkdir include
for x in string stdio stdlib ...; do
  touch include/$x.h
done
```

[...]

PR #107, opened 2021-03-07
fix strtoimax behavior when base 2 or 16 is supplied

alisonatwork:
While working on #104 I found that hex strings were not being parsed correctly. This was due to a typo in `strtoimax` where if base 2 or base 16 was explicitly supplied, it would not advance the pointer to the start of the number, so it would always resolve 0. I noticed `strtoumax` had a similar bug, so fixed it there too.


jart:
Good find! Thank you for sending this.

PR #108, opened 2021-03-07
Fix certain isystem headers including the wrong files

dosisod:


jart:
Thanks for updating these. The repository doesn't use them, but they might potentially be helpful for anyone wanting better include compatibility.

PR #109, opened 2021-03-07
Add tgmath.h

dosisod:
This PR adds the `tgmath.h` header, which allows for type generic math operations via C11's `_Generic` keyword.


jart:
It looks like the build is failing because `tgmath.h` is being pulled into the amalgamation header. Could you move it to the isystem folder?

jart:
Oh also tgmath.h needs to include math.h

dosisod:
Done. Still, when I compile on my system it works just fine.

Also, worth noting:

```
$ gcc --version
gcc (Ubuntu 5.4.0-6ubuntu1~16.04.12) 5.4.0 20160609
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

GCC is using a realllly old version (5.4.0).

jart:
LGTM. We've got GCC 9.x tucked away in the third_party folder. GNU Make and Almquist Shell are the only things we use from the Travis build environment.

PR #110, opened 2021-03-07
Set errno in strtol family of functions

alisonatwork:
This change makes sure that we set `errno` to `ERANGE` when the result of the `strtol` family of functions is out of range. This change caused problems in chibi cc, which was relying on non-standard behavior of `strtol` overflowing, so I changed it to call `strtoul` instead.

A further optimization might be to move this up into `strtoimax`, passing the limits as an argument. It might allow us to exit earlier if we know the number will go out of range. I was playing around with the way `strtoimax` is trying to figure out if it's out of range - it's using `bsr` macro and `bsrmax` ASM snippet - but it feels like there is something not working quite right in that code, and I need more time to figure it out. For now this will at least bring us more standards-compliance.

### DIFF
--- a/libc/fmt/strtol.c
+++ b/libc/fmt/strtol.c
@@ -17,6 +17,7 @@
 │ PERFORMANCE OF THIS SOFTWARE.                                                │
 ╚─────────────────────────────────────────────────────────────────────────────*/
 #include "libc/fmt/conv.h"
+#include "libc/errno.h"
 #include "libc/limits.h"
 
 /**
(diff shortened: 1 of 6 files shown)


alisonatwork:
FWIW I have a fix for `strtoimax` as well now, but still working on an elegant way to have the example calculator app continue supporting 128-bit unsigned ints. I'll open a new PR for that one.

jart:
Thanks for cleaning up my technical debt with those functions! I'm very happy to see you've also added tests.

I've had mixed feelings about having functions like `strtol` call `strtoimax`. If each of these functions had its own implementation then we could use fast overflow detection techniques like `__builtin_mul_overflow`.

alisonatwork:
Yeah, on my follow-up PR #111 I made a fix/change to `strtoimax` so it (hopefully) handles 128 bit overflow, but now that subtly changes the behavior of other `strtoimax`-backed functions like `atoi`, so I'm not sure if it's exactly the right fix.

PR #111, opened 2021-03-07
set errno when out of range in strtoimax

alisonatwork:
This is a follow-up to #110 in a separate PR because the change is a bit trickier to review.

The most important change is this bug fix in `strtoimax`:
```diff
-    if (!diglet || !x || (bits = bsr(diglet) + bsrmax(x)) < 127) {
+    if (!x || (bits = (diglet ? bsr(diglet) : 0) + bsrmax(x * base)) < 127) {
```

The problem is that if either `diglet` or `x` is 0, then `bsr` (or `bsrmax`) will return undefined. However, we can't just skip the check if `diglet` is 0, because when we multiply by `base` we still need to make sure that the result will be in range. The code looks a bit weird like this, but it does work.

One of the existing tests `testTwosBane` failed after changing this, but I moved it to `strtoumax` instead. What I've assumed here is that `strtoimax` should always check for overflow, but `strtoumax` is a faster implementation without bounds checking. This means that the `calculator` app needs to be updated too, so it can still perform bit-wise arithmetic on 128-bit unsigned ints. I did this by checking the first character of the token and calling `strtoimax` or `strtoumax` if it wasn't negative. That's not the cleanest solution, but I think it should work for most use cases.

The other behavior change here is that `atoi`, `atol` etc which according to the spec don't do error checking, will now still set `errno` to `ERANGE` when they go outside of the 128 bit signed int because they are backed by `strtoimax`. But... I think that should be okay.

### DIFF
--- a/libc/fmt/strtoimax.c
+++ b/libc/fmt/strtoimax.c
@@ -17,6 +17,7 @@
 │ PERFORMANCE OF THIS SOFTWARE.                                                │
 ╚─────────────────────────────────────────────────────────────────────────────*/
 #include "libc/fmt/conv.h"
+#include "libc/errno.h"
 #include "libc/limits.h"
 #include "libc/nexgen32e/bsr.h"
 #include "libc/str/str.h"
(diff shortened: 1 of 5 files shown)


jart:
Wow I can't believe I passed zero to bsr. This change looks good to me. I wouldn't be too concerned about `tool/build/calculator.c` since it's a toy included for demo purposes that's also used for a little bit of testing. If it gets in the way we can remove it. The changes you made to it look fine. Ideally it would would support infinite precision, but then it would be less helpful for the cases when I want to quickly see what a libm function does.

PR #116, opened 2021-03-09
Clean headers and add a working wsl/linux configuration for vscode

alisonatwork:
The current Visual Studio Code configuration appears to be trying to make the IntelliSense work by hacking the MSVC compiler to parse the Cosmopolitan headers. This only works (to whatever degree it does) when running VS Code on Windows, but it doesn't work when you are using VS Code on Linux, or using the remote VS Code server on a WSL installation.

I have updated the existing config to be called `Win32`, which means it will still be selected by default on Windows, and added a second `Linux` configuration which uses the included compiler to determine the IntelliSense settings (behind the scenes it does `${compilerPath} ${compilerArgs} -Wp,-v -E -dD -x c -m64 /dev/null` or clang/MSVC equivalent to figure out the defines). I'm not sure if I got all the right CFLAGS out of the box, but this seems to be something similar to what default `make` mode uses.

It might make non-WSL Windows users lives a bit easier if we had a configuration mode with all the correct switches set up for the `cross9.zip` from https://justine.lol/cosmopolitan/windows-compiling.html but I haven't tried to compile like that yet so can't test it. **Update** I have downloaded the cross9 gcc and it works perfectly. See: https://github.com/alisonatwork/cosmopolitan/commit/5de17ffc8ecdf1ff551d85940b67a8a9341ea492 

One quirk: you need to run the `make` once through to set up the links for the bundled C compiler before starting VS Code, or it won't be able to find the compiler, and then it just falls back to some kind of 32-bit safe mode with no defines at all.

I also cleaned up some duplicate and unused defines while I was here.

CC: @nicholatian - this is a follow-up to your #40 PR


alisonatwork:
BTW tip for anyone messing around on VS Code, I highly recommend turning on debug logging: https://code.visualstudio.com/docs/cpp/enable-logging-cpp and checking the "C/C++"

If you restart with the debug logging on, it will show you all the C compiler commands it runs in the output tab, even the behind-the-scenes ones it uses to bootstrap IntelliSense.

jart:
Looks good to me.

> The current Visual Studio Code configuration appears to be trying to make the IntelliSense work by hacking the MSVC compiler to parse the Cosmopolitan headers

The headers fall back to ANSI C89 mode. That's an intended behavior for the public amalgamation release. Supporting things like msvc for development within this codebase is not what I want. It sounds like you've found a way for us to keep having vscode support without needing to care about msvc. That would be fabulous.

> I highly recommend turning on debug logging

Should we put that in the `.vscode/settings.json` file? I prefer to have build systems print what they're doing.

> Update I have downloaded the cross9 gcc and it works perfectly

I'd support a change that downloads cross9 automatically for vscode users on windows, provided it verifies a hardcoded sha256 sum. cross9 should work not only for code completion, but for building the entire repository on windows without needing to change any configuration or having to resort to a meta-makefile like cmake. The only issue I foresee is argument size limits, which can be addressed.

> One quirk: you need to run the make once through to set up the links for the bundled C compiler before starting VS Code

On Linux we can fix that with a shell script wrapper that runs `third_party/gcc/unbundle.sh` to extract the compiler if it's not there and then says `exec o/third_party/gcc/bin/x86_64-linux-musl-gcc "$@"`. This is one of the services provided by the `build/bootstrap/compile.com` executable but that program would likely be overkill for VSCode.

PR #118, opened 2021-03-10
Add noreturn macro

dosisod:
C11 defines `noreturn` as a macro to `_Noreturn` in the `stdnoreturn.h` header.

### DIFF
--- a/libc/isystem/stdnoreturn.h
+++ b/libc/isystem/stdnoreturn.h
@@ -1,0 +1,15 @@
+#ifndef LIBC_ISYSTEM_STDNORETURN_H_
+#define LIBC_ISYSTEM_STDNORETURN_H_
+
+#if !(__ASSEMBLER__ + __LINKER__ + 0)
+#if __STDC_VERSION__ + 0 >= 201112
+COSMOPOLITAN_C_START_
+
+#define noreturn _Noreturn
+
+COSMOPOLITAN_C_END_
+
+#endif /* C11 */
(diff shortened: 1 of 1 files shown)


libc/isystem/stdnoreturn.h
```diff
@@ -0,0 +1,15 @@
+#ifndef LIBC_ISYSTEM_STDNORETURN_H_
+#define LIBC_ISYSTEM_STDNORETURN_H_
+
+#if !(__ASSEMBLER__ + __LINKER__ + 0)
+#if __STDC_VERSION__ + 0 >= 201112
+COSMOPOLITAN_C_START_
+
+#define noreturn _Noreturn
```
jart, line 8: libc/integral/c.inc defines a token called `wontreturn` so you could make `noreturn` simply be that. We could also potentially update `wontreturn` to use the C11 definition if it's available. Here's what we've got now:

```c
#ifndef wontreturn
#if !defined(__STRICT_ANSI__) &&      \
    (__has_attribute(__noreturn__) || \
     (__GNUC__ + 0) * 100 + (__GNUC_MINOR__ + 0) >= 208)
#define wontreturn __attribute__((__noreturn__))
#else
#define wontreturn
#endif
#endif
```

jart:
Thanks for your contribution! For the record `COSMOPOLITAN_C_START_` is only needed when you would otherwise use `extern "C"` in a header but it's no matter.

PR #122, opened 2021-03-10
Implemented missing socketpair

fabriziobertocci:
Added the missing implementation of socketpair(). Since Win32 doesn't have socketpair (as far as I know), I have implemented a simple wrapper that creates two individual sockets, set them in non-blocking mode and connects them together.
sysv implementation simply calls into sys_socketpair() that was already implemented.

### DIFF
--- a/libc/sock/socketpair-nt.c
+++ b/libc/sock/socketpair-nt.c
@@ -29,108 +29,225 @@
 #include "libc/sysv/consts/sock.h"
 #include "libc/sysv/consts/af.h"
 #include "libc/sysv/consts/ipproto.h"
+#include "libc/sysv/consts/inaddr.h"
 
-textwindows int sys_socketpair_nt(int family, int type, int protocol, int sv[2]) {
-    struct sockaddr_storage ss;
-    struct sockaddr_in *sa = (struct sockaddr_in *)&ss;
-    uint32_t ss_len;
-    int spR = -1, spW = -1;
-    int opt;
-    int rc = -1;
+/* Polyfill for socketpair() on Windows with family=AF_INET and 
+ * type=SOCK_STREAM or SOCK_SEQPACKET
+ */
+textwindows int sys_socketpair_nt_stream(int family, int type, int proto, int sv[2]) {
+  struct sockaddr_storage ss;
+  struct sockaddr_in *sa = (struct sockaddr_in *)&ss;
+  uint32_t ss_len;
+  int spR = -1, spW = -1;
+  int opt;
+  int rc = -1;
+  int listensock = -1;
 
-    int listensock = -1;
-    sv[0] = -1;     /* INVALID_SOCKET */
-    sv[1] = -1;     /* INVALID_SOCKET */
+  /* Avoid warnings for unused parameters */
+  (void)family;
+  (void)proto;
 
-    memset(&ss, 0, sizeof(ss));
-    if (family == AF_INET){
-        sa->sin_family = family;
-        sa->sin_addr.s_addr = htonl(0x7f000001);
-        sa->sin_port = 0;
-        ss_len = sizeof(struct sockaddr_in);
-    } else {
-        return -1;
+  sv[0] = -1;     /* INVALID_SOCKET */
+  sv[1] = -1;     /* INVALID_SOCKET */
+
+  memset(&ss, 0, sizeof(ss));
+  sa->sin_family = AF_INET;
+  sa->sin_addr.s_addr = htonl(INADDR_LOOPBACK);
+  sa->sin_port = 0;
+  ss_len = sizeof(struct sockaddr_in);
+
+  listensock = sys_socket_nt(AF_INET, type, IPPROTO_TCP);
+  if (listensock == -1) {
+    errno = WSAGetLastError();
+    goto done;
+  }
+
+  if (bind(listensock, (struct sockaddr *)&ss, ss_len) == -1) {
+    goto done;
+  }
+
+  if (listen(listensock, 1) == -1) {
+    goto done;
+  }
+
+  memset(&ss, 0, sizeof(ss));
+  ss_len = sizeof(ss);
+  if (getsockname(listensock, (struct sockaddr *)&ss, &ss_len) < 0) {
+    goto done;
+  }
+
+  sa->sin_family = AF_INET;
+  sa->sin_addr.s_addr = htonl(INADDR_LOOPBACK);
+  ss_len = sizeof(struct sockaddr_in);
+
+  spR = socket(AF_INET, type, IPPROTO_TCP);
+  if (spR == -1) {
+    goto done;
+  }
+
+  /* Set non-blocking */
+  opt = fcntl(spR, F_GETFL, 0);
+  if (opt == -1) {
+    goto done;
+  }
+  if(fcntl(spR, F_SETFL, opt | O_NONBLOCK) == -1) {
+    goto done;
+  }
+
+  if (connect(spR, (struct sockaddr *)&ss, ss_len) < 0) {
+    errno = WSAGetLastError();
+    if (errno != EINPROGRESS && errno != EWOULDBLOCK) {
+      goto done;
     }
+  }
 
-    listensock = sys_socket_nt(family, SOCK_STREAM, IPPROTO_TCP);
-    if (listensock == -1){
-        goto done;
+  spW = accept(listensock, NULL, 0);
+  if(spW == -1) {
+    errno = WSAGetLastError();
+    if (errno != EINPROGRESS && errno != EWOULDBLOCK) {
+      goto done;
     }
+  }
+  /* Set non-blocking */
+  opt = fcntl(spW, F_GETFL, 0);
+  if (opt == -1) {
+    goto done;
+  }
+  if(fcntl(spW, F_SETFL, opt | O_NONBLOCK) == -1) {
+    goto done;
+  }
 
-    if (bind(listensock, (struct sockaddr *)&ss, ss_len) == -1){
-        goto done;
-    }
-
-    if (listen(listensock, 1) == -1){
-        goto done;
-    }
-
-    memset(&ss, 0, sizeof(ss));
-    ss_len = sizeof(ss);
-    if (getsockname(listensock, (struct sockaddr *)&ss, &ss_len) < 0){
-        goto done;
-    }
-
-    sa->sin_family = family;
-    sa->sin_addr.s_addr = htonl(0x7f000001);
-    ss_len = sizeof(struct sockaddr_in);
-
-    spR = socket(family, SOCK_STREAM, IPPROTO_TCP);
-    if (spR == -1){
-        goto done;
-    }
-
-    /* Set non-blocking */
-    opt = fcntl(spR, F_GETFL, 0);
-    if (opt == -1) {
-        goto done;
-    }
-    if(fcntl(spR, F_SETFL, opt | O_NONBLOCK) == -1){
-        goto done;
-    }
-
-    if (connect(spR, (struct sockaddr *)&ss, ss_len) < 0){
-        errno = WSAGetLastError();
-        if (errno != EINPROGRESS && errno != EWOULDBLOCK){
-            goto done;
-        }
-    }
-
-    spW = accept(listensock, NULL, 0);
-    if(spW == -1){
-        errno = WSAGetLastError();
-        if (errno != EINPROGRESS && errno != EWOULDBLOCK){
-            goto done;
-        }
-    }
-    /* Set non-blocking */
-    opt = fcntl(spW, F_GETFL, 0);
-    if (opt == -1) {
-        goto done;
-    }
-    if(fcntl(spW, F_SETFL, opt | O_NONBLOCK) == -1){
-        goto done;
-    }
-
-    rc = 0; /* Success */
+  rc = 0; /* Success */
 
 done:
-    if (rc == -1) {
-        errno = WSAGetLastError();
-        if (spR != -1) {
-            close(spR);
-        }
-        if (spW != -1) {
-            close(spW);
-        }
-    } else {
-        sv[0] = spR;
-        sv[1] = spW;
+  if (rc == -1) {
+    if (spR != -1) {
+      close(spR);
     }
+    if (spW != -1) {
+      close(spW);
+    }
+  } else {
+    sv[0] = spR;
+    sv[1] = spW;
+  }
 
-    /* Both success and failure */
-    if (listensock != -1) {
-        close(listensock);
-    }
-    return rc;
+  /* Both success and failure */
+  if (listensock != -1) {
+    close(listensock);
+  }
+  return rc;
 }
+
+
+int sys_socketpair_nt_dgram(int family, int type, int proto, int sv[2]) {
+  struct sockaddr_in sa;
+  uint32_t sa_len;
+  int opt;
+  int spR = -1, spW = -1;
+  int rc = -1;
+
+  /* Avoid warnings for unused parameters */
+  (void)family;
+  (void)proto;
+
+  sv[0] = -1;     /* INVALID_SOCKET */
+  sv[1] = -1;     /* INVALID_SOCKET */
+  memset(&sa, 0, sizeof(sa));
+  sa.sin_family = AF_INET;
+  sa.sin_addr.s_addr = htonl(INADDR_LOOPBACK);
+  sa.sin_port = 0;
+  sa_len = sizeof(struct sockaddr_in);
+
+  /* Creates a read socket */
+  spR = sys_socket_nt(AF_INET, SOCK_DGRAM, IPPROTO_UDP);
+  if (spR == -1) {
+    errno = WSAGetLastError();
+    goto done;
+  }
+
+  /* ... and bind it to whatever port is available on localhost */
+  if (bind(spR, (struct sockaddr *)&sa, sa_len) < 0) {
+    goto done;
+  }
+
+  /* Read the port # that bind obtained */
+  memset(&sa, 0, sizeof(sa));
+  sa_len = sizeof(sa);
+  if (getsockname(spR, (struct sockaddr *)&sa, &sa_len) < 0) {
+    goto done;
+  }
+  // The port # is on: sa.sin_port
+  // printf("RD Bound: port=%d, addr=%08x\n", sa.sin_port, sa.sin_addr.s_addr);
+
+  /* Creates a write socket */
+  spW = sys_socket_nt(AF_INET, SOCK_DGRAM, IPPROTO_UDP);
+  if (spW == -1) {
+    errno = WSAGetLastError();
+    goto done;
+  }
+  /* ... and connect it to the read socket port */
+  if (connect(spW, (struct sockaddr *)&sa, sa_len) < 0) {
+    goto done;
+  }
+
+  /* So far we got connection only in one way:
+   *      WR ----> RD
+   * To ensure symmetric behavior, read back the port bound of the 
+   * write socket */
+  memset(&sa, 0, sizeof(sa));
+  sa_len = sizeof(sa);
+  if (getsockname(spW, (struct sockaddr *)&sa, &sa_len) < 0) {
+    goto done;
+  }
+
+  // The write socket is bound at: sa.sin_port
+  //printf("WR Bound: port=%d, addr=%08x\n", sa.sin_port, sa.sin_addr.s_addr);
+
+  // ... finally connect the READ socket to the WRITE socket
+  if (connect(spR, (struct sockaddr *)&sa, sa_len) < 0) {
+    goto done;
+  }
+
+  /* Now we finally got both bi-directional sockets: RD <---> WR */
+
+  /* Finally set both sockets as non-blocking */
+  opt = fcntl(spW, F_GETFL, 0);
+  if (opt == -1) {
+    goto done;
+  }
+  if(fcntl(spW, F_SETFL, opt | O_NONBLOCK) == -1) {
+    goto done;
+  }
+
+  opt = fcntl(spR, F_GETFL, 0);
+  if (opt == -1) {
+    goto done;
+  }
+  if(fcntl(spR, F_SETFL, opt | O_NONBLOCK) == -1) {
+    goto done;
+  }
+
+  rc = 0;
+
+done:
+  if (rc == -1) {
+    errno = WSAGetLastError();
+    if (spR != -1) {
+      close(spR);
+    }
+    if (spW != -1) {
+      close(spW);
+    }
+  } else {
+    sv[0] = spR;
+    sv[1] = spW;
+  }
+
+  return rc;
+}
+
+
+
+
+
--- a/libc/sock/socketpair-nt.c
+++ b/libc/sock/socketpair-nt.c
@@ -39,7 +39,6 @@ textwindows int sys_socketpair_nt_stream(int family, int type, int proto, int sv
   struct sockaddr_in *sa = (struct sockaddr_in *)&ss;
   uint32_t ss_len;
   int spR = -1, spW = -1;
-  int opt;
   int rc = -1;
   int listensock = -1;
 
@@ -85,15 +84,6 @@ textwindows int sys_socketpair_nt_stream(int family, int type, int proto, int sv
     goto done;
   }
 
-  /* Set non-blocking */
-  opt = fcntl(spR, F_GETFL, 0);
-  if (opt == -1) {
-    goto done;
-  }
-  if(fcntl(spR, F_SETFL, opt | O_NONBLOCK) == -1) {
-    goto done;
-  }
-
   if (connect(spR, (struct sockaddr *)&ss, ss_len) < 0) {
     errno = WSAGetLastError();
     if (errno != EINPROGRESS && errno != EWOULDBLOCK) {
@@ -107,14 +97,6 @@ textwindows int sys_socketpair_nt_stream(int family, int type, int proto, int sv
     if (errno != EINPROGRESS && errno != EWOULDBLOCK) {
       goto done;
     }
-  }
-  /* Set non-blocking */
-  opt = fcntl(spW, F_GETFL, 0);
-  if (opt == -1) {
-    goto done;
-  }
-  if(fcntl(spW, F_SETFL, opt | O_NONBLOCK) == -1) {
-    goto done;
   }
 
   rc = 0; /* Success */
@@ -143,7 +125,6 @@ done:
 int sys_socketpair_nt_dgram(int family, int type, int proto, int sv[2]) {
   struct sockaddr_in sa;
   uint32_t sa_len;
-  int opt;
   int spR = -1, spW = -1;
   int rc = -1;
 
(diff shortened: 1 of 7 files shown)


fabriziobertocci:
This PR is part of building Mosquitto for Cosmopolitan. See: https://github.com/fabriziobertocci/mosquitto/tree/cosmopolitan

libc/sock/socketpair-nt.c
```diff
@@ -0,0 +1,136 @@
...
+    sv[1] = -1;     /* INVALID_SOCKET */
+
+    memset(&ss, 0, sizeof(ss));
+    if (family == AF_INET){
+        sa->sin_family = family;
+        sa->sin_addr.s_addr = htonl(0x7f000001);
+        sa->sin_port = 0;
+        ss_len = sizeof(struct sockaddr_in);
+    } else {
+        return -1;
```
alisonatwork, line 52: Here we are missing `errno`. I believe it should be `EAFNOSUPPORT` for this case.

libc/sock/socketpair.c
```diff
@@ -0,0 +1,53 @@
...
+ * @asyncsignalsafe
+ */
+int socketpair(int family, int type, int protocol, int sv[2]) {
+  if (family == AF_UNSPEC) {
+    family = AF_INET;
+  } else if (family == AF_INET6) {
+    /* Recommend IPv6 on frontend serving infrastructure only. That's
+       what Google Cloud does. It's more secure. It also means poll()
+       will work on Windows, which doesn't allow mixing third layers. */
+    return epfnosupport();
```
alisonatwork, line 46: I think perhaps this should be `EAFNOSUPPORT` too. I have searched around for `EPFNOSUPPORT` and I can find some references to it in the Linux kernel, but it does not seem to be a POSIX `errno`.

libc/sock/socketpair-nt.c
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    listensock = sys_socket_nt(family, SOCK_STREAM, IPPROTO_TCP);
```
alisonatwork, line 55: Is it intentional to not pass through the `type` and `protocol` that was supplied? (Also on line 78.) I would imagine that if for some reason Windows can't support other types or protocols, that should be handled in the `sys_socket_nt` implementation.

libc/sock/socketpair-nt.c
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    memset(&ss, 0, sizeof(ss));
+    ss_len = sizeof(ss);
+    if (getsockname(listensock, (struct sockaddr *)&ss, &ss_len) < 0){
+        goto done;
+    }
+
+    sa->sin_family = family;
+    sa->sin_addr.s_addr = htonl(0x7f000001);
```
alisonatwork, line 75: Perhaps can use `INADDR_LOOPBACK` here.

alisonatwork:
This is cool. I noticed this blog from a few years ago saying AF_UNIX should work on contemporary Windows: https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/ If we can detect that feature it might be nice to allow AF_UNIX to go through for `socketpair` since I imagine most Linux-based code will not pass AF_INET (see https://man7.org/linux/man-pages/man2/socketpair.2.html)

libc/sock/socketpair-nt.c
fabriziobertocci: Right. Added

libc/sock/socketpair.c
fabriziobertocci: Hmm it seems `EAFNOSUPPORT` is a valid error POSIX code for `socketpair()`: https://pubs.opengroup.org/onlinepubs/009696699/functions/socketpair.html

libc/sock/socketpair-nt.c
fabriziobertocci: Hmm well good point. According to the Winsock doc, `WSASocket` (called from `sys_socket_nt`) handles other types/protocols. So I suspect if `socketpair()` is invoked with `type=SOCK_DGRAM` (apparently perfectly valid) the implementation is going to fail miserably when it calls `listen()` because DGRAM is a disconnected socket.
Now, it seems that `socketpair()` also support `type=SOCK_SEQPACKET`, and that should be ok since SEQPACKET is a connected socket... 
All right Let me run some tests and then I'll have to re-implement it. I'm thinking of something along those lines (but I need to make sure it works with DGRAM and SEQPACKET first):
```
  if (type==SOCK_DGRAM) {
     // create a pair of disconnected socket bound together
     fd1 = socket(AF_INET, SOCK_DGRAM, proto);
     bind(fd1, <localhost, any port>);
     fd2 = socket(AF_INET, SOCK_DGRAM, proto);
     connect(fd2, <localhost, port_where_fd1_is_bound>);
  } else if ((type == SOCK_STREAM) || (type == SOCK_SEQPACKET)) {
     // Similar to current implementation
  } else {
     // Unsupported type
  }
```

libc/sock/socketpair-nt.c
fabriziobertocci: Funny, I did use INADDR_LOOPBACK at first, but I could not find it defined anywhere (and gcc complained). I must have missed (is defined in `libc/sysv/consts/inaddr.h`)...
Changed.

fabriziobertocci:
> This is cool. I noticed this blog from a few years ago saying AF_UNIX should work on contemporary Windows: https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/ If we can detect that feature it might be nice to allow `AF_UNIX` to go through for `socketpair` since I imagine most Linux-centric code will not pass `AF_INET` (see https://man7.org/linux/man-pages/man2/socketpair.2.html)

Hmm interesting. Yes, I agree, socketpair() is mostly designed to work with AF_UNIX... But how can we detect if AF_UNIX is supported? `socketpair` was used by Mosquitto, and I guess it forced to use AF_INET because AF_UNIX is not available in all the Windows.
Before I go down this rabbit hole of dealing with this multitude of combination of platform/family/type/proto, let me know what are your thoughts here. I think for now we should just forget about supporting `AF_UNIX` on Windows (and always use AF_INET and ignore and type/proto on this case)...

libc/sock/socketpair.c
alisonatwork: I wonder if `EPFNOSUPPORT` is a historical return value from the days when `PF_INET` and `AF_INET` were theoretically different? I think this is one of the most fun things about working on code that supports legacy functionality, it's as much archaeology/history as it is programming.

alisonatwork:
I will defer to @jart what she would like to do about this, since it's her project and she knows better what the philosophy should be.

My personal feeling is that in the ideal situation, you should be able to compile a single program and have it Just Work on any of the platforms (Windows, BSD, Linux). That means it is okay to do some polyfilling of behavior that we know will definitely fail on a certain platform. Since Windows is supported back to Vista, I think overriding `AF_UNIX` with `AF_INET` for the purpose of putting together a functioning `socketpair` is acceptable, even if it does do something slightly different than what the caller intended.

I do feel like it would be best to detect `afunix.sys` capabilities and use them if they're there, since that would produce a more correct result, but it does add complexity that perhaps is not worth it at this stage.

fabriziobertocci:
@alisonatwork, @jart: Ok so this turned out to be bigger than I expected, anyway, I have (re)implemented socketpair with `AF_INET` for Windows. I have two versions: one that build a stream pair of bi-directional connected sockets (using `SOCK_STREAM`) and another version that used datagram (`SOCK_DGRAM`). I have tested both implementations (on Linux, as I don't have a handy Windows build machine around) and they seems to be working as expected. 

Now, regarding supporting `AF_UNIX` on Windows, as Alison pointed out, some newer version of Windows supports it. Right now in my implementation, if the caller of `socketpair()` request `AF_UNIX` (on Windows), I'm calling the `AF_INET` version. I can look into detecting if AF_UNIX is available and do a polyfill for that, but not any time soon.
Alternatively, if socketpair() is called with AF_UNIX we could fail with `EAFNOSUPPORT` (but I don't like this solution)...

alisonatwork:
@fabriziobertocci you have opened a can of worms here!

Try adding this unit test to `test/libc/sock/socketpair_test.c`
```c
/*-*- mode:c;indent-tabs-mode:nil;c-basic-offset:2;tab-width:8;coding:utf-8 -*-│
│vi: set net ft=c ts=2 sts=2 sw=2 fenc=utf-8                                :vi│
╞══════════════════════════════════════════════════════════════════════════════╡
│ Copyright 2021 Alison Winters                                                │
│                                                                              │
│ Permission to use, copy, modify, and/or distribute this software for         │
│ any purpose with or without fee is hereby granted, provided that the         │
│ above copyright notice and this permission notice appear in all copies.      │
│                                                                              │
│ THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL                │
│ WARRANTIES WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED                │
│ WARRANTIES OF MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE             │
│ AUTHOR BE LIABLE FOR ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL         │
│ DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR        │
│ PROFITS, WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER               │
│ TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR             │
╚─────────────────────────────────────────────────────────────────────────────*/
#include "libc/assert.h"
#include "libc/dce.h"
#include "libc/sock/sock.h"
#include "libc/sysv/consts/af.h"
#include "libc/sysv/consts/sock.h"
#include "libc/testlib/testlib.h"

TEST(socketpair, testAfUnix) {
  int fd[2];
  const char ping[] = "ping";
  const char pong[] = "pong";
  char buf[32];
  if (IsWindows()) {
    // fails on windows
    return;
  }
  ASSERT_NE(-1, socketpair(AF_UNIX, SOCK_STREAM, 0, fd));
  ASSERT_EQ(sizeof(ping), write(fd[0], ping, sizeof(ping)));
  ASSERT_EQ(sizeof(ping), read(fd[1], buf, sizeof(ping)));
  EXPECT_STREQ(ping, buf);
  ASSERT_EQ(sizeof(pong), write(fd[1], pong, sizeof(pong)));
  ASSERT_EQ(sizeof(pong), read(fd[0], buf, sizeof(pong)));
  EXPECT_STREQ(pong, buf);
  ASSERT_NE(-1, close(fd[0]));
  ASSERT_NE(-1, close(fd[1]));
}

TEST(socketpair, testAfInet) {
  int fd[2];
  const char ping[] = "ping";
  const char pong[] = "pong";
  char buf[32];
  if (!IsWindows()) {
    // fails on unix
    return;
  }
  ASSERT_NE(-1, socketpair(AF_INET, SOCK_STREAM, 0, fd));
  ASSERT_EQ(sizeof(ping), write(fd[0], ping, sizeof(ping)));
  ASSERT_EQ(sizeof(ping), read(fd[1], buf, sizeof(ping)));
  EXPECT_STREQ(ping, buf);
  ASSERT_EQ(sizeof(pong), write(fd[1], pong, sizeof(pong)));
  ASSERT_EQ(sizeof(pong), read(fd[0], buf, sizeof(pong)));
  EXPECT_STREQ(pong, buf);
  ASSERT_NE(-1, close(fd[0]));
  ASSERT_NE(-1, close(fd[1]));
}
```

On Windows, this fails as follows:

```
$ ./socketpair_test.com
error:test/libc/sock/socketpair_test.c:55: socketpair_testAfInet()
        ASSERT_NE(-1, socketpair(AF_INET, SOCK_STREAM, 0, fd))
                need -1 (or 0xffffffffffffffff) ≠
                 got -1 (or 0xffffffffffffffff)
        EINVAL/err=87/errno:87/GetLastError:203 The parameter is incorrect.
        C:/Users/alicl/tmp/socketpair_test.com @ surfyboxx
1 / 1 tests failed
```

And if you check `fcntl-nt.c` you will see that `F_SETFL` is not supported. You might want to use `ioctlsocket` instead (see https://docs.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-ioctlsocket).

alisonatwork:
Although, looking at the test, it doesn't seem like it really gets us any closer to cross-platform working if `AF_INET` always fails on Linux and `AF_UNIX` always fails on Windows. That means any programmer who wants to use `socketpair` has to have some kind of run-time check of their own to select what to do on different platforms anyway. That kinda undermines the purpose, I think. From that perspective, it seems like a better solution would be to implement AF_UNIX behind the scenes as AF_INET, for old versions of Windows, and errno out on AF_INET altogether, since that seems to be the "standard" UNIX behavior. Not sure what the BSDs do here.

jart:
Hey folks great work. I'm coming back online now after a deep and dreamless slumber.

> I will defer to @jart what she would like to do about this, since it's her project and she knows better what the philosophy should be.

It's our project Alison, and yours too Fabrizio.

Based on my reading of the [glibc docs](https://www.gnu.org/software/libc/manual/html_node/Socket-Pairs.html) the *purpose* of the `socketpair` system call is to create a bidirectional pipe. Therefore socket a misnomer, and as such, the correct thing to do on Windows is likely passing `kNtPipeAccessDuplex` to `CreateNamedPipe`. See also [libc/calls/pipe-nt.c](https://github.com/jart/cosmopolitan/blob/master/libc/calls/pipe-nt.c).

@fabriziobertocci could you please send a short email to jtunney@gmail.com confirming that your intent is to assign me the copyright to code you contribute to cosmopolitan?

fabriziobertocci:
Sigh, I knew I was going down a rabbit hole with no end in sight... 
Failing on Windows if AF_UNIX is not supported is not a good option in my opinion. Take for example Mosquitto (the starter of this mess): in file [`net/net_mosq.c`] (https://github.com/fabriziobertocci/mosquitto/blob/master/lib/net_mosq.c#L1073) they wraps socketpair in their own `net__socketpair`. For the Windows implementation they just use AF_INET with SOCK_STREAM. And this appears to be a decent workaround.
NOTE: I just noticed that setting the created socket as non-blocking is a requirement only of Mosquitto and is not the default behavior for `socketpair()` so I have removed the part that calls `fcntl()` (that was also causing problems...).

@jart I just saw your reply as I was typing this: ok, email sent. Regarding the use of named pipes... ok we can try to replace socketpair with sys_pipe_nt... but not having a Windows machine I can use for testing, makes everything hard for me.

alisonatwork:
This is definitely a tricky problem.

Ideally if you compile with Cosmopolitan, it shouldn't activate the `#ifdef WIN32` branch of `net/net_mosq.c` - you should just compile a single standard/portable version and let Cosmopolitan do the OS-specific mapping. So the right thing to do would be to support `AF_UNIX` on Windows in Cosmopolitan in some way that is compatible with how other OSes would do it. The whole point is to not have to compile two different binaries!

However, Mosquitto is (sensibly) expecting that a file descriptor returned from `socketpair` is, in fact, a socket, and that on UNIX you can call `fcntl` on it. There is another `#ifdef WIN32` branch in the same file you shared which sets the socket to nonblocking in the Windows way (using `ioctlsocket`), but the Cosmopolitan compile won't include that branch, so it would fail at runtime on Windows when doing the `fcntl`.

It seems to me there are several things that should be done here.

1. Some kind of Windows implementation for `socketpair` in Cosmopolitan, which imo should use Windows 10 `AF_UNIX` support by default, since that could cause the least amount of surprises if developers subsequently use the returned file descriptor in a UNIX-y way. Only supporting `SOCK_STREAM` seems fine in this case, since that is all Windows supports and anyway it seems to be the most commonly-used type on UNIX too.
2. Some kind of fallback `socketpair` in Cosmopolitan for Windows <10, which should implement something similar to `AF_UNIX` using a different technology behind the scenes. Named pipe would be faster and more Windows-y than a local TCP/UDP socket, but I do wonder if named pipes might introduce other unexpected behavior, e.g. is the `HANDLE` (= `int64_t`) returned from `CreateNamedPipeA` really drop-in compatible with UNIX `int`-sized file descriptor?
3. I wonder if there is something we can do with `fcntl` on Windows so that the presumably fairly common case of setting `O_NONBLOCK` can be replaced with an `ioctl`? On the other hand, we already do the right thing with Linux-style `SOCK_NONBLOCK` on Windows, so perhaps the advice to developers should be "just do it the Linux way", which means you can skip the spurious `fcntl` syscalls anyway. But skipping that might break code on BSDs that don't support it.

In any case, if you don't have a Windows box to test on, this is going to be very difficult to solve nicely 😄  Once we get Windows build working, we could set up CI to run tests on a Windows box too, which should make it easier.

alisonatwork:
@jart would something like this be awful for the third case?

```diff
diff --git a/libc/calls/fcntl-nt.c b/libc/calls/fcntl-nt.c
index 7dce493b..778ec988 100644
--- a/libc/calls/fcntl-nt.c
+++ b/libc/calls/fcntl-nt.c
@@ -22,8 +22,10 @@
 #include "libc/nt/enum/fileflagandattributes.h"
 #include "libc/nt/enum/filesharemode.h"
 #include "libc/nt/files.h"
+#include "libc/nt/winsock.h"
 #include "libc/sysv/consts/f.h"
 #include "libc/sysv/consts/fd.h"
+#include "libc/sysv/consts/fio.h"
 #include "libc/sysv/consts/o.h"
 #include "libc/sysv/errfuns.h"

@@ -35,6 +37,12 @@ textwindows int sys_fcntl_nt(int fd, int cmd, unsigned arg) {
         return g_fds.p[fd].flags & (O_ACCMODE | O_APPEND | O_ASYNC | O_DIRECT |
                                     O_NOATIME | O_NONBLOCK);
       case F_SETFL:
+        if ((arg & O_NONBLOCK) && __isfdkind(fd, kFdSocket)) {
+          if (__sys_ioctlsocket_nt(fd, FIONBIO, (uint32_t[1]){1}) == -1) {
+            return __winsockerr();
+          }
+          return 0;
+        }
         /*
          * - O_APPEND doesn't appear to be tunable at cursory glance
          * - O_NONBLOCK might require we start doing all i/o in threads
```

jart:
Yes we can do that so long as you say `weaken(__sys_ioctlsocket_nt)` and add `STATIC_YOINK("__sys_ioctlsocket_nt")` to `libc/sock/yoink.inc`. But as I mentioned earlier, I don't believe `socketpair` should need winsock at all on NT. Can't the polyfill just call `CreateNamedPipe` instead on Windows?

fabriziobertocci:
Justine,
Yes, I agree with you that sockepair should just use named pipes on
Windows. Now, the question is why the MQTT folks decided to use
AF_INET+SOCK_STREAM instead?... oh well, I need to find a machine around
(physical or virtual) with Windows devenv to try first...
Fab

alisonatwork:
I'm cool with named pipe! It is a simpler and more performant solution.

On why MQTT went with TCP socket, who knows? I don't think they are the first - it seems like there are many of these polyfills floating around on the net. I saw one in Python, there's an open PR for Ocaml to put one in, there's one in libuv, and another one in libevent... Seems to be pretty common. This is one of the issues that I think Cosmopolitan is addressing - right now every library or every app that wants to work cross-platform implements their own hacks to try make it work. It would be much nicer to live in a world where we had a platform-agnostic libc to start with. So, happy to help out wherever I can. If you need a Windows guinea pig to test something just let me know.

fabriziobertocci:
Well I had some older Win7 machines & VMs around, but I guess if I want to
test a newer version that supports AF_UNIX I would need a Win10 VM... so I
am setting up a VM right now with Win10 Home.
Stay tuned.
Fab

fabriziobertocci:
@alisonatwork @jart
Ok, so I have this possible solution. I have not committed any code because I want to first check with you a couple of things:
The top-level implementation of socketpair is now simple:
```
...
  if (!IsWindows()) {
    return sys_socketpair(family, type, protocol, sv);
  }
  return sys_socketpair_nt(family, type, protocol, sv);
}
```
There is only one version for Windows that handles all the cases:
```
int sys_socketpair_nt(int family, int type, int proto, int sv[2]) {
    int64_t hpipe, h1, h2;
    int reader, writer;
    char16_t pipename[64];
    uint32_t mode;

    // Supports only AF_UNIX
    if (family != AF_UNIX) {
        errno = EAFNOSUPPORT;
        return -1;
    }

    mode = kNtPipeWait;
    if (type == SOCK_STREAM) {
        mode |= kNtPipeReadmodeByte | kNtPipeTypeByte;
    } else if ((type == SOCK_DGRAM) || (type == SOCK_SEQPACKET)) {
        mode |= kNtPipeReadmodeMessage | kNtPipeTypeMessage;
    } else {
        errno = EOPNOTSUPP;
        return -1;
    }

    CreatePipeName(pipename);
    if ((reader = __reservefd()) == -1) return -1;
    if ((writer = __reservefd()) == -1) {
        __releasefd(reader);
        return -1;
    }
    if ((hpipe = CreateNamedPipe(pipename, 
            kNtPipeAccessDuplex,
            mode, 
            1, 
            65536, 
            65536,
            0, 
            &kNtIsInheritable)) == -1) {
        __winerr();
        __releasefd(writer);
        __releasefd(reader);
        return -1;
    }

    h1 = CreateFile(pipename, 
            kNtGenericWrite | kNtGenericRead, 
            0,  // Not shared
            &kNtIsInheritable, 
            kNtOpenExisting, 0, 0);
    if (h1 == -1) {
        CloseHandle(hpipe);
        __winerr();
        __releasefd(writer);
        __releasefd(reader);
        return -1;
    }

    g_fds.p[reader].kind = kFdFile;
    g_fds.p[reader].flags = 0;      // TODO
    g_fds.p[reader].handle = hpipe;

    g_fds.p[writer].kind = kFdFile;
    g_fds.p[writer].flags = 0;      // TODO
    g_fds.p[writer].handle = h1;

    sv[0] = reader;
    sv[1] = writer;
    return 0;
}
```
Few questions:
* This version uses the function `CreatePipeName`. This function is defined in `libc/calls/pipe-nt.c`. It should not be duplicated (together with the other functions/globals: `UintToChar16Array` and `kPipeNamePrefix`). Where should be a good place to place this kind of shared code?
* the `g_fds.p[].flags`: what is the use? What should be the value?

I have a test app that uses socketpair() and I have tested both on Linux and Windows and for what I can see, it seems to be working (consistent behavior between Linux and Windows), with both `SOCK_STREAM`, `SOCK_DGRAM` and `SOCK_SEQPACKET`).
Before I push any changes, let me know your thoughts on this.

Fab

jart:
That looks about right. I'm not too concerned about code duplication. We could probably incorporate @alisonatwork's test too after this change.

alisonatwork:
Yep, looks good to me too. Something I am discovering as I walk round the codebase is a lot of effort is spent on allowing the smallest possible link to happen. It makes sense since all Cosmopolitan binaries are static, so we don't want to bring in any more than we need. Code duplication can probably be refactored out in future PRs if it's really a problem.

My test is trivial, if we are only testing `AF_UNIX`! Feel free to rewrite to test different sock types if you like. Adding a test for `pipe` would probably look pretty similar.

jart:
That's correct. Part of the motivation is that programs should be autonomous and only pay for what they use. Otherwise we'd be like glibc where static hello world is 800kb or Go Hello World where it links in support for cuneiform.

jart:
@fabriziobertocci Let me know when the comment is incorporated so I can press the merge button!

jart:
Oh also concerning this:

> the g_fds.p[].flags: what is the use? What should be the value?

What could happen is `SOCK_CLOEXEC` could be or'd into `type` in which case you'd want flags to be `O_CLOEXEC` instead of 0. That's fine to leave for a follow-up change since it's not something that's commonly used.

fabriziobertocci:
Ok, so here is everything. I have consolidated the shared functions to create the pipe name (through internal.h). Please double check: those functions are not declared "textwindows" as I don't know exactly why we need to explicitly specify which function goes to .text area and which one does not. 
Anyway, I have also added the test suggested by @alisonatwork (reviewed to test unix sockets both with stream and datagram mode). It passes for me both on Windows and Linux (yes I eventually set up a Win10 VM...).
There is still missing the setting of the `g_fds.p[].flags`... I may do it tomorrow (so perhaps review those changes first but don't close the PR yet).

jart (approved):
Looks good to me! Thank you @fabriziobertocci for contributing this!

Also thank you @alisonatwork for contributing the test which ensures this contribution will be regression tested automatically by me on freebsd, openbsd, netbsd, rhel7, rhel5, xnu, win7, and win10 each time I type `make test` into the console. Should any issues arise in our support vector I'll fix it.

> Please double check: those functions are not declared "textwindows" as I don't know exactly why we need to explicitly specify which function goes to .text area and which one does not.

`textwindows` is a minor detail you are free to let me worry about since there isn't a test currently that spells out its intent. If anyone's curious the goal with that keyword is simply to have the Windows support code get linked into the same part of the binary, which gives us a marginal performance gain on other platforms where those pages needn't be loaded off disk. For example, I know how much Linux users love Linux, and this design helps ensure they needn't give up a single microsecond of latency to gain the benefits of having Cosmopolitan's broad compatibility baked-in.

jart:
I can confirm this change works great on all supported platforms. Polyfilling SOCK_CLOEXEC was tricky since it required polyfilling BSD too and as such I took care of it in 4177489. Feel free to send additional PRs for anything else you need, and once again thank you again for contributing to Cosmopolitan!

fabriziobertocci:
Hi @Justine Tunney ***@***.***>
> Out of curiosity what's the purpose of this comment? Is it a text editor
/ documentation generator / style guide thing?

It's the manual folding tag for vim. That let me custom define folds around
functions (in C) or methods/blocks around C++. I found that automatic
folding is not convenient, so I set:
```
set foldmethod=marker
```
then use `za` (open fold), `zc` (close fold), `zM` (close all folds), ...
to open/close the folds.
That's my way to manage my mental sanity when I work with large files with
a lot of functions.

About vim, I just realized that my code doesn't follow your indentation
standard (I usually indent with 4 spaces), but it should be easy to
re-indent the whole file (`=` function in vim).


>  once again thank you again for contributing to Cosmopolitan!
No problem, it was a fun exercise even for me to review AF_UNIX sockets and
Windows named pipes (that I have never used before).
I will probably continue to play around and understand more Cosmopolitan in
my free time...
One last question: pthread (or any support for threads). Not available ?
Regards,
Fab

jart:
I'm glad to hear you had fun. Right now that's the prime directive. 

Style isn't too much of a concern for me. My editor runs clang-format on save based on the config in the root of the project. So the worst that can happen is diff noise when I edit a file. If you want to keep the style of your contributions in tact, you can add a `/* clang-format off */` line to the file. The two lines at the start of each file also attempt to help Emacs and Vim understand the indentation style, but there's only so much they can do.

Threads are something that should happen, I want to happen, is bound to happen, but unlikely to be anytime soon. Based on the numerology in [syscalls.sh](https://github.com/jart/cosmopolitan/blob/master/libc/sysv/syscalls.sh) there's no obvious path forward because all the different systems don't agree on how threads should be done, because they weren't part of the original design of the fifth bell system, and end up being a radical departure considering how their requirements leak into everything.

One of the things that's nice about Cosmo is it makes fork() fast because programs are tiny which means fewer pages need to be copied on write. It's technically possible to simulate threads by simply having all your memory mappings be shared in which case forked processes behave as de facto threads, although I seem to recall Windows being strict about wanting msync() to be called.

PR #123, opened 2021-03-10
Added random() and srandom() functions

fabriziobertocci:
Added an inlined simple wrapper for random() and srandom() (that call into the existing rand/srand).

### DIFF
--- a/libc/rand/rand.h
+++ b/libc/rand/rand.h
@@ -26,6 +26,16 @@ uint64_t rdrand(void);
 uint64_t rdseed(void);
 float randf(void);
 
+/* alias for rand()  */
+forceinline int random(void) {
+    return rand();
+}
+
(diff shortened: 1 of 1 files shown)


fabriziobertocci:
This PR is part of building Mosquitto for Cosmopolitan. See: https://github.com/fabriziobertocci/mosquitto/tree/cosmopolitan

jart:
It's exciting to hear that Mosquitto intends to use Cosmopolitan! You should link to the project. I'm assuming it's this one? https://github.com/eclipse/mosquitto

Thanks for sending this change. It seems POSIX specifies that `random()` needs to be implemented as a non-linear additive feedback RNG whereas `rand()` is is just a linear congruential generator. I'll close out this PR after I've vendored the code for what Musl is doing here.

PR #124, opened 2021-03-11
Add missing functions to tgmath

dosisod:
Turns out when I implemented #109 I forgot a lot of functions/macros, so this PR adds them.


jart:
It's worth noting that some compilers like chibicc which support C11 may define `__STDC_NO_COMPLEX__` because they don't support complex types. It would be nice to have this header accommodate that, but since our chibicc fork isn't using isystem right now, I don't see why that should be a blocker. Thanks for your contribution!

PR #131, opened 2021-03-19
Added macro to alias eai2str to gai_strerror

fabriziobertocci:
Apparently for Linux the correct name of the function to return a string form of the error returned by the `getaddrinfo` is called `gai_strerror` and not `eai2str` (see [man page](https://linux.die.net/man/3/gai_strerror) ).

I have just added a `#define` to alias `eai2str` to `gai_strerror` (since the signature of those functions are the same).

### DIFF
--- a/libc/dns/dns.h
+++ b/libc/dns/dns.h
@@ -47,9 +47,7 @@ struct addrinfo {
 int getaddrinfo(const char *, const char *, const struct addrinfo *,
                 struct addrinfo **) paramsnonnull((4));
 int freeaddrinfo(struct addrinfo *);
-const char *gai_strerror(int errcode);
-const char *eai2str(int);
-#define gai_strerror        eai2str
+const char *gai_strerror(int);
 int dnsnamecmp(const char *, const char *) paramsnonnull();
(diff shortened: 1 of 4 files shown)


jart (changes requested):
Thanks for discovering this. Could we just refactor the code to use the standard name?

```
$ grep -R eai2str
tool/build/runit.c:    FATALF("%s:%hu: EAI_%s %m", g_hostname, g_runitdport, eai2str(rc));
tool/net/dig.c:      fprintf(stderr, "getaddrinfo failed: %d (EAI_%s)\n", rc, eai2str(rc));
libc/dns/dns.h:const char *eai2str(int);
libc/dns/eai2str.c:const char *eai2str(int code) {
```

fabriziobertocci:
oh absolutely, it should be an easy change as I did a grep to see where
that function is being used (there are not that many places)....
Fab

fabriziobertocci:
ok done, I have also renamed the file `libc/dns/eai2str.c` to `libc/dns/gai_strerror.c` to maintain the same pattern function name = file name.

jart (approved):
Looks good! Thank you!

PR #132, opened 2021-03-21
Added prototype for srandom(). 

fabriziobertocci:
Function is already defined in file random.c (together with the other random functions like initstate, setstate, random) but not exposed in the header file


jart (approved):
Looks good. Thanks for spotting this!

PR #142, opened 2021-03-30
Add missing syscalls

### LINKED ISSUE #135: Missing syscalls
I did some digging around in glibc's `unistd.h` file and found a bunch of syscalls (of value) that are not declared in the `cosmopolitan.h` header:

* ftruncate64
* fpathconf
* encrypt
* ctermid
* cuserid
* crypt
* confstr
* getgroups
* setgroups
* gethostid
* gethostname
* getlogin
* getlogin_r
* lockf
* lockf64
* pathconf
* truncate64
* swab
* setpgrp
* lchown

In addition, there are these ones, which have either been removed or are platform-specific. Should we care about them?

* ttyslot (removed from POSIX 2001)
* getpagesize (removed from posix 2001)
* getpass  (removed from posix 2001)
* getwd (removed from POSIX 2008)
* ualarm (removed from POSIX 2008)
* syncfs ("linux specific")
* vhangup ("linux specific")
* getusershell (BSD)
* getdtablesize (BSD)
* sethostname (BSD)
* sethostid (BSD)
* endusershell (BSD)
* acct (SVr4, 4.3BSD)

dosisod:
Closes #135 

After doing some more digging, I found that a lot of these syscalls are actually defined, but do not have function definitions.

I put these in the `runtime.h` file, but there might be a more apt place to put them.


jart (approved):

jart:
Thanks! I think `runtime.h` should be fine.

PR #143, opened 2021-04-01
Added missing definition for struct ip_mreq

fabriziobertocci:
`struct ip_mreq` is used by the `setsockopt()` functions `IP_ADD_MEMBERSHIP` and `IP_DROP_MEMBERSHIP`.
These functions appears to be supported by Cosmopolitan (the constants are defined), although I have not tested it yet.
The `struct ip_mreq` definition was simply missing.

### DIFF
--- a/libc/sock/sock.h
+++ b/libc/sock/sock.h
@@ -48,6 +48,12 @@ struct sockaddr_storage {
   };
 };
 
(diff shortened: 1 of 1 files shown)


jart (approved):
Looks good. Thank you!

PR #153, opened 2021-04-16
Added OpenBrowser function, opens browser in Windows, GNU and XNU

### LINKED ISSUE #98: Suggestion: Launch browser when opening binary
If this is possible then it would add to the portability of redbean and minimize distribution issues with end users.

Rotario:
Added a simple OpenBrowser function that runs just before RedBean begins processing requests. 
Fixes #98 
Maybe this should be added to be optionally turned off, it would be quite annoying if you're just hosting a socket and don't need GUI. I think it should probably default to on though, because then a double click on the executable would open your browser, rather than clunkily opening CMD and passing the argument.
I'm not much of a C programmer (at all!) so feedback is always appreciated!
Thanks

### DIFF
--- a/tool/net/redbean.c
+++ b/tool/net/redbean.c
@@ -319,7 +319,6 @@ static bool terminated;
 static bool uniprocess;
 static bool invalidated;
 static bool logmessages;
-static bool launchbrowser;
 static bool checkedmethod;
 static bool connectionclose;
 static bool keyboardinterrupt;
@@ -619,8 +618,7 @@ static void GetOpts(int argc, char *argv[]) {
         logmessages = true;
         break;
       case 'b':
-        launchbrowser = true;
-        // logbodies = true;
+        logbodies = true;
         break;
       case 'z':
         printport = true;
@@ -1473,8 +1471,10 @@ static void LaunchBrowser() {
   } else {
     prog = "xdg-open";
   }
-  snprintf(openbrowsercommand, sizeof(openbrowsercommand),
-           "%s http://127.0.0.1:%d", prog, ntohs(serveraddr.sin_port));
+  struct in_addr addr = serveraddr.sin_addr;
+  if (addr.s_addr == INADDR_ANY) addr.s_addr = htonl(INADDR_LOOPBACK);
+  snprintf(openbrowsercommand, sizeof(openbrowsercommand), "%s http://%s:%d",
+           prog, inet_ntoa(addr), ntohs(serveraddr.sin_port));
   DEBUGF("Opening browser with command %s\n", openbrowsercommand);
   system(openbrowsercommand);
 }
@@ -2655,7 +2655,6 @@ void RedBean(int argc, char *argv[]) {
   inbuf.p = xvalloc(inbuf.n);
   hdrbuf.n = 4 * 1024;
   hdrbuf.p = xvalloc(hdrbuf.n);
-  if (launchbrowser) LaunchBrowser();
   while (!terminated) {
     if (zombied) {
       ReapZombies();
--- a/tool/net/redbean.c
+++ b/tool/net/redbean.c
@@ -2579,6 +2579,19 @@ static void TuneServerSocket(void) {
   LOGIFNEG1(setsockopt(server, IPPROTO_TCP, TCP_QUICKACK, &yes, sizeof(yes)));
 }
 
+static void OpenBrowser(const char *serveraddrname) {
+  char openbrowsercommand[255];
+  if (IsWindows()){
+    snprintf(openbrowsercommand, sizeof(openbrowsercommand), "explorer http://%s", serveraddrname);
+  } else if (IsXnu()) {
+    snprintf(openbrowsercommand, sizeof(openbrowsercommand), "open http://%s", serveraddrname);
+  } else {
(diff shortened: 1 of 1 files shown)


fredericosilva:
I've seen the command "start" being used for that purpose on windows. Do you know how it differs from explorer?

tool/net/redbean.c
```diff
@@ -2631,6 +2647,8 @@ void RedBean(int argc, char *argv[]) {
   inbuf.p = xvalloc(inbuf.n);
   hdrbuf.n = 4 * 1024;
   hdrbuf.p = xvalloc(hdrbuf.n);
+  // TODO: Maybe make this an optional argv?
```
jart, line 2650: I'd recommend `-b`. All you have to do is add it to the `getopt()` invocation as well as the man page string near the top of the redbean.c file. We need it because redbean is designed to be automated. I've been working on unit tests that launch it as a subprocess within a unit test, using the `redbean.com -zp0` flags which bind to a random port and then prints it to stdout.

Here's the work in progress on the integration test: https://github.com/jart/cosmopolitan/blob/master/test/tool/net/redbean_test.c Most of the other tests for redbean are stored in https://github.com/jart/cosmopolitan/tree/master/test/net/http right now.

tool/net/redbean.c
```diff
@@ -2579,6 +2579,22 @@ static void TuneServerSocket(void) {
...
+  char openbrowsercommand[255];
+  char *prog;
+  if (IsWindows()) {
+    prog = "explorer";
+  } else if (IsXnu()) {
+    prog = "open";
+  } else {
+    prog = "xdg-open";
+  }
+  snprintf(openbrowsercommand, sizeof(openbrowsercommand), "%s http://%s", prog,
```
jart, line 2592: One issue with `serveraddrname` is it might be `0.0.0.0:8080`. What I'd recommend doing is always use 127.0.0.1 and then the port can be obtained via `ntohs(serveraddr.sin_port)`.

jart (changes requested):
Thanks for sending this!

tool/net/redbean.c
jart: Another thing you might consider is adding a Lua API that plugs into your launcher function. That way, if someone wants to distribute a redbean that doesn't need the `-b` flag, they can just put something like `LaunchBrowser()` in their `/tool/net/.init.lua` file which would make launching the browser the default behavior.

Rotario: That's a really good idea, I think having to pass the -b flag is a little cumbersome, as it would be nice to just "double click" the exe (which obviously won't pass any args) but still have a browser launch, so adding the lua init would fix that.

tool/net/redbean.c
Rotario: Also a note on 0.0.0.0 is that yeah that is serveraddrstr resolves to, but the server still works with 0.0.0.0 as well as 127.0.0.1. After doing some reading - 0.0.0.0 is "every local ip on this machine" whereas 127.0.0.1 is just "loopback" or localhost. Should I still change it to 127.0.0.1 for functional reasons? or is it just a cosmetic thing.

Rotario:
I've added Lua API and -b flag now

tool/net/redbean.c
jart: Oh wow you're correct. I knew `0.0.0.0` would `bind()` to all local interfaces. I never considered it would work on the client side with `connect()`. Linux and FreeBSD appear to explicitly define `connect(0.0.0.0)` as `connect(127.0.0.1)` but that does not appear to be the case in all the os / browser / client combinations I've tried. I also just realized the user might explicitly bind to `192.168.0.x` or something. So the right thing to do here might be:

```c
  uint32_t addr;
  char ip4buf[16];
  addr = serveraddr.sin_addr.s_addr;
  if (ntohl(addr) == INADDR_ANY) addr = htonl(INADDR_LOOPBACK);
  snprintf(openbrowsercommand, sizeof(openbrowsercommand), "%s http://%s:%d\n",
           prog, inet_ntop(AF_INET, &addr, ip4buf, sizeof(ip4buf)),
           ntohs(serveraddr.sin_port));
```

tool/net/redbean.c
jart: Oh right that's true. Judging by your reaction, I think you're instinct is right. The Lua API alone seems perfect. Let's hold off on the flag unless someone explicitly requests it.

tool/net/redbean.c
Rotario: Alright, I've added this new code, removed the -b flags and tested multiple different `-l` flags (127.0.0.1 and 127.0.0.2) on Ubuntu 20.04, browser launches at the right addresses.
I guess this needs testing on windows and BSD etc?

Rotario:
Alright, I've added this new code, removed the -b flags and tested multiple different `-l` flags (127.0.0.1 and 127.0.0.2) on Ubuntu 20.04, browser launches at the right addresses.
I guess this needs testing on windows and BSD etc?

jart (approved):
Looks good! Don't worry too much about testing different platforms. The only requirement for contributors is that automated Travis tests are green. I'm happy to carry the burden the rest of the way on your behalf, since I've got a pretty good automated environment running on my local network that I haven't integrated into GitHub workflows yet.

PR #168, opened 2021-05-17
Add ISO 646 header

dosisod:
From [CPP Reference](https://en.cppreference.com/w/c/language/operator_alternative).

I read the contribution guide, and since this is a header-only patch, do I have to do anything else regarding copyright? I would be fine with just [unlicense](https://unlicense.org/)-ing this patch/future patches.


jart (approved):
Looks good. Thank you!

No legal requirements needed for this change.

PR #172, opened 2021-05-23
Added getnameinfo

ahgamut:
implementation of getnameinfo is similar to getaddrinfo, with additional internal functions. I tested it with an example program on `1.1.1.1` and `8.8.4.4` and the output matches glibc.

Two helper functions are currently stub:

When performing an rDNS lookup, I have to look through `/etc/hosts` first, but `struct HostsTxt` is already sorted according to *name*. Should I do a linear search or re-sort it according to address (and then sort back)?

The service name is currently mapped to `0` always. I'll try to add a `struct ServicesTxt` similar to `struct HostsTxt` if I figure it out, or just a simple parse of `/etc/services`.

### DIFF
--- a/libc/dns/consts.h
+++ b/libc/dns/consts.h
@@ -3,7 +3,9 @@
 #include "libc/sock/sock.h"
 #if !(__ASSEMBLER__ + __LINKER__ + 0)
 
-#define DNS_TYPE_A 1
+#define DNS_TYPE_A   0x01
+#define DNS_TYPE_PTR 0x0c
+
 #define DNS_CLASS_IN 1
 
 #define kMinSockaddr4Size \
--- a/libc/dns/dns.h
+++ b/libc/dns/dns.h
@@ -1,5 +1,6 @@
 #ifndef COSMOPOLITAN_LIBC_DNS_DNS_H_
 #define COSMOPOLITAN_LIBC_DNS_DNS_H_
+#include "libc/calls/weirdtypes.h"
 #include "libc/dns/resolvconf.h"
 #include "libc/sock/sock.h"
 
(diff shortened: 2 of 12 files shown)


jart (changes requested):
Thanks for sending this! The /etc/services stuff looks orthogonal to this change. Could we remove that and do it as a separate PR if you need it?

ahgamut:
Ok, I'll look to add a separate PR for parsing `/etc/services` later (maybe along with the `struct servent` stuff).

What do you think about `ResolveHostsReverse`? Should I implement that in this PR or leave it as a stub?

jart:
getnameinfo is a reverse lookup function so that other function seems on-topic.

ahgamut:
Removed the services lookup stub. 

I used the HostsTxt internal structure for the address -> name lookup, so the logarithmic-time lookup of names is lost if one keeps alternating between name and address lookups.

jart:
We can always make it bidirectional if performance ends up being an issue for someone's use case. Will review as soon as I get a chance.

jart (approved):
LGTM. Thank you!

ahgamut:
I tested getnameinfo on the system glibc and looks like it performs a linear lookup: if I have multiple entries with the same IP, glibc picks the first one that matches, but this implementation picks a different one because it starts at the middle.

jart:
You'd have to replace `qsort` with a stable sorting algorithm and then replace `bsearch` with the leftmost variant of binary search. I'd recommend doing brute force linear since we're only comparing on uint32 ip addresses.

ahgamut:
I'll submit a PR that uses just brute-force linear search for `ResolveHostsReverse`.

Even then, the `getnameinfo` output would differ between glibc and cosmopolitan when there are multiple entries mapped to the same IP address. Consider the below `/etc/hosts` file:

```bash
localhost    127.0.0.1
ahost        127.0.0.1
bhost        127.0.0.1
mhost        127.0.0.1
xhost        127.0.0.1
yhost        127.0.0.1
zhost        127.0.0.1
```

Now if I call `getnameinfo` to find the result for `127.0.0.1`, `localhost` is the expected return value, but I will get either `ahost` or `mhost` (depending on where the search starts) because `SortHostsTxt` has sorted the entries by name, and `localhost` is no longer first.

PR #183, opened 2021-06-05
Added wcstok and wmemcmp

ahgamut:
wcstok similar to strtok_r
wmemcmp just calls wcsncmp (same functionality)


jart (approved):
Thank you!

ahgamut:
I found `wmemcmp` because the header said it was available, but the linker said it was missing. 
I know `forkpty` is another such function that has a declaration, but no stub.

Is there a list of such functions (declared in header, but no implementation/stub) so that I can keep track?

PR #185, opened 2021-06-06
Add sqlite3 support to Lua scripts in Redbean

pkulchenko:
Opening this PR per discussion in https://github.com/jart/cosmopolitan/issues/97#issuecomment-855186562. I sent a separate email with copyright assignment.

### DIFF
--- a/tool/net/lsqlite3.c
+++ b/tool/net/lsqlite3.c
@@ -1,0 +1,2408 @@
+/************************************************************************
+* lsqlite3                                                              *
+* Copyright (C) 2002-2016 Tiago Dionizio, Doug Currie                   *
+* All rights reserved.                                                  *
+* Author    : Tiago Dionizio <tiago.dionizio@ist.utl.pt>                *
+* Author    : Doug Currie <doug.currie@alum.mit.edu>                    *
+* Library   : lsqlite3 - an SQLite 3 database binding for Lua 5         *
+*                                                                       *
+* Permission is hereby granted, free of charge, to any person obtaining *
+* a copy of this software and associated documentation files (the       *
+* "Software"), to deal in the Software without restriction, including   *
+* without limitation the rights to use, copy, modify, merge, publish,   *
+* distribute, sublicense, and/or sell copies of the Software, and to    *
+* permit persons to whom the Software is furnished to do so, subject to *
+* the following conditions:                                             *
+*                                                                       *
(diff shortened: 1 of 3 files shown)


jart:
Good execution on this change. Thank you for your contribution. I'll push a new release of redbean to the website as soon as I have a chance.

mskyttner:
Exciting work! I don't have deep knowledge of the codebase, but still dare to ask a couple of questions:

- It is very nice that inprocess sqlite3 persistence becomes part of redbean and sqlite3 has a (by default not compressed, I believe) storage format that is platform independent. With some of the (optional but widely used) sqlite3 extensions (such as the Full Text Search extension, FTS5), it could do a lot of what most need from a database. Including functioning as a "no-nosql" database with BLOBs. I would maybe even be the preferred choice for many. What will be the included extensions to sqlite3 embedded in redbean?
- For someone creating an actually portable executable with redbean and utilizing lua for dynamic server pages, where would/should database files be placed? And if such a web app would persist data not in memory but on disk, where would files be written?

mskyttner:
Here is a demo from duckdb (similar to sqlite3 in many ways) with the TPC-H database: 

https://duckdb.org/demo/

The webapp source code is here:

https://github.com/duckdb/duckdb-web/tree/master/demo

What would be an outline of steps needed to do a similar demo with redbean and the TPC-H dataset?

pkulchenko:
> What will be the included extensions to sqlite3 embedded in redbean?

@mskyttner, it looks like the following extensions are included (judging by the source code; I haven't tested any of them): FTS3, FTS5, JSON1, STMT, and RTree.

> For someone creating an actually portable executable with redbean and utilizing lua for dynamic server pages, where would/should database files be placed?

If the DB is read-only, my understanding that it can either be placed inside the redbean archive itself (with no compression) or outside of it.

> And if such a web app would persist data not in memory but on disk, where would files be written?

The DB file would need to be placed in a "protected" directory (similar to how the lua files are placed). I should be able to test that later today.

pkulchenko:
> For someone creating an actually portable executable with redbean and utilizing lua for dynamic server pages, where would/should database files be placed? And if such a web app would persist data not in memory but on disk, where would files be written?

@mskyttner, I confirmed that when the DB file is referenced with a relative path (like `"test.db"`), it's created in the same directory as redbean itself (although I haven't tested the daemonized configuration).

PR #187, opened 2021-06-08
Add missing $MODE reference to lsqlite in redbean makefile (ref #185).

pkulchenko:
This fixes compiling SQLite support in redbean with non-default MODE configuration.


jart:
Thank you!

Issue #189, opened 2021-06-09
What about including InfoZip?


acailly:
This is a real question, I don't know enough to say if it is a good idea or a completely dumb one :smile: 

This project is marvel: crossplatform web server, include Lua, is a self modifying archive and I saw that you plan to include Sqlite :heart_eyes: 

The part I struggled the most in the beginning is the archive part (see #184), and my conclusion was: use InfoZip, it works every time :tada: 
So now I can get a redbean exe, put some assets in it and here is my crossplatform webapp
But if I want to develop on several env, say Linux and Windows, I have to get InfoZip exe on each platform

So I thought maybe it is possible to build InfoZip from source?

And maybe it is possible to go further, and this is where I have no idea if or how it it feasible, what about including InfoZip so that we can do something like `redbean.com zip redbean.com hello.html` ?

jart:
It's totally doable to make an APE binary out of InfoZip's `zip` and `unzip` commands. It could be as simple as saying `gcc *.c ...`  in the appropriate InfoZip folder, provided of course you pass the recommended flags in the README, as well as `-isystem cosmopolitan/libc/isystem` for compatibility. That's using the amalgamated release.

Another option is this project has a third_party folder. If you go through the commit history, and look at the process we've followed in the past where we've added other projects using the two-commit style, then I would accept a pull request that does that for InfoZip. That will give people an even easier path to building their own APE InfoZip. They would just git clone cosmopolitan and then `make o//third_party/infozip/zip.com`. 

Please note, if we do that, I'm not interested at this time in distributing the zip.com binary myself on the website, since there currently isn't an intent to be in the distributor business. I would prefer that we convince the InfoZip authors of APE's merits and help them improve their build process so that they can be the ones publishing that zip.com binary.

jart:
As for `redbean.com zip redbean.com hello.html` one Lua function you might find interesting is [`StoreAsset`](https://justine.lol/redbean/#StoreAsset) which effectively does the same thing as the `zip` command, except with POSIX advisory locks, rather than atomic replacement.

acailly:
Thank you for your answer @jart 
I really like the pragmatic approach you take, not willing to make it an all purpose project impossible to maintain :smiley: 

If I find the time to do so, I will try to make an APE binary of InfoZip, but I'm not fluent with development in C

Regarding `StoreAssets`, I've seen it but since it is `currently only supported on Linux, XNU, and FreeBSD` that doesn't work if I have one of my dev env on Windows

[...]

PR #191, opened 2021-06-10
Info-ZIP's zip utility in third_party

ahgamut:
As discussed in #189, zip utility from Info-ZIP:

http://infozip.sourceforge.net/Zip.html

source code obtained as `zip30.tar.gz` from:

https://sourceforge.net/projects/infozip/files/

First commit is just all the files extraced from `zip30.tar.gz`, second commit adds build config.

`infozip.mk` is based on `unix/Makefile` in the `zip` folder. Headers are not checked during the build, only executables are produced, no static libs.


third_party/infozip/infozip.mk
```diff
@@ -0,0 +1,134 @@
...
+	$(THIRD_PARTY_ZIP_UTIL_OBJS1)					\
+	$(THIRD_PARTY_ZIP_UTIL_OBJS2)
+
+THIRD_PARTY_ZIP_LARGE_OBJS = \
+	o/$(MODE)/third_party/infozip/zip/zip.o			\
+	o/$(MODE)/third_party/infozip/zip/zipsplit.o	\
+	o/$(MODE)/third_party/infozip/zip/fileio.o		\
+	o/$(MODE)/third_party/infozip/zip/fileio_.o		
+
+o/$(MODE)/third_party/infozip/zip/%_.o: third_party/infozip/zip/%.c
```
ahgamut, line 72: The `infozip.mk` recipe has the same file (for example `fileio.c`) compiled in two different ways: an extra `-DUTIL` flag is passed for `fileio_.o` but not for `fileio.o`. Does something like this happen in any other Makefile in the repo?

The workaround I used here was to create a special rule for the `%_.o` targets in this folder based on the default `%.o` rule.

Alternatively, I can just create copies of the source files (for example, `fileio_.c` as an exact copy of `fileio.c`) which would avoid the wildcard rule above, because I can just add `-DUTIL` to `OVERRIDE_CFLAGS` for the `_.o` files. @jart thoughts?

jart: It does happen. What you did here is fine. Another technique I normally use is this:

```make
$(THIRD_PARTY_MBEDTLS_A_OBJS):                                          \
                        OVERRIDE_CFLAGS +=                              \
                                -fdata-sections                         \
                                -ffunction-sections
```

jart (approved):
Thank you!

third_party/infozip/infozip.mk
ahgamut: The problem I was running into was:

* `fileio.c` is part of the `zip` source code, `fileio_.c` is not
* `fileio.o` is a target that requires `fileio.c` to be compiled 
* `fileio_.o` is a target that requires the same `fileio.c` to be compiled with `-DUTIL` 
* the `%.o` rule in `build/rules.mk` expects `fileio_.c` to exist for the `fileio_.o` target, so adding `OVERRIDE_CFLAGS` alone does not work unless I create the missing `fileio_.c`

I think I have not properly understood how Makefile targets work; I shall lookup the docs and try some examples later.

PR #196, opened 2021-06-21
make getnameinfo simpler

ahgamut:
The getnameinfo implementation requires an address -> name lookup on the
hosts file (ie struct `HostsTxt`) and the previous implementation used
flags to check whether `HostsTxt` was sorted according to address or name,
and then re-sorted it if necessary. Now getnameinfo lookup does not
require sorting, it does a simple linear lookup. The sorting function for `HostsTxt` 
and the related tests) are reverted to what they were in f721d61b54f9b5dab6743edc289f185869363255.

See #172 for discussion.


jart (approved):
Thank you!

PR #197, opened 2021-06-23
Implemented support for ioctl for SIOCGIFxxx functions to enumerate network interfaces

fabriziobertocci:
All right, this is getting there. 
Implemented support for the following ioctl() operations:
* SIOCGIFCONFIG: reads and enumerate all the network interfaces
* SIOCGIFADDR: reads network address for a given interface
* SIOCGIFFLAGS: reads network flags for a given interface
* SIOCGIFNETMASK: reads network netmask for a given interface
* SIOCGIFBRDADDR: reads network broadcast address for a given interface
* SIOCGIFDSTADDR: reads peer destination address for a given interface (not supported for Windows)

The proposed change contains wrappers around all the above functionalities for BSD and Windows to adjust the data structures to the Linux one (the standard interface used here).


libc/sock/internal.h
```diff
@@ -53,6 +53,45 @@ struct sockaddr_un_bsd {
...
+    struct  sockaddr_bsd ifru_netmask;
+    struct  sockaddr_bsd ifru_dstaddr;
+    struct  sockaddr_bsd ifru_broadaddr;
+    short   ifru_flags;
+    char    ifru_pad[16];   /* used as padding */
+  } ifr_ifru;
+};
+
+#pragma pack(4)
+struct ifconf_bsd {
```
jart, line 82: I hadn't heard of pragma pack before. Other ways of doing it are `_Alignas(4)` on the field or `forcealign(4)`. It's also worth noting that XNU is the only BSD-like system that has this peculiarity.  I'll see if I can iterate on this locally since I've got a bunch of VMs set up.

libc/sock/socket-nt.c
```diff
@@ -19,12 +19,19 @@
...
 #include "libc/sock/internal.h"
 #include "libc/sock/yoink.inc"
 #include "libc/sysv/consts/fio.h"
 #include "libc/sysv/consts/o.h"
 #include "libc/sysv/consts/sock.h"
 
+/* ioctl(SIOCGIFCONFIG) for Windows need to access the following functions through 
+ * weak reference. This ensure those symbols are not stripped during final link
+ */
+STATIC_YOINK("GetAdaptersAddresses");
```
jart, line 32: The optimal place for these yoinks might actually be in the `__IOCTL_DISPATCH()` macro, similar to the hack we're using for `__SIGACTION_YOINK()`. It also technically shouldn't be a problem for code in `LIBC_CALLS` to depend on `LIBC_NT_FOO` or `LIBC_STR` since those packages are higher up in the topological order of things and would not create a cyclic dependency.

jart (approved):
LGTM. This is quite a change and I'm looking forward to using it. I've left some high level feedback. I should have a change out later today that iterates on this and makes sure everything works on OpenBSD / FreeBSD / NetBSD too.

fabriziobertocci:
Hi @jart  yes, it's quite a change, that's why it took me so long (especially the Windows support was very time consuming because their API is so fragmented... you can do it with WSAIoctl, or with GetAdaptersAddresses, or ... ).
Anyway, please test it on various BSD variations because I do not have any machine to test the changes. I have tested only on Linux (Mint 19.3/Ubuntu 18.04), MacOS (Mojave), and Windows 10.

Regarding the location of the `STATIC_YOINK`, feel free to move it. You have way better visibility on the project than me.

There are other things that we should probably discuss, some aspects that may be debatable and can be optimized.
Yesterday I was thinking that the current algorithm to find a unique name for the network interface on Windows is naive to say the least (it seems to be working, as I tested it with my Windows 10 VM with 4 physical network with multiple aliases, with name clashing..., but it's not efficient nor elegant).
But again, those kind of ioctl() will never be in a critical path of an application, probably 99% of the time they will be used during the initialization of an app, so efficiency is not a requirement.

Anyway, I'll be happy to continue the discussion, just go through the code and comment it. I can submit new PR to improve it.

jart:
I'm going to help you improve the code. How about I submit a pull request
to the project and invite you to review it?

jart:
I'm glad we got your code into the codebase so we can start going back and
forth with real collaboration. One thing I'm noticing that I wish I could
have been there working with you in real life, since this is the kind of
thing that could have been easily pointed out, is the importance of the
libc/sysv/consts.sh file. Many of the constants hold placeholder zeros on
Windows because no one has had time to give them definitions. I can imagine
that being a source of pain when writing a change like this.

fabriziobertocci:
Hi Justine,
Yes we can definitely improve the code. As I said you know the internals
way better than anyone else. I have started too many projects myself and
brought in collaborators and it's always a pain point to transfer all I had
in my brain to others.
Let me know how this collaboration works... or I can figure it out... not a
big deal.
About the libc/sysv/consts.sh.
What constants are you referring to? All those #defines and enums that are
in the iphlpapi.h ?
I thought that the goal of using constant variables instead of macros is so
that the value can be defined at run-time depending on the platform (e.g. a
EAGAIN code could have different values on BSD and Linux)... the reason why
I did not put those constants in there is because at the end those ones
were only used by the Windows implementation.
Anyway, let me know... As usual, I'm happy to help.
Fab

jart:
You use the IFF_* constants in the NT code. For example you commented out
IPP_POINTOPOINT and you probably did that because right now consts.sh
defines it as zero.

Do you know if there's any reason why we can't use OnLinkPrefixLength to
compute the network address, broadcast address, etc?

fabriziobertocci:
See inline:


On Sat, Jun 26, 2021 at 1:22 AM Justine Tunney ***@***.***>
wrote:

> You use the IFF_* constants in the NT code. For example you commented out
> IPP_POINTOPOINT and you probably did that because right now consts.sh
> defines it as zero.
>
I only understood what you meant with this only after reviewing your code
changes. No, the reason why I commented out support for POINTTOPOINT on
Windows is because I couldn't (quickly) figure out which field stores the
peer address in the struct returned by GetAdaptersAddresses.

Do you know if there's any reason why we can't use OnLinkPrefixLength to
> compute the network address, broadcast address, etc?
>
No reason...well, it's mostly a lack of understanding in this huge struct.
There are just too many fields and I didn't notice the OnLinkPrefixLength.
I ran some tests and I saw that the FirstPrefix list had the kind of
information I needed. It's very clumsy the way they organize the
information... 3 list nodes for each unicast address + 2 at the end related
to the adapter itself...
Definitely if OnLInkPrefixLength has the correct value, it'll be easier to
use it for network address/bcast as it won't require to use the
NtIpAdapterPrefix list...
Fab



>
> On Fri, Jun 25, 2021 at 2:41 PM fabriziobertocci ***@***.***>
> wrote:
>
> > Hi Justine,
> > Yes we can definitely improve the code. As I said you know the internals
> > way better than anyone else. I have started too many projects myself and
> > brought in collaborators and it's always a pain point to transfer all I
> had
> > in my brain to others.
> > Let me know how this collaboration works... or I can figure it out...
> not a
> > big deal.
> > About the libc/sysv/consts.sh.
> > What constants are you referring to? All those #defines and enums that
> are
> > in the iphlpapi.h ?
> > I thought that the goal of using constant variables instead of macros is
> so
> > that the value can be defined at run-time depending on the platform
> (e.g. a
> > EAGAIN code could have different values on BSD and Linux)... the reason
> why
> > I did not put those constants in there is because at the end those ones
> > were only used by the Windows implementation.
> > Anyway, let me know... As usual, I'm happy to help.
> > Fab
> >
> > On Fri, Jun 25, 2021 at 8:37 PM Justine Tunney ***@***.***>
> > wrote:
> >
> > > I'm glad we got your code into the codebase so we can start going back
> > and
> > > forth with real collaboration. One thing I'm noticing that I wish I
> could
> > > have been there working with you in real life, since this is the kind
> of
> > > thing that could have been easily pointed out, is the importance of the
> > > libc/sysv/consts.sh file. Many of the constants hold placeholder zeros
> on
> > > Windows because no one has had time to give them definitions. I can
> > imagine
> > > that being a source of pain when writing a change like this.
> > >

jart:
WIN32 is clumsy but it's a well-defined clumsiness and that's something I can respect. I'm glad we managed to spot that CIDR field. It would have been easier if the docs just called it CIDR.

One thing I'm having a hard time figuring out is `ioctl(SIOCGIFBRDADDR)`. Linux seems to always return `0.0.0.0` and no other operating system supports it. They all define it. But they just return -1 `EINVAL`. Is there ever a case where a broadcast address can be something other than `(ip & netmask) | (~netmask & -1)`?

Here's the code I've been using:

```c
TEST(ioctl_siocgifconf, test) {
  size_t n;
  char *data;
  int socketfd;
  struct ifreq *ifr;
  struct ifconf conf;
  char addrbuf[1024];
  uint32_t ip, netmask, broadcast;
  bool foundloopback = false;
  data = gc(malloc((n = 4096)));
  ASSERT_NE(-1, (socketfd = socket(AF_INET, SOCK_DGRAM, 0)));
  conf.ifc_buf = data;
  conf.ifc_len = n;
  ASSERT_NE(-1, ioctl(socketfd, SIOCGIFCONF, &conf));
  for (ifr = (struct ifreq *)data; (char *)ifr < data + conf.ifc_len; ++ifr) {
    if (ifr->ifr_addr.sa_family != AF_INET) continue;
    ip = ntohl(((struct sockaddr_in *)&ifr->ifr_addr)->sin_addr.s_addr);
    EXPECT_NE(-1, ioctl(socketfd, SIOCGIFNETMASK, ifr));
    netmask = ntohl(((struct sockaddr_in *)&ifr->ifr_addr)->sin_addr.s_addr);
    EXPECT_NE(-1, ioctl(socketfd, SIOCGIFBRDADDR, ifr));
    broadcast = ntohl(((struct sockaddr_in *)&ifr->ifr_addr)->sin_addr.s_addr);
#if 1
    fprintf(stderr,
            "%s\n"
            "  ip      %hhu.%hhu.%hhu.%hhu\n"
            "  netmask %hhu.%hhu.%hhu.%hhu\n"
            "  bcast   %hhu.%hhu.%hhu.%hhu\n"
            "\n",
            ifr->ifr_name, ip >> 24, ip >> 16, ip >> 8, ip, netmask >> 24,
            netmask >> 16, netmask >> 8, netmask, broadcast >> 24,
            broadcast >> 16, broadcast >> 8, broadcast);
#endif
    if (ip == 0x7f000001) {
      foundloopback = true;
      EXPECT_EQ(0xFF000000, netmask);
    }
  }
  EXPECT_TRUE(foundloopback);
  ASSERT_NE(-1, close(socketfd));
}
```

PR #200, opened 2021-07-01
add struct hostent and related functions

ahgamut:
`gethostbyname` internally calls `getaddrinfo` to fill out the `struct hostent` details.
`gethostbyaddr` internally calls `getnameinfo` to fill out the `struct hostent` details.

`sethostent`, `gethostent`, and `endhostent` are stubs in the [musl source](https://github.com/ifduyue/musl/blob/master/src/network/ent.c), so stubbed here as well.

also added `h_errno`, `herror`, and `h_strerror` for completeness.


jart (approved):
LGTM Thank you!

PR #209, opened 2021-07-06
Add protoent and netent

ahgamut:
`struct netent` related functions are stubs in the `musl` source code (see [here](https://github.com/ifduyue/musl/blob/master/src/network/netname.c) and [here](https://github.com/ifduyue/musl/blob/master/src/network/ent.c)), so left them as stubs here.

`struct protoent` related functions follow from `struct servent`: the function names, calls, returns, errors, and tests are very similar.  Of course, `/etc/protocols` is read instead of `/etc/services`.

The protocols file on my Windows 10 system is called `<location>\\etc\\protocol`, not `protocols`, so I assumed that must be the case for Windows in general.

The [`man` page entry for `getprotobyname`](https://linux.die.net/man/3/getprotobyname)  does not specify case-sensitive comparisons, so the implementation uses `strcasecmp` for comparing names/aliases.

`getprotobyname`/`getprotobynumber` calls output matches glibc, except for the `protoent.p_aliases`.

Surprisingly the [`musl` implementations of `getprotoby*`](https://github.com/ifduyue/musl/blob/master/src/network/proto.c) do not read `/etc/protocols`, there's just a hardcoded list of protocols in the source code.

### DIFF
--- a/libc/dns/getaddrinfo.c
+++ b/libc/dns/getaddrinfo.c
@@ -59,6 +59,7 @@ int getaddrinfo(const char *name, const char *service,
         (hints && (hints->ai_flags & AI_PASSIVE) == AI_PASSIVE)
             ? INADDR_ANY
             : INADDR_LOOPBACK;
+    *res = ai;
     return 0;
   }
   if (inet_pton(AF_INET, name, &ai->ai_addr4->sin_addr.s_addr) == 1) {
--- a/libc/dns/getaddrinfo.c
+++ b/libc/dns/getaddrinfo.c
@@ -53,7 +53,7 @@ int getaddrinfo(const char *name, const char *service,
 
   if (!name && !service) return EAI_NONAME;
   if (!name && (hints->ai_flags & AI_CANONNAME)) return EAI_BADFLAGS;
-  if (service) {
+  if (service && (port = parseport(service)) == -1) {
     if (hints->ai_socktype == SOCK_STREAM)
       strcpy(proto, "tcp");
     else if (hints->ai_socktype == SOCK_DGRAM)
@@ -62,9 +62,8 @@ int getaddrinfo(const char *name, const char *service,
       strcpy(proto, "");
 
     if ((port = LookupServicesByName(service, proto, sizeof(proto), NULL, 0,
-                                     NULL)) == -1) {
-      if ((port = parseport(service)) == -1) return EAI_NONAME;
-    }
+                                     NULL)) == -1)
+      return EAI_NONAME;
   }
   if (!(ai = newaddrinfo(port))) return EAI_MEMORY;
   if (service) ai->ai_addr4->sin_port = htons(port);
--- a/libc/dns/getaddrinfo.c
+++ b/libc/dns/getaddrinfo.c
@@ -55,11 +55,11 @@ int getaddrinfo(const char *name, const char *service,
   if (!name && (hints->ai_flags & AI_CANONNAME)) return EAI_BADFLAGS;
   if (service) {
     if (hints->ai_socktype == SOCK_STREAM)
-      strncpy(proto, "tcp", sizeof(proto));
+      strcpy(proto, "tcp");
     else if (hints->ai_socktype == SOCK_DGRAM)
-      strncpy(proto, "udp", sizeof(proto));
+      strcpy(proto, "udp");
     else /* ai_socktype == 0 */
-      proto[0] = '\0';
+      strcpy(proto, "");
 
     if ((port = LookupServicesByName(service, proto, sizeof(proto), NULL, 0,
                                      NULL)) == -1) {
--- a/libc/dns/getaddrinfo.c
+++ b/libc/dns/getaddrinfo.c
@@ -21,12 +21,15 @@
 #include "libc/dns/dns.h"
 #include "libc/dns/hoststxt.h"
 #include "libc/dns/resolvconf.h"
+#include "libc/dns/servicestxt.h"
 #include "libc/fmt/conv.h"
 #include "libc/mem/mem.h"
+#include "libc/runtime/gc.h"
 #include "libc/sock/sock.h"
 #include "libc/str/str.h"
 #include "libc/sysv/consts/af.h"
 #include "libc/sysv/consts/inaddr.h"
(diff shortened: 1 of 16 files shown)


ahgamut:
@jart [RFC 1700](https://datatracker.ietf.org/doc/html/rfc1700) says that port number 0 in `/etc/services` is reserved.

When `getaddrinfo` calls `parseport`, `atoi` returns 0 if the input string cannot be converted to a number (for eg `atoi("hello")` is 0), and `parseport` allows `0` as a valid service number, so `getaddrinfo` returns successfully even when called with an invalid service name,  which doesn't match with glibc.

jart:
We should absolutely fix that. `port="0"` is a valid input since that's used to bind to a random o/s assigned port. However getaddrinfo() should fail if the input is something like `"hceurhcreu"` or `"-1"`.

libc/dns/getaddrinfo.c
```diff
@@ -43,11 +46,26 @@ int getaddrinfo(const char *name, const char *service,
...
   struct addrinfo *ai;
+  char proto[32];
   port = 0;
+
   if (!name && !service) return EAI_NONAME;
-  if (service && (port = parseport(service)) == -1) return EAI_NONAME;
   if (!name && (hints->ai_flags & AI_CANONNAME)) return EAI_BADFLAGS;
+  if (service) {
+    if (hints->ai_socktype == SOCK_STREAM)
+      strncpy(proto, "tcp", sizeof(proto));
```
jart, line 58: strncpy() is a misunderstood function. I only know of one legitimate use case for it. Doing binary search with a big endian buffer. For example:

```c
static const struct People {
  char name[8];
  int age;
} kPeople[] = {
    {"alice", 29},  //
    {"bob", 42},    //
};

int GetAge(const char *name) {
  char k[8];
  int m, l, r;
  l = 0;
  r = ARRAYLEN(kPeople) - 1;
  strncpy(k, s, 8);
  while (l <= r) {
    m = (l + r) >> 1;
    if (READ64BE(kPeople[m].s) < READ64BE(k)) {
      l = m + 1;
    } else if (READ64BE(kPeople[m].s) > READ64BE(k)) {
      r = m - 1;
    } else {
      return kPeople[m].age;
    }
  }
  return -1;
}
```

In this case it would be fine to say strcpy() since we know proto's length.

libc/dns/getservbyname.c
```diff
@@ -28,33 +28,30 @@
...
   }
 
-  p = LookupServicesByName(name, &localproto, ptr0->s_name, DNS_NAME_MAX, NULL);
-  if (p == -1) {
-    // localproto got alloc'd during the lookup?
-    if (!proto && localproto != proto) free(localproto);
-    return NULL;
-  }
+  localproto[0] = '\0';
+  if (proto) strncpy(localproto, proto, DNS_NAME_MAX);
```
jart, line 48: This line isn't memory safe. You'd want to do something like this:

```c
if (proto) {
  if (strlen(proto) > DNS_NAME_MAX) return 0;
  strcpy(localproto, proto);
} else {
  strcpy(localproto, "");
}
```

Update other file too please.

libc/dns/prototxt.c
```diff
@@ -0,0 +1,190 @@
...
+  }
+}
+
+/**
+ * Opens and searches /etc/protocols to find name for a given number.
+ *
+ * format of /etc/protocols is like this:
+ *
+ *      # comment
+ *      # NAME      PROTOCOL    ALIASES
```
jart, line 61: Could you clean up the formatting of this indented code example documentation?

libc/dns/prototxt.c
```diff
@@ -0,0 +1,190 @@
...
+    number = strtok_r(NULL, "/ \t\r\n\v", &tok);
+    if (name && number) {
+      alias = name;
+      while (alias && strcasecmp(alias, protoname) != 0)
+        alias = strtok_r(NULL, " \t\r\n\v", &tok);
+
+      if (alias) /* alias matched with protoname */
+      {
+        result = atoi(number);
+        strncpy(buf, name, bufsize);
```
jart, line 174: Consider `strlcpy` since that actually does the thing `strncpy` should have done. `strncpy` doesn't guarantee a NUL terminator so it can't be used to copy a string to a fixed size buffer with silent truncation. It's also possible to use `memccpy` for this purpose. We have a SWAR implementation of `memccpy` since I like that function a lot. However it's got a more cumbersome API since it requires that you check the return value.

https://github.com/jart/cosmopolitan/blob/fe881982b5da8d34999e8fdb13bd62c382391032/libc/str/memccpy.c#L29-L92

libc/dns/servicestxt.c
```diff
@@ -180,15 +188,11 @@ int LookupServicesByName(const char *servname, char **servproto, char *buf,
...
-          result = htons(atoi(port));
-          strncpy(buf, name, bufsize);
-          found = 1;
-        } else if (strcasecmp(proto, servproto[0]) == 0) {
-          result = htons(atoi(port));
-          strncpy(buf, name, bufsize);
+        if (!servproto[0] || strncasecmp(proto, servproto, servprotolen) == 0) {
+          result = atoi(port);
           found = 1;
+          if (!servproto[0]) strncpy(servproto, proto, servprotolen);
```
jart, line 194: Please update strncpy here too.

jart (changes requested):

libc/dns/prototxt.c
ahgamut: `memccpy` failure handling is nice: `memccpy` returns NULL on failure, `strlcpy` returns the length copied, which I have to check with something else, and `strncpy` simply returns the destination pointer.

ahgamut:
> We should absolutely fix that. `port="0"` is a valid input since that's used to bind to a random o/s assigned port. However getaddrinfo() should fail if the input is something like `"hceurhcreu"` or `"-1"`.

I updated `parseport` to use `strtoimax` instead of `atoi` (still uses base 10).   
Now `getaddrinfo` returns success for `port="0"`, and correctly fails on the other invalid inputs.

jart (approved):
LGTM. Thanks so much!

PR #210, opened 2021-07-09
Update LuaGetParam (via LuaPushUrlParams) to return empty table

pkulchenko:
when no parameters are present (#97).


pkulchenko:
Ha, this has just been implemented in feb0f9fb. Closing...

PR #213, opened 2021-07-12
Update GetHeaders and GetParams descriptions in redbean help.

pkulchenko:
It's a minor change, but here is the rationale:

1. I find the distinction between table and array confusing here (since the same structure is used in Lua); this may also conflict with the array structure if/when it's introduced
2. The usage of square brackets doesn't allow for its use to mark optional elements; for example, the second description would need to be changed for something like `array[array[name:str[,value:str]]]`, which is ambiguous and confusing to read.
3. Using `=` shows that it's a hash and not an array (while it's actually an array in the second case)

I hope this syntax reads better and is easier to understand.

### DIFF
--- a/tool/net/help.txt
+++ b/tool/net/help.txt
@@ -570,7 +570,7 @@ FUNCTIONS
            message, then this function will fold those multiple entries into
            a single string.
 
-   GetHeaders() → table[name:str,value:str]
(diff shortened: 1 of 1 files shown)


jart (approved):
LGTM

PR #218, opened 2021-07-24
Redbean add fetch with table

pkulchenko:
This adds three changes (all in separate commits):

- Adds ability to supply a table with options as a second parameter (instead of the body). So far only "method" and "body" are recognized, but I plan to add at least "headers" in the future. If the second parameter is a string, then POST request is sent as before.
- Fixes providing Content-Length leader where appropriate. It's skipped for some requests (GET/HEAD/DELETE and few others) when 0
- Accepts arbitrary methods. Non-standard methods are needed for some of the use cases I've been working on. All method names are upper-cased to stay consistent with the standard ones.

### DIFF
--- a/tool/net/redbean.c
+++ b/tool/net/redbean.c
@@ -3702,15 +3702,21 @@ static int LuaFetch(lua_State *L) {
                            .ai_flags = AI_NUMERICSERV};
 
   /*
-   * Get args.
+   * Get args: url [, body | {method = "PUT", body = "..."}]
    */
   urlarg = luaL_checklstring(L, 1, &urlarglen);
-  if (lua_isnoneornil(L, 2)) {
+  if (lua_istable(L, 2)) {
+    lua_settop(L, 2); // discard any extra arguments
+    lua_getfield(L, 2, "body");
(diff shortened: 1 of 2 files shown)


tool/net/redbean.c
```diff
@@ -3686,33 +3686,48 @@ static int LuaFetch(lua_State *L) {
...
     body = luaL_checklstring(L, 2, &bodylen);
-    method = kHttpPost;
+    method = kHttpMethod[kHttpPost];
   }
+  // provide Content-Length header unless it's zero and not expected
+  methodidx = GetHttpMethod(method, -1);
+  if (bodylen > 0 ||
+    !(methodidx == kHttpGet || methodidx == kHttpHead ||
+      methodidx == kHttpTrace || methodidx == kHttpDelete || methodidx == kHttpConnect))
+    conlenhdr = gc(xasprintf("Content-Length: %d\r\n", bodylen));
```
jart, line 3730: Strictly speaking you'd want `%zu` instead of `%d` but it shouldn't be a problem unless you're planning on uploading files greater than 2gb.

jart (approved):
Looks good. Thank you!

PR #220, opened 2021-08-02
Add Sleep to Lua in redbean.

pkulchenko:
Per suggestion in #97. This is something that is used in Lapis and may come handy when need to avoid throttling for http requests.

### DIFF
--- a/tool/net/help.txt
+++ b/tool/net/help.txt
@@ -859,6 +859,10 @@ FUNCTIONS
    Slurp(filename:str) → str
            Reads file data from local file system.
 
+   Sleep(seconds:number)
+           Sleeps the specified number of seconds (can be fractional). The
+           smallest interval is a millisecond.
+
    Route([host:str,[path:str]])
            Instructs redbean to follow the normal HTTP serving path. This
            function is useful when writing an OnHttpRequest handler, since
--- a/tool/net/redbean.c
+++ b/tool/net/redbean.c
@@ -4879,6 +4879,11 @@ static int LuaLog(lua_State *L) {
   return 0;
 }
 
+static int LuaSleep(lua_State *L) {
+  usleep(1000000 * luaL_checknumber(L, 1));
+  return 0;
+}
+
 static int LuaIsHiddenPath(lua_State *L) {
   size_t n;
   const char *s;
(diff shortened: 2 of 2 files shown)


pavel212:
And also as zip is already there, could it be provided to Lua as well?
something like: 

```
int LuaDeflate(lua_State * L){
  size_t s;
  void * d = Deflate(lua_tostring(L, 1), lua_len(L, 1), &s);
  lua_pushlstring(L, d, s);
  free(d);
  return 1;
}

int LuaInflate(lua_State * L){
  size_t s = lua_tointeger(L, 2);
  char * d = xmalloc(s);
  Inflate(d, s, lua_tostring(L, 1), lua_len(L, 1));
  lua_pushlstring(L, d, s);
  free(d);
  return 1;
}
```

probably with a few additional argument sanity checks.

pkulchenko:
@pavel212, what would be the documentation for these two methods (following the same pattern as exists in [help.txt](https://github.com/jart/cosmopolitan/blob/master/tool/net/help.txt)) or the use case?

jart:
Yes we can expose Deflate and Inflate to Lua.

tool/net/redbean.c
```diff
@@ -4879,6 +4879,11 @@ static int LuaLog(lua_State *L) {
   return 0;
 }
 
+static int LuaSleep(lua_State *L) {
+  usleep(1000000 * luaL_checknumber(L, 1));
```
jart, line 4883: You technically could say 1e6 if you wanted to.

tool/net/help.txt
```diff
@@ -859,6 +859,10 @@ FUNCTIONS
    Slurp(filename:str) → str
            Reads file data from local file system.
 
+   Sleep(seconds:number)
+           Sleeps the specified number of seconds (can be fractional). The
+           smallest interval is a millisecond.
```
jart, line 864: Microsecond. If you want to nanosecond granular sleeps you could spinlock on `nowl()` using `asm("pause")` if `seconds < 2e-6`

jart (approved):
Looks good to merge. See comments.

tool/net/help.txt
pkulchenko: Funny typo; will fix in the next update (along with `1e6`).

pkulchenko:
We can surely add Inflate/Deflate, but I'm curious about the use case, as for Inflate the original size has to come from somewhere.

pavel212:
LuaDeflate(data:str) -> compressed_data:str
Compress data

LuaInflate(comressed_data:str, original_size:int) -> data:str
Decompress compressed_data


To store assets outside redbean.com, but compressed, (inside sqlite?).
and yes, original size should be stored additionally, still better than nothing. 

Or does Cosmopolitan libc "zip: magic" work on other archives outside as well, e.g. db = sqlite3.open("somefile.zip/some.db")?

PR #222, opened 2021-08-04
Redbean lua array

pkulchenko:
@jart, this implements lua-array support, as described in the [README](https://github.com/pkulchenko/lua-array/tree/lua-array-5.4.3#readme) with one exception: `table.remove` doesn't shrink the table, so `#tbl` remains the same, even though the element is deleted. All the other code examples should work as described. I've added author's license/copyright statement in exactly the same way as was done for other integrated components (only in one main file though).

I'm not sure what to do with the documentation, but maybe we'll have a section in the main docs, similar to how `RE` and `SQLite` are described.

### DIFF
--- a/third_party/lua/lapi.c
+++ b/third_party/lua/lapi.c
@@ -18,6 +18,7 @@
 #include "third_party/lua/lstate.h"
 #include "third_party/lua/lstring.h"
 #include "third_party/lua/ltable.h"
+#include "larray.h"
 #include "third_party/lua/ltm.h"
 #include "third_party/lua/lua.h"
 #include "third_party/lua/lundump.h"
@@ -309,6 +310,12 @@ LUA_API int lua_isstring (lua_State *L, int idx) {
 LUA_API int lua_isuserdata (lua_State *L, int idx) {
   const TValue *o = index2value(L, idx);
   return (ttisfulluserdata(o) || ttislightuserdata(o));
+}
+
+
+LUA_API int lua_isarray (lua_State *L, int idx) {
+  const TValue *o = index2value(L, idx);
+  return (isvalid(L, o) && ttype(o) == LUA_TTABLE && hvalue(o)->truearray);
 }
 
 
@@ -752,10 +759,46 @@ LUA_API void lua_createtable (lua_State *L, int narray, int nrec) {
   t = luaH_new(L);
   sethvalue2s(L, L->top, t);
   api_incr_top(L);
-  if (narray > 0 || nrec > 0)
+  if (narray > 0 || nrec > 0) {
     luaH_resize(L, t, narray, nrec);
+  }
   luaC_checkGC(L);
   lua_unlock(L);
+}
+
+
+LUA_API void lua_createarray (lua_State *L, int narray) {
+  Array *a;
+  lua_lock(L);
+  a = luaA_new(L);
+  setavalue2s(L, L->top, a);
+  api_incr_top(L);
+  if (narray > 0)
+    luaA_resize(L, a, narray);
+  luaC_checkGC(L);
+  lua_unlock(L);
+}
+
+
+LUA_API void lua_resize (lua_State *L, int idx, unsigned int size) {
+  TValue *o;
+  Table *t;
+  unsigned int i, oldsize;
+  lua_lock(L);
+  o = index2value(L, idx);
+  api_check(L, ttistable(o), "table expected");
+  t = hvalue(o);
+  oldsize = t->sizeused;
+  if (size > luaH_realasize(t)) {
+    luaH_resizearray(L, t, size);
+  }
+  lua_unlock(L);
+  /* set removed elements to nil when shrinking array size */
+  for(i = size + 1; i <= oldsize; i++) {
+    lua_pushnil(L);
+    lua_seti(L, idx, i);
+  }
+  t->sizeused = size;
 }
 
 
@@ -867,7 +910,10 @@ LUA_API void lua_seti (lua_State *L, int idx, lua_Integer n) {
   lua_lock(L);
   api_checknelems(L, 1);
   t = index2value(L, idx);
-  if (luaV_fastgeti(L, t, n, slot)) {
+  if (ttisarray(t)) {
+    luaA_setint(L, avalue(t), n, s2v(L->top - 1));
+  }
+  else if (luaV_fastgeti(L, t, n, slot)) {
     luaV_finishfastset(L, t, slot, s2v(L->top - 1));
   }
   else {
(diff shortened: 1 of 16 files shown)


jart (approved):
Outstanding. It'll be nice to not have O(n) append and len. I'll update the documentation in a follow-up explaining the Lua language extensions we've made.

jart:
Further work is needed on this. If I say `[3]` it creates a table, not an array. I confirmed `#[3]` calls the O(n) code path. It also appears the create array code isn't linked.

```
master jart@nightmare:~/cosmo$ nm o/third_party/lua/lua.com.dbg | grep luaA_new
$?=1 master jart@nightmare:~/cosmo$ nm o/third_party/lua/lua.com.dbg | grep lua_createarray
$?=1 master jart@nightmare:~/cosmo$
```

Things should be fine in the meantime since all the Lua tests still pass.

```
(m=dbg; make -j8 MODE=$m o/$m/third_party/lua/lua.com && cd third_party/lua/test/ && ~/cosmo/o/$m/third_party/lua/lua.com all.lua)
```

pkulchenko:
I'll check; I have all the tests passing for lua-array as well, but I didn't check what code path is being followed for `#[3]` or similar code. I *do* get different results depending on whether I use `{}` or `[]`, so there is clearly some difference, for example, `a = []; a[10] = true; print(#a)` produce different results from `a = {}; a[10] = true; print(#a)`.

pkulchenko:
@jart, I can't reproduce this, as the shortcut path is being followed when `#a` is evaluated: `truearray` should be `true` here: https://github.com/jart/cosmopolitan/blob/master/third_party/lua/ltable.c#L931. createarray is indeed not being used; I should remove all that code (even though it's present in the original).

jart:
To reproduce, add a printf statement here:

https://github.com/jart/cosmopolitan/blob/aeeb85142246b3fc9f4267ccdd88358e163ce118/third_party/lua/lvm.c#L683-L700

Then say:

```
make -j8 o//third_party/lua/lua.com &&
o//third_party/lua/lua.com -e 'print(#[3])'
```

See also:

https://github.com/jart/cosmopolitan/blob/aeeb85142246b3fc9f4267ccdd88358e163ce118/third_party/lua/lparser.c#L941-L948

pkulchenko:
Right, but the shortcut for `truearray` is inside `luaH_getn(h)` call.

jart:
Ah I see. Then it sounds like we're good. I turned on `-ffunction-sections` so if the code I mentioned earlier that's not being linked, is worth keeping around (for the C API), then we should keep it around, since there's no longer any bloat cost to doing so.

pkulchenko:
I still plan to do a clean up to keep the code simple. The included benchmarks show that the speed up for #a is about 30%.

jart:
I might have been wrong earlier about my assumptions regarding lua table complexity. At least in the average case. I'm trying to demonstrate that an improvement of some sort is being made by the Lua array implementation and it seems to use the same amount of cpu and the same if not more memory. Here's some scripts I've whipped up so far:

```lua
n = 2
a1 = os.clock()
while n < 8388608 do
   t1 = os.clock()
   x = {}
   for i = 1,n do
      x[#x+1] = i
   end
   z = #x
   t2 = os.clock()
   print(string.format("n=%d %g", z, t2 - t1))
   n = n * 2
end
a2 = os.clock()
print(string.format("%g", a2 - a1))
```

Also:

```lua
n = 2
a1 = os.clock()*1e6
while n < 8388608*4 do
   x = {}
   for i = 1,n do
      x[math.floor(math.random() * n) + 1] = i
   end
   for i = 1,n do
      x[i] = i
   end
   z = 0
   t1 = os.clock()*1e6
   for i = 1,n/2 do
      z = z + x[math.floor(math.random() * n) + 1]
   end
   t2 = os.clock()*1e6
   print(string.format("n=%d %g", math.log(n, 2), math.log(t2 - t1, 2)))
   n = n * 2
end
a2 = os.clock()*1e6
print(string.format("%g", a2 - a1))
```

pkulchenko:
I'm puzzled a bit myself. It looks like there are two implementations: (1) larray-based one with a special type and some additional handling and (2) table-based one with `truearray` checks. The former doesn't seem to be used, so really only the latter is used, but I was trying to trace the history of lua-array changes to see where the switch was made and couldn't find it.

It does look like Lua 5.4.3 made significant improvements in array handling, as my benchmarks show some improvements, but less drastic difference between the reported results for original and modified code.

In terms of the benchmark, you're not using the advantages of the array, as you're growing almost the same way, as you'd be growing a table. The advantage will be seen when the array is pre-allocated, so try iterating it from the back or setting `n` initially:

```
   x = []
   x[n] = n --<-- commenting out this line shows significant difference for me
   for i = 1,n do
      x[i] = i
   end
```

pkulchenko:
Just to add one clarification about Lua array management. Lua actually has proper arrays as part of its table structure, so if you do iterate and create arrays by assigning elements 1,2,3, etc., the performance is not going to be much different from lua-array, as this is exactly what's being used. The performance *will* be different in the case of random assignment, as if you assign an index beyond the current array size (by more than one), then it will be allocated in the hash part (so, `t[1000]=1000` will go to the hash part if the array is empty). When the hash part is full, the array is going to be resized to figure out a full sequence to populate it.

Given this, the advantage of lua-array will be seen on using pre-allocated arrays (as `a[1000]` allocates array of 1000 elements, but `t[1000]` does no such thing) and calculating the array size. Rewriting already existing/allocated elements has the same performance in both cases (that includes random access for array elements, assuming they don't cause re-allocation).

jart:
The preallocation technique you suggested makes it go faster for `n<512` but for larger sizes the performance becomes identical, since it doesn't change complexity. The Lua C API has a parameter when creating tables for specifying a capacity hint. If performance on small sizes is a concern then we'd probably get the same benefit by just exposing that to Lua. If that's the case then it probably isn't worth changing the language. More. Pythonic syntax would be nice to have. But if Lua was recently fixed to make table ops have the same complexity class as an array then maybe we should stick to that.

jart:
As for heap growth, there's still things I've been meaning to do at a deeper level in the C library for improving its performance. For example I want to use `mremap()` on Linux but haven't had a chance to get around to it yet.

pkulchenko:
> If that's the case then it probably isn't worth changing the language.

Yes, there is a bit of a boost, but may be not sufficient enough (and smaller than I expected) to warrant the associated language changes. Do you want me to pull it out?

pkulchenko:
@jart, here is the revert branch for these changes: https://github.com/pkulchenko/cosmopolitan/tree/lua-array-revert

jart:
Yes please send that change. I'm glad we experimented with improving the language and failed since now we know redbean is even better engineered than I thought it was and that we can provide more value to the user with greater simplicity and certainty.

PR #226, opened 2021-08-06
Add following redirects to redbean Fetch.

pkulchenko:
@jart, this adds redirect following for Fetch, as the current implementation doesn't follow the redirects, which is probably an expected behavior. I also updated the documentation with related options. I kept it enabled by default, but can still be disabled if `followredirect=false` is passed.

I modeled the clearing of resource after the current processing, but would appreciate if you could take a quick look.


jart (approved):
LGTM

PR #230, opened 2021-08-07
Fix redbean's Lua-function Log, so it can be used in the global scope.

w13b3:
Before this fix Log crashed the redbean when Log was used in the global scope.
The demo versions of redbean has a Log in a main function, therefore it didn't crash.
Lua doesn't need a main function and every other redbean Lua-function can also be used in the most outer scope.

### DIFF
--- a/tool/net/redbean.c
+++ b/tool/net/redbean.c
@@ -4911,7 +4911,7 @@ static int LuaLog(lua_State *L) {
   level = luaL_checkinteger(L, 1);
   if (LOGGABLE(level)) {
     msg = luaL_checkstring(L, 2);
-    lua_getstack(L, 1, &ar);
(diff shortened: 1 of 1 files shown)


jart (approved):
Thank you for spotting this and taking the time to send a PR!

Also welcome as a new contributor. Even though it's a one line change, we need to go through one quick hurdle. Could you please email Justine Tunney <jtunney@gmail.com> and say that you intend to assign her the copyright to the changes you contribute to Cosmopolitan? See [CONTRIBUTING.md](https://github.com/jart/cosmopolitan/blob/master/CONTRIBUTING.md) for further details. We make the ask because it helps us ship the tiniest most legally permissive prim and proper binaries. You also only need to do it one time, so any future changes you send us can be quickly merged.

Thanks!

PR #235, opened 2021-08-09
Python 3.6.14 in third party

ahgamut:
First commit adds unmodified source from https://github.com/python/cpython/tree/3.6, the latest commit [`0a0a135`](https://github.com/python/cpython/commit/0a0a135bae2692d069b18d2d590397fbe0a0d39a) is used.

Second commit adds the changes from https://github.com/ahgamut/cpython/tree/cosmo_py36, squashed for convenience.

Third commit adds the necessary build config. No need to run `configure`, `pyconfig.h` and `Modules/config.c` can be modified along with the Makefiles when updating (ie if `rewinddir` is implemented or `sqlite` is added as a module to Python).

A good chunk of the Python standard library is added to the APE ZIP store to enable running `python.com`. Without the stdlib in the ZIP store, `python.com` will only work when called from the root of the repo, because it searches only itself and relatvie paths for files.

```bash
make MODE=tiny -j4 o/tiny/third_party/python
o/tiny/third_party/python/python.com
```

Compiled APE is 6.3MB with `MODE=tiny`. 

* Using `-isystem libc/isystem`, did not change all source files to use absolute header paths.
* Currently not doing the `.h.ok` checks or converting any `.c` files to `.inc` files
* Unnecessary files can be removed, but not done yet
* Have not tested this APE build with the test suite yet


ahgamut:
@jart `libc/isystem/sys/ioctl.h` tries to include a non-existent file `libc/sysv/consts/blk.h`

jart (approved):

jart:
I'm so pleased to be witnessing Cosmopolitan build a 6mb hermetic Python ZIP executable. Approved!

Note for anyone reading, we've only got 75% of the tests passing so far, so this is a WIP.

You mentioned performance earlier. I'm witnessing a similar slowdown doing a simple `time python.com -c 'print(2+2)'`. The bottleneck looks like `sys.path` i/o. I see you already managed to get all the python code within the zip structure. If we can make the default path resolution favor that, then it'll have a transformative impact on the performance of the Python language, since loading code from ZIP doesn't require i/o. That's always been Java's secret sauce for making bloat work at scale.

Some of the local C code changes made by commit "Source changes for compilation" are deserving of TODO comments or explanations. For example, I don't understand why GIL calls need to be commented out, since I'd've assumed they'd be a no-op in our environment. The `getdents64` thing looks like it could potentially be a big pain if we can't persuade Python to use the normal readdir() api.

Could you clarify more about the intent behind the `calculate_path()` changes? Ideally what I'd want is we just change the default `sys.path` to `['zip:.python', '']` and then let the C / POSIX APIs abstract the details of pulling it out of the zip structure. We'll likely want to add some flags to `zip.com` that let us tune the paths auto-generated to `.zip.o` files.

ahgamut:
Already got merged! Awesome.

> Some of the local C code changes made by commit "Source changes for compilation" are deserving of TODO comments or explanations. For example, I don't understand why GIL calls need to be commented out, since I'd've assumed they'd be a no-op in our environment. 

If I remember correctly, I have commented out the GIL calls at exactly one spot. Everything else was handled by disabling threading. I think there should be a `#ifdef WITH_THREAD` or something similar there, because I don't remember messing with the GIL for Python 2.7.

The other changes I did for 3.6 mostly followed from 2.7, so the explanations may be scattered across the commit messages of both my repos. I'll add them here or in #141 for discussion if that helps.

> Could you clarify more about the intent behind the calculate_path() changes? Ideally what I'd want is we just change the default sys.path to ['zip:.python', ''] and then let the C / POSIX APIs abstract the details of pulling it out of the zip structure. We'll likely want to add some flags to zip.com that let us tune the paths auto-generated to .zip.o files.

I think it should be fine to simplify `calculate_path`. The important thing is that the Python needs to be able to import the `encodings` module, and then the `site` module to set other stuff up. (Python 2.7 didn't need `encodings` by default during startup, was easier to set up).

There are so many entries in `sys.path` because I was experimenting.

* Python has a bunch of starting spots for imports (`Lib`, `Lib/site-packages`, `build/lib-linux-`, `$PYTHONPATH`, `$PYTHONHOME`), and I was juggling these options to find the best combo for experimentation.
* adding more packages - I included local paths because I was trying to download and use packages from `pip`. I can't `pip install`  directly into the zip because ETXTBSY, so I used to install them to a local dir `./Lib/site-packages` and keep that entry in `sys.path`.
* `python.com -m test <test_name>` sometimes tries to write a file in the directory of `__file__`, which doesn't play well with the zip store. I had the ZIP as the first entry before running the tests recently.

ahgamut:
It might be useful to compile the Python stdlib into bytecode before adding to the APE ZIP store. I am pretty sure it's faster, but haven't benchmarked it.

jart:
That would be nice, not only for performance, but because I'd rather not have Python saving files to source directories. We'll want to modify Python to put its pyc files in `o/` if it exists. But more importantly we need some way to solve the chicken and egg problem of having the build system compile pyc files without running Python code.

ahgamut:
> But more importantly we need some way to solve the chicken and egg problem of having the build system compile pyc files without running Python code.

How about a minimal `python.com` in `build/bootstrap/`? We select a minimal subset of the stdlib for now, and add more to the zip store if it helps the build.

ahgamut:
Alternatively, it would be interesting to have a C-based tool that converts Python3 source code to valid bytecode. A `pyobj.com`, say, to create the `.pyc` file before calling `zipobj.com` to get a `.pyc.zip.o`.

`python.com` as part of the build adds a lot of new possibilities, but startup time would be the issue. Building the Cosmopolitan repo from source is a nice test of `python.com`'s speed: if `python.com` needs to called for every `.pyc.zip.o` or even for groups of files across multiple applications, then it needs to be really fast. 

Maybe we can attempt some of the optimizations used in Python 3.7+, PyPy, or other implementations of Python (quick search shows Pyston and Cinder).

ahgamut:
> We'll want to modify Python to put its pyc files in o/ if it exists.

The compile function in `Lib/py_compile.py` allows to specify the target file's location. The default is to use `__pycache__` or the same location as the source file (PEP 3147vs PEP 488), but that  can be modified.

> I'd rather not have Python saving files to source directories.

`python.com -B` prevents writing the `.pyc` file.

ahgamut:
> It might be useful to compile the Python stdlib into bytecode before adding to the APE ZIP store. I am pretty sure it's faster, but haven't benchmarked it.

@jart bytecode compilation speeds up imports considerably. Startup time is almost halved if `.pyc` files are available for import.

`time python.com -BESc "2+2"` with the APE in this repo (partially tuned for cosmo, zip store contains only `.py` files):

```
real    0m0.116s
user    0m0.108s
sys     0m0.008s
```

`time python.com -BESc "2+2"` with the APE compiled using the amalgamation in my [`cosmo_py36` fork](https://github.com/ahgamut/cpython/tree/cosmo_py36) (not tuned for cosmo, but zip store contains only `.pyc` files):

```
real    0m0.056s
user    0m0.048s
sys     0m0.008s
```

jart:
I'm not sure if we have the resources to do native style compilation of Python code just yet. Maybe in the future! But we can totally have the build system precompile and embed pyc files. You're right it speeds things up a lot. It'll help us make tinier binaries too. Let's consolidate discussion in #141.

PR #237, opened 2021-08-10
Update redbean to include stack trace in Lua errors.

pkulchenko:
This adds a stack trace to Lua run-time errors in redbean.

### DIFF
--- a/tool/net/redbean.c
+++ b/tool/net/redbean.c
@@ -3200,12 +3200,13 @@ static char *ServeLua(struct Asset *a, const char *s, size_t n) {
   effectivepath.p = s;
   effectivepath.n = n;
   if ((code = FreeLater(LoadAsset(a, &codelen)))) {
-    int status = luaL_loadbuffer(L, code, codelen, FreeLater(strndup(s, n)));
+    int status = luaL_loadbuffer(L, code, codelen,
+      FreeLater(xasprintf("@%s", FreeLater(strndup(s, n)))));
     if (status == LUA_OK && LuaCallWithTrace(L, 0, 0) == LUA_OK) {
       return CommitOutput(GetLuaResponse());
     } else {
       char *error;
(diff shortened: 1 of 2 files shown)


pkulchenko:
@jart, I just realized that it only covers internal errors reported to the logs from `LuaRun`, but there are other places (like `ServeLua` that can benefit from the same handling). I'll work on generalizing the patch to include those places as well, but feel free to merge, as it may take a bit for me to work on this.

tool/net/redbean.c
```diff
@@ -3134,16 +3170,26 @@ static char *GetLuaResponse(void) {
   return p;
 }
 
+static bool IsLoopbackClient() {
+  uint32_t ip;
+  uint16_t port;
+  GetClientAddr(&ip, &port);
```
jart, line 3176: Please change this to GetRemoteAddr() so that backtraces won't leak to external hosts in well-configured proxy scenarios.

jart (changes requested):

tool/net/redbean.c
pkulchenko: Good point; pushed the fix along with a couple of other changes. The error location is reported now as `"/sql.lua"` instead of `[string "/sql.lua"]`.

tool/net/redbean.c
```diff
@@ -1062,6 +1062,35 @@ static void Daemonize(void) {
   ChangeUser();
 }
 
+static int LuaCallWithTrace(lua_State *L, int nargs, int nres) {
+  int nresults, status;
+  // create a coroutine to retrieve traceback on failure
+  lua_State *co = lua_newthread(L);
```
jart, line 1068: Now I'm so happy I asked you. There's a level of Lua wizardry happening here I would not have easily been able to figure out on my own. Good show.

jart (approved):
LGTM

PR #238, opened 2021-08-11
fix sysconfig check for build vars

ahgamut:
`python.mk` was changed to have `-DPLATFORM='"cosmo"'` and `-DMULTIARCH='"x86_64-cosmo"'`. 

This caused an error when loading the the interpreter because `Lib/sysconfig.py` looks for a file called `_sysconfigdata_($PLATFORM)_$(MULTIARCH).py`, which did not exist in the ZIP store of the APE. 
I renamed the existing sysconfigdata file and changed the makefile recipe accordingly.

IIRC some part of stdlib use `sys.getplatform()` to decide whether Python is on POSIX or Windows, and might get confused when "cosmo" is returned. I'll submit a PR for that once I can show the error.


jart (approved):
Thanks!

PR #239, opened 2021-08-11
Update redbean Lua to use integer instead of number where appropriate

pkulchenko:
This fixes an issue with port/ip/pid numbers printed/handled as floating point numbers instead of integers.


pkulchenko:
Resolved; thank you!

PR #240, opened 2021-08-12
running tests on the Python APE within the repo

ahgamut:
Added `os.pipe` because it was missing and that caused a lot of tests to fail. 
Removed threading necessity from the test suite for similar reasons.

Here's how I ran the test suite

```bash
make -j4 o//third_party/python
cd third_party/python/Lib/
../../../o/third_party/python/python.com -m test -x test_bytes test_format
```

(`test_bytes` triggers the `CHECK_NE (serious bug upstream)` in `third_party/python/Objects/longobject.c`)

278 tests pass.
63 tests fail.
(61+2) tests are  skipped.
3 tests do nothing.

Pass rate: 278/341 = **81.5%**.

Of the failing tests, some pass if run outside of the test suite for e.g. `python.com third_party/python/Lib/test/test_zipfile.py`.

Note that these results are different from the pass rate in #141.

### DIFF
--- a/third_party/python/Lib/test/libregrtest/main.py
+++ b/third_party/python/Lib/test/libregrtest/main.py
@@ -342,6 +342,7 @@ class Regrtest:
                 and len(self.good) > 1):
                 print("All", end=' ')
             print(count(len(self.good), "test"), "OK.")
+            printlist(self.good)
 
         if self.ns.print_slow:
             self.test_times.sort(reverse=True)
--- a/third_party/python/Lib/test/test_code.py
+++ b/third_party/python/Lib/test/test_code.py
@@ -104,7 +104,10 @@ consts: ('None',)
 
 import inspect
 import sys
-import threading
+try:
+    import threading
+except ImportError as e:
+    threading = None
 import unittest
 import weakref
 try:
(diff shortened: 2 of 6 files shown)


jart (approved):
Nice. It sounds like my coding spree these last few days has been a net win for Python. Thanks for the test runner commands. I was wondering about that. I'll refactor things so we don't need to run `cd` to invoke it. Hopefully we can make the Python tests go fast enough, and have few enough side-effects that they can be run as part of the normal build process.

Do we know if there's any tests that passed before but now fail?

ahgamut:
> Do we know if there's any tests that passed before but now fail?

I think `test_bytes` and `test_format` pass if the `CHECK_NE` is removed.  
The best way to check would to be compile `python.com` in the `cosmo_py36` branch with the amalgamation and compare both APEs on the test suite. The APE compiled with the amalgamation passes 290+/345, but the one compiled via the repo passes only 278/341. Listing the differing tests would help.

The `test` module in python messes with `sys.path` when running the tests: some tests pass/fail depending on whether an import is available. I think it would be better would be to set `sys.path` as `[""]` ie current dir only before running the tests, let me see if that helps.

PR #241, opened 2021-08-13
Redbean lua memory

pkulchenko:
Improves Lua memory usage and reporting.

### DIFF
--- a/tool/net/redbean.c
+++ b/tool/net/redbean.c
@@ -1080,6 +1080,8 @@ static int LuaCallWithTrace(lua_State *L, int nargs, int nres) {
   } else {
     // move results to the main stack
     lua_xmove(co, L, nresults);
+    // make sure the stack has enough space to grow
+    luaL_checkstack(L, nres - nresults, NULL);
     // grow the stack in case returned fewer results
     // than the caller expects, as lua_resume
     // doesn't adjust the stack for needed results
--- a/tool/net/redbean.c
+++ b/tool/net/redbean.c
@@ -3132,6 +3132,9 @@ static char *ServeStatusz(void) {
   AppendLong1("lastmeltdown", shared->lastmeltdown);
   AppendLong1("workers", shared->workers);
   AppendLong1("assets.n", assets.n);
+#ifndef STATIC
+  AppendLong1("lua.memory", lua_gc(L, LUA_GCCOUNT)*1024 + lua_gc(L, LUA_GCCOUNTB));
+#endif
   ServeCounters();
   AppendRusage("server", &shared->server);
   AppendRusage("children", &shared->children);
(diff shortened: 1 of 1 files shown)


tool/net/redbean.c
```diff
@@ -3130,6 +3135,9 @@ static char *ServeStatusz(void) {
   AppendLong1("lastmeltdown", shared->lastmeltdown);
   AppendLong1("workers", shared->workers);
   AppendLong1("assets.n", assets.n);
+#ifndef STATIC
+  AppendLong1("lua.memory", lua_gc(L, LUA_GCCOUNT)*1024 + lua_gc(L, LUA_GCCOUNTB));
```
jart, line 3139: This is really cool but please note it only reports Lua memory usage for the main process.

jart (approved):
LGTM

jart:
I had to remove the explicit Lua GC pass because it was causing crashes. Our tests caught it. Our Travis tests should have caught it before merging. Unfortunately GitHub removed them without warning, for reasons that aren't entirely clear to me. So we'll likely want to setup GitHub workflows soon.

As for the Lua GC thing, it'd be nice to know why it would cause crashes. However even if that's addressed, we shouldn't do it unless we're certain we're need it. We'd also want to measure how many nanoseconds it takes to execute if Lua has done nothing, since `CollectGarbage` is on the critical pass of the QPS benchmarking.

tool/net/redbean.c
pkulchenko: Yes, that was the intent, to see the usage in the main process.

pkulchenko:
> As for the Lua GC thing, it'd be nice to know why it would cause crashes.

Is there a way for me to review the results?

> However even if that's addressed, we shouldn't do it unless we're certain we're need it. 

Lua GC will kick in sooner or later, so an explicit call gives you better control over when in the process this happens, but it does need to be weighted against the cost of calling it too frequently.

> We'd also want to measure how many nanoseconds it takes to execute if Lua has done nothing, since CollectGarbage is on the critical pass of the QPS benchmarking.

I agree; maybe only do it as part of `HandleHeartbeat`?

jart:
Heartbeat only applies to the server process.  I don't know much about Lua's garbage collector.  Doesn't it kick in automatically? What advantage does explicitly invoking it give us?

pkulchenko:
Right, it will kick in automatically, but calling it explicitly gives us more control over when exactly it's called, as when it's automatic, it's called when the Lua interpreter decides it's time for it, which may be at some inopportune time from the redbean perspective.

Not critical, I was just trying to find a good time to call it. Right now it you keep refreshing `/statusz` page, you'll see that Lua.memory fluctuates between 50k and 80k; with the explicit call it was exactly the same every request (although I agree that it may be an overkill to call it after each message).

Another advantage of the explicit call is that we can then turn the GC off completely and call it when needed, as there is still some small overhead for Lua to check when to run GC automatically.

PR #244, opened 2021-08-14
Add GetRandomBytes to redbean Lua

pkulchenko:
I considered `GetRandom`, `GetRandomString`, and `GetRandomBytes` and think the last is the best one.

### DIFF
--- a/tool/net/redbean.c
+++ b/tool/net/redbean.c
@@ -4471,6 +4471,21 @@ static int LuaSha512(lua_State *L) {
   return LuaHasher(L, 64, mbedtls_sha512_ret_512);
 }
 
+static int LuaGetRandomBytes(lua_State *L) {
+  char *p;
+  size_t n = luaL_optinteger(L, 1, 16);
+  if (!(n > 0 && n <= 256)) {
+    luaL_argerror(L, 1, "not in range 1..256");
+    unreachable;
+  }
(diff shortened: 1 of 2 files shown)


tool/net/redbean.c
```diff
@@ -4471,6 +4471,16 @@ static int LuaSha512(lua_State *L) {
   return LuaHasher(L, 64, mbedtls_sha512_ret_512);
 }
 
+static int LuaGetRandomBytes(lua_State *L) {
+  char *p;
+  size_t n = luaL_optinteger(L, 1, 16);
+  p = malloc(n);
+  CHECK_EQ(n, getrandom(p, n, 0));
```
jart, line 4478: getrandom() won't return more than 256 bytes. If the requested amount is larger then you could provide cheap random using `rngset(buf, sizeof(buf), rand64, -1);`.

jart (changes requested):

jart:
Here's the random number battery tests if you're interested: https://justine.lol/cosmopolitan/random.txt

tool/net/redbean.c
pkulchenko: Good point; added a check for this. I don't think anything longer 256 would be needed.

jart (approved):

PR #245, opened 2021-08-15
typo in strtoll causing recursion

ahgamut:
found when running `python.com -m test test_cmath` lead to random segfaults.


jart (approved):
Thank you!

PR #246, opened 2021-08-16
Add ServeRedirect function to redbean Lua

pkulchenko:
This is useful for dynamic redirects (and is used in various frameworks, for example, in lapis).

### DIFF
--- a/tool/net/redbean.c
+++ b/tool/net/redbean.c
@@ -3059,6 +3059,32 @@ static int LuaServeIndex(lua_State *L) {
   return 1;
 }
 
+static int LuaServeRedirect(lua_State *L) {
+  size_t loclen;
+  const char *location;
+  int code;
+  OnlyCallDuringRequest("ServeRedirect");
+
+  code = luaL_checkinteger(L, 1);
+  if (!(300 <= code && code <= 399)) {
+    luaL_argerror(L, 1, "bad status code for redirect");
+    unreachable;
+  }
+  location = luaL_checklstring(L, 2, &loclen);
+
+  if (msg.version < 10) {
+    (void)ServeError(505, "HTTP Version Not Supported");
+    lua_pushboolean(L, false);
+  } else {
+    LOGF("REDIRECT %d to %s", code, location);
+    luaheaderp = AppendHeader(
+          SetStatus(code, GetHttpReason(code)), "Location",
+          FreeLater(EncodeHttpHeaderValue(location, loclen, 0)));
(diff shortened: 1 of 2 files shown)


tool/net/redbean.c
```diff
@@ -3059,6 +3059,32 @@ static int LuaServeIndex(lua_State *L) {
...
+  location = luaL_checklstring(L, 2, &loclen);
+
+  if (msg.version < 10) {
+    (void)ServeError(505, "HTTP Version Not Supported");
+    lua_pushboolean(L, false);
+  } else {
+    LOGF("REDIRECT %d to %s", code, location);
+    luaheaderp = AppendHeader(
+          SetStatus(code, GetHttpReason(code)), "Location",
+          FreeLater(EncodeHttpHeaderValue(location, loclen, 0)));
```
jart, line 3082: `EnocdeHttpHeaderValue` might return `NULL` if it isn't latin1 encodable. There's also only a limited amount of space in this buffer. `LuaSetHeader` will do the appropriate checks. Are we sure we need a function like this in core?

jart (changes requested):

tool/net/redbean.c
pkulchenko: I'll fix the code to make it work based on your recommendation.

In terms of whether it belongs in core or not, I'm on the fence. I can implement it in Lua, but we don't have any Lua core that comes included with redbean, so there is no good place to put it in. Also, I personally see `ServeRedirect` to be used more frequently than `ServeListing`, but it may be good to have both available.

I'd like to discuss where we should draw a line, as I'm planning on submitting a PR for `SetCookie` and `GetCookie` functions, but maybe it's something to include in one of the (to be provided) Lua modules.

My take on this is simple: if there is no way to access it from Lua, then it must be in the core (for example, hashing functions). If there is a way to implement it in Lua, but it's likely to be used frequently (and both ServeRedirect and SetCookie falls into that category), I'd also prefer to include it, so it doesn't need to be re-implemented. I'm also looking at openresty/nginx and all these functions are available there (in fact, most of my PRs stem from updating Lapis to work with redbean instead of openresty/nginx).

pkulchenko: Updated to check for `EncodeHttpHeaderValue` returning NULL.

@jart, would you also like me to update `HandleRedirect`, as it may have a similar issue?

jart: Ah good catch. Sounds good to me.

jart (approved):
LGTM

PR #248, opened 2021-08-18
tell _frozen_importlib to consider bytecode first

ahgamut:
In `Lib/importlib/_bootstrap_external.py`, I have the interpreter consider the `SourcelessFileLoader` (ie `.pyc`) first and only then fall back to the `SourceFileLoader`. I run `freeze.com` to update the `importlib*.inc` with this change.

(Note: Why not avoid the above change by simply excluding the `.py` files to the ZIP store?)

the change in `pylifecycle.c` is with respect to my comment [here](https://github.com/jart/cosmopolitan/issues/141#issuecomment-900500282): there is a *tiny* speed benefit to having the C extensions imported earlier.

in `python-stdlib.mk` all I did was a find-and-replace. 

We don't have to have to follow PEP 3147: `Lib/X.py` doesn't have to be compiled into `Lib/__pycache__/X.cpython-36.pyc`, the __pycache__ + version-name-in-extension business matters only if we are mixing `.pyc` files from different versions of Python. 

Since we are within the ZIP store of a Python 3.6 APE, it's highly unlikely there will be a mix of `.pyc` files, so we can stick to the legacy format of `X.py` to `X.pyc`. This ought to make the Makefile recipe simpler as well.

I removed the `#if` in `Modules/getpath.c` and just set `module_search_path` to a static value. Now the `calculate_path` function is skipped entirely; the APE works without issue, so I don't think it is that important.

The final change was to make `site.py` skip modifying the location of the zip store, which messed with everything.

The sum effect of all these changes: `time python.com -c "2+2"` now takes 0.044s


jart (approved):
This is very good. Excellent insight on the absolutize btw. We might want to make Python not do that in general, because absolute paths aren't portable. I'm so happy to hear we don't need the `__pycache__` subdirectory. I was worried about that one.

PR #253, opened 2021-08-19
Update timestamp delta format for better log processing

pkulchenko:
This still produces full timestamp to simplify log processing, but changes microsecond separator from `.` to `+` to show delta measurements.


jart (approved):
LGTM

PR #254, opened 2021-08-19
remove pycache from makefile

ahgamut:
running `make -j4 o//third_party/python` fails because `__pycache__` folders are not present in the repo.


jart:
Thanks for sending this. I've addressed the problem (and others!) in 7bf5cd18350c5088ac7bf62589780636d4aa6479.

PR #256, opened 2021-08-22
Redbean log improv

pkulchenko:
This implements several changes:

- Updated log level for redbean messages to better work with verbose/silent logging
- Moved `FATALF` call to the end of `HandleForkFailure`
- Added categories for each log message
- Replaced `perror` with an appropriate log message using `%m`
- Replaced `flogf(); exit();` with (added) `DIEF`
- Slightly reorganized Lua error messages for consistency and reduced number of messages

### DIFF
--- a/libc/log/log.h
+++ b/libc/log/log.h
@@ -72,6 +72,15 @@ extern unsigned __log_level; /* log level for runtime check */
   ((!__builtin_constant_p(LEVEL) || (LEVEL) <= LOGGABLELEVEL) && \
    (LEVEL) <= __log_level)
 
+// die with an error message without backtrace and debugger invocation
+#define DIEF(FMT, ...)                                              \
+  do {                                                              \
+    ++ftrace;                                                       \
+    flogf(kLogError, __FILE__, __LINE__, NULL, FMT, ##__VA_ARGS__); \
+    exit(1);                                                        \
+    unreachable;                                                    \
+  } while (0)
+
 #define FATALF(FMT, ...)                                              \
   do {                                                                \
     ++ftrace;                                                         \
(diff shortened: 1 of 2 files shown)


libc/log/log.h
```diff
@@ -72,6 +72,15 @@ extern unsigned __log_level; /* log level for runtime check */
   ((!__builtin_constant_p(LEVEL) || (LEVEL) <= LOGGABLELEVEL) && \
    (LEVEL) <= __log_level)
 
+// die with an error message unconditionally
+#define DIEF(FMT, ...)                                              \
+  do {                                                              \
+    ++ftrace;                                                       \
+    flogf(kLogError, __FILE__, __LINE__, NULL, FMT, ##__VA_ARGS__); \
```
jart, line 79: I actually didn't intend for FATALF() to be conditional. Could we fix that? Thanks for spotting it.

jart (changes requested):

libc/log/log.h
pkulchenko: @jart, I think the problem is not that `FAULTF` is conditional (I should have been a bit more clear on this as I intended to reference some additional processing that only happens when `kLogFatal` is used). The main issue is that it has some additional output (https://github.com/jart/cosmopolitan/blob/master/libc/log/vflogf.c#L112-L114) and that it uses `__die`, which shows backtrace and breaks in a debugger, so it can't (or shouldn't) be used to report an error and simply exit (and should most likely be reserved for internal issues that would require some debugging). `DIEF` fills that need (it's a bit shorted than ERROR-AND-DIE).

I just noticed that `__die` also restores tty (calls `__restore_tty();`). Should this be done in `DIEF` as well?

pkulchenko: @jart, I updated the comment for clarity. Also, I can switch some of `DIEF` back to `FATALF` if you think it's going to work better (for those that involve `socket` and `getsockname` errors inside `Listen` method.

jart: Oh I see. That sounds fine. For what it's worth, 00611e9b should help the crash reporting system work better since it makes discovery of the .com.dbg file more resilient.

jart (approved):
LGTM

PR #258, opened 2021-08-25
Fix redbean StoreAsset zip update when Deflate algorithm is used

pkulchenko:
This fixes an issue with assets incorrectly stored in the archive (not compressed, but marked as deflated).


jart (approved):

jart:
Thanks!

PR #259, opened 2021-08-25
Added _multiprocessing C extension to stdlib

ahgamut:
Also moved `PYTHON_YOINK` of `email` and `encodings.idna` so that `http.server` would work in MODE=tiny.
(If tree-shaking for Python stdlib is possible, then the above change will be handled during the build process).

Should the `PYTHON_YOINK`s be somewhere else than `python.c`? 

Many stdlib components have `X.py` and `_X.c` (locale, multiprocessing, sqlite for example) so it may be better to add `PYTHON_YOINK("sqlite.py")` to `_sqlite.c` and then skip adding `sqlite3.pyc` if `_sqlite3` is not compiled.


jart:
Hmm that's true the build system needs awareness of where these python files live.

Your JSON manifest idea is absolutely the sort of thing we're going to need. I built a similar tool some time ago called [PACKAGE.COM](https://github.com/jart/cosmopolitan/blob/master/tool/build/package.c) which works by the same principle for the C source code.

```
master jart@nightmare:~/cosmo$ bf o/libc/zipos/zipos.a.pkg
╛∩╛∩☺                                   ♪♠      ä♠      └       
♪       ↕        •      ♪       ↕       └◙                      
                                                                
o//libc/zipos/zipos.a.pkg o//libc/zipos/zipos.o o//libc/zipos/cl
ose.o o//libc/zipos/fcntl.o o//libc/zipos/find.o o//libc/zipos/f
stat.o o//libc/zipos/get.o o//libc/zipos/lseek.o o//libc/zipos/n
otat.o o//libc/zipos/open.o o//libc/zipos/parseuri.o o//libc/zip
os/read.o o//libc/zipos/stat-impl.o o//libc/zipos/stat.o zip_uri
_support __zip_end __zip_start __zipos_close __zipos_fcntl __zip
os_fstat __zipos_lseek __zipos_open __zipos_parseuri __zipos_rea
d __zipos_stat __zipos_notat __zipos_close mcount g_fds __hostos
 sys_close __vforked CloseHandle free __zipos_fcntl mcount g_fds
 O_CLOEXEC einval __zipos_find mcount GetZipCdirOffset GetZipCdi
rRecords __assert_fail memcmp __zipos_fstat mcount __zipos_get _
_zipos_stat_impl efault __zipos_get mcount sigfillset SIG_BLOCK 
sigprocmask O_RDONLY program_executable_name open SIG_SETMASK ge
tfiledescriptorsize close mmap endswith memmem GetZipCdir munmap
 __zipos_lseek mcount einval __zipos_notat mcount g_fds efault e
inval __zipos_parseuri mcount inflateInit2 enomem inflate eio en
otsup inflateEnd abort GetZipCfileOffset calloc GetZipLfileUncom
pressedSize crc32_z EIO errno dup __ensurefds g_fds O_CLOEXEC ma
lloc GetZipLfileCompressedSize close free __zipos_open O_ACCMODE
 O_RDONLY einval __zipos_get __zipos_find enoent enoexec __zipos
_parseuri mcount strlen __zipos_read mcount MemCpy __zipos_stat_
impl mcount MemSet GetZipCfileOffset GetZipLfileUncompressedSize
 GetZipLfileCompressedSize GetZipCfileTimestamps einval __zipos_
stat mcount __zipos_get __zipos_find enoent efault __zipos_stat_
impl enoexec                                                    
→   ☻     ↑Ç ►  P♂      ÿ♥↑Ç ►  p☺↑Ç ►  ↨       $       -       
áò Ç ►  0   ☻     ↑Ç ►  á0      ╚↓↑Ç ►  ╕↨↑Ç ►  ▬       *       
-       ░¥ Ç ►  F   ☻     ↑Ç ►  αU      ¿+↑Ç ►  α)↑Ç ►  ‼       
*       -       └Ñ Ç ►  ╲   ☻     ↑Ç ►  02      ≡←↑Ç ►  α↓↑Ç ►  
▬       .       E       ≡╣ Ç ►  q   ☻     ↑Ç ►  └E      p#↑Ç ►  
¿!↑Ç ►  ‼       *       -        ╩ Ç ►  ç   ☻     ↑Ç ►  ¿]      
`/↑Ç ►  H,↑Ç ►  !       ,       -       ►╥ Ç ►  ¢   ☻     ↑Ç ►  
░A      x!↑Ç ►  α▼↑Ç ►  ◄       *       -        ┌ Ç ►  ▒   ☻   
  ↑Ç ►  ╪M      8(↑Ç ►  @&↑Ç ►  §       ,       -       0Γ Ç ►  
╟   ☻     ↑Ç ►  αî      xL↑Ç ►  (H↑Ç ►  .       *       -       
► ↓Ç ►  ▄   ☻     ↑Ç ►  H%      P¶↑Ç ►  ╕↕↑Ç ►  ◄       *       
-       Ç> Ç ►  ⌡   ☻     ↑Ç ►  86      °↔↑Ç ►  `∟↑Ç ►  ◄       
*       -        ◘↓Ç ►  ◙☺  ☻     ↑Ç ►  ÉM      8(↑Ç ►  (&↑Ç ►  
▬       *       -        _ Ç ►  $☺  ☻     ↑Ç ►  ╚L      H‘↑Ç ►  
8%↑Ç ►  ▬       *       -       0►↓Ç ►                          
▌☺  ☺!☺ &☻  ☺!☻ R☻  ☺!♥ €☻  ☺!♦ ╪☻  ☺!♣ ü♥  ☺!♠ ¥♥  ☺!• ¬♦  ☺!◘ 
∙♦  ☺!○ ↑♣  ☺!◙ ╕♣  ☺!♀ 3♣  ☺!♂ 9☺  ☺☺                          

```

What the binary dump shows us is that the `.pkg` files define:

1. The object names
2. What symbols they define
3. What symbols they depend on

The `x_DEPS` variables in our `.mk` files basically consist of those `.pkg` files. We can have `PYCOMP.COM` plug into those, in order to tell pure Python dependencies apart from Python C modules. All it has to do is ask the `.pkg` files if a `.globl` ELF symbol named `"PyInit_" + pkgname` exists. If it does, then it's a builtin module. If it doesn't exist, then either (a) you need to update your x_DIRECTDEPS list in your makefile, or (b) it must be a pure Python source.

The hard part is figuring out how we're going to incorporate knowledge of pure Python source locations into the build, in a way that's scalable. I mean scalable in the sense that, I'm sure we're going to end up vendoring other third party libraries that are written in Python. They'll need to plug harmoniously and independently in the build the same way native libraries do. It's something I'm going to sit and think about for a little while.

jart:
Ah I think a solution is starting to become clear in my mind. 

So right now, here's all the steps we have to go through in our makefile to compile Python sources into APE binaries:

```sh
pycomp.com -o prefix/base.pyc prefix/base.py
zipobj.com -P.python -C1 -o prefix/base.py.zip.o prefix/base.py
zipobj.com -P.python -C1 -o prefix/base.pyc.zip.o prefix/base.pyc

zipobj.com -P.python -C1 -o prefix/module/.zip.o prefix/module/

pycomp.com -o prefix/module/__init__.pyc prefix/module/__init__.py
zipobj.com -P.python -C1 -o prefix/module/__init__.py.zip.o prefix/module/__init__.py
zipobj.com -P.python -C1 -o prefix/module/__init__.pyc.zip.o prefix/module/__init__.pyc

pycomp.com -o prefix/module/nested.pyc prefix/module/nested.py
zipobj.com -P.python -C1 -o prefix/module/nested.py.zip.o prefix/module/nested.py
zipobj.com -P.python -C1 -o prefix/module/nested.pyc.zip.o prefix/module/nested.pyc

ld -T ape.lds -o foo.com foo.o prefix/base.py.zip.o ...

unzip -vl foo.com
.python/base.py
.python/base.pyc
.python/module/
.python/module/__init__.py
.python/module/__init__.pyc
.python/module/nested.py
.python/module/nested.pyc
```

So we simplify this to:

```sh
pycomp.com -o prefix/base.py.o prefix/base.py
# generated by pycomp! zipobj.com -P.python -C1 -o prefix/base.py.zip.o prefix/base.py
# generated by pycomp! zipobj.com -P.python -C1 -o prefix/base.pyc.zip.o prefix/base.pyc
# defines elf symbol "py:base"
# defines elf symbol ".python/enum.py"
# defines elf symbol ".python/enum.pyc"
# depends on symbol "__zip_start"

# pycomp generates directory entry automatically for __init__.py sources!
# zipobj.com -P.python -C1 -o prefix/module/.zip.o prefix/module/

pycomp.com -o prefix/module/__init__.py.o prefix/module/__init__.py
# generated by pycomp! zipobj.com -P.python -C1 -o prefix/module/__init__.py.zip.o prefix/module/__init__.py
# generated by pycomp! zipobj.com -P.python -C1 -o prefix/module/__init__.pyc.zip.o prefix/module/__init__.pyc
# defines elf symbol "py:module"
# defines elf symbol ".python/module/"
# defines elf symbol ".python/module/__init__.py"
# defines elf symbol ".python/module/__init__.pyc"
# depends on symbol "__zip_start"

pycomp.com -o prefix/module/nested.py.o prefix/module/nested.py
# generated by pycomp! zipobj.com -P.python -C1 -o prefix/module/nested.py.zip.o prefix/module/nested.py
# generated by pycomp! zipobj.com -P.python -C1 -o prefix/module/nested.pyc.zip.o prefix/module/nested.pyc
# defines elf symbol "py:module.nested"
# defines elf symbol ".python/module/nested.py"
# defines elf symbol ".python/module/nested.pyc"
# depends on symbol "__zip_start"
# depends on symbol "py:module"
```

Then we feed the .py.o files into package.com and the linkage should work.

ahgamut:
~~`base.py` defining symbol `enum.py`?~~ The other two examples make it clear. 

> All it has to do is ask the .pkg files if a .globl ELF symbol named "PyInit_" + pkgname exists. If it does, then it's a builtin module. If it doesn't exist, then either (a) you need to update your x_DIRECTDEPS list in your makefile, or (b) it must be a pure Python source.

Just for reference, Python2 extensions use "init_" pkgname for starting the module, for eg. the [`_greenlet` C extension source](https://github.com/python-greenlet/greenlet/blob/master/src/greenlet/greenlet.c#L2010) has an ifdef for Py2 vs Py3.

jart:
Name shouldn't matter. I determined it was easier to have each Python C module define a fake symbol `"pyc:mod.name"` which is exactly the same as what would be generated by pyobj.com.

jart:
Could you help review the byte code parsing I'm doing? You were correct that relative imports are tricky. I don't fully understand their syntax to be honest. I'm certainly going to need your help, making sure all the edge cases are covered.

https://github.com/jart/cosmopolitan/blob/44c87b83ff5a192c8be836686b3b3e909101db63/third_party/python/pyobj.c#L246-L301

ahgamut:
> The hard part is figuring out how we're going to incorporate knowledge of pure Python source locations into the build, in a way that's scalable.  I mean scalable in the sense that, I'm sure we're going to end up vendoring other third party libraries that are written in Python. They'll need to plug harmoniously and independently in the build the same way native libraries do

This can possibly be done with Python's `setup.py` machinery (+ `OpenExecutable` if it's absolutely needed), but I am not sure if it is better than the current method.

The Stage1/Stage2 builds in `python.mk` are similar to how Python adds the stdlib extensions:

1. Build minimal `python` for Stage 1
2. run `setup.py` with minimal `python` for adding pure-python parts in Stage 2 (see [here](https://github.com/ahgamut/cpython/blob/cosmo_py36/Makefile.pre.in#L1571) for example.

jart:
QQ: How much does the ordering of `_PyImport_Inittab` matter?

ahgamut:
> Could you help review the byte code parsing I'm doing? You were correct that relative imports are tricky. I don't fully understand their syntax to be honest. I'm certainly going to need your help, making sure all the edge cases are covered.

The edge cases are mostly because it is not clear from the import statement whether a whole module or just a function is being imported. (IIRC `import os.path` is an annoying example of this ambiguity). Might require a file check call. 

I'll compare the output of `pyobj` to see what I can do.

jart:
I'm not too concerned about that practice. It's actually banned at Google. https://google.github.io/styleguide/pyguide.html#22-imports

ahgamut:
> QQ: How much does the ordering of `_PyImport_Inittab` matter?

I don't think the ordering matters. The start of `_PyImport_Inittab` is filled in by the `makesetup` shell script when building Python: look through `Modules/Setup.local`, then `Modules/Setup`, and fill the imports in that order. Proof: `_decimal` is first in the list, and `_decimal` is first in the [`Module/Setup.local` in my `cosmo_py36` fork](https://github.com/ahgamut/cpython/blob/cosmo_py36/Modules/Setup.local#L49).

What is important is that the `PyInit_pkgname` function is available for the interpreter to call during imports, and the corresponding name/entry is present in Inittab so that it can perform the lookup when importing a builtin module.

jart:
Right now I'm working on a change that's going to tree shake config.c and it's going to be epic.

jart:
Oh gosh you're right. I see your point now.

```
$?=1 master jart@nightmare:~/cosmo$ o//third_party/python/httpserver.com -c 'import http.server'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/zip/.python/http/server.py", line 90, in <module>
    import email.utils
  File "/zip/.python/email/utils.py", line 40, in <module>
    from email.charset import Charset
  File "/zip/.python/email/charset.py", line 17, in <module>
    from email import errors
ImportError: cannot import name 'errors'
```

Doodle.

jart:
You know what? We're using a bona fide linker. It actually doesn't matter what is/isn't a file. All we have to do is generate ELF symbols for the functions and variables each module defines.

ahgamut:
> You know what? We're using a bona fide linker. It actually doesn't matter what is/isn't a file. All we have to do is generate ELF symbols for the functions and variables each module defines.

That seems like too much, no? I thought we can just have a file check in `pyobj.c` to see if `import email.utils` actually points to an `emails/utils.py` or not.

jart:
We can't have build commands probe the file system because the makefile definition has no awareness of it. So it's a side-effect that makes the build non-deterministic. Unless of course we do what Python does where they have a single command compile the world. But then the build can't be parallelized. I don't care about performance, since that's a second-order effect of having a deterministic hermetic parallelized build.

Right now we can build the python standard library in a few milliseconds and I'd love to keep it that way. I believe you're right that static analysis is the answer. It goes without saying it's not perfect, but I'm crossing my fingers hoping we can make it work good enough. We will have to refactor a few things, like:

```py
from .base_events import *
from .coroutines import *
from .events import *
from .futures import *
from .locks import *
from .protocols import *
from .queues import *
from .streams import *
from .subprocess import *
from .tasks import *
from .transports import *

__all__ = (base_events.__all__ +
           coroutines.__all__ +
           events.__all__ +
           futures.__all__ +
           locks.__all__ +
           protocols.__all__ +
           queues.__all__ +
           streams.__all__ +
           subprocess.__all__ +
           tasks.__all__ +
           transports.__all__)
```

Because code like that is even more evil than `tensorflow.__init__` so it'd have to be spelled out like:

```py
from .base_events import BaseEventLoop
# etc.
__all__ = ["long", "list", "of", "names", ...]
```

ahgamut:
Simple refactor fix: run `yapf`/`isort` with default Google style guide settings on the stdlib?

> from .base_events import *

IIUC the above line is not an issue: if `module/X.py` has the above line, all that we need to do is ensure `modules/base_events.py` is available.

jart:
What's yapf and isort?

Are you sure? What if some other module says `from asyncio import BaseEventLoop`? In that case, we need to yoink `asyncio/__init__.o` which will in turn need to yoink `asyncio/base_events.o`.

ahgamut:
https://github.com/PyCQA/isort sorts import statements in Python

https://github.com/google/yapf formats Python code

keep imports in a standard, simple format, to make it easier for collecting static dependencies?

ahgamut:
> Are you sure? What if some other module says from asyncio import BaseEventLoop? In that case, we need to yoink asyncio/__init__.o which will in turn need to yoink asyncio/base_events.o.

I expect the below format to be sufficient:

```json
    "Lib/asyncio/__init__.py": {
        "builtins": [
            "sys",
            "_overlapped"
        ],
        "files": [
            "Lib/asyncio/windows_events.py",
            "Lib/selectors.py",
            "Lib/asyncio/base_events.py",
            "Lib/asyncio/unix_events.py",
            "Lib/asyncio/__init__.py",
            "Lib/asyncio/subprocess.py",
            "Lib/asyncio/streams.py",
            "Lib/asyncio/transports.py",
            "Lib/asyncio/protocols.py",
            "Lib/asyncio/coroutines.py",
            "Lib/asyncio/events.py",
            "Lib/asyncio/locks.py",
            "Lib/asyncio/queues.py",
            "Lib/asyncio/futures.py",
            "Lib/asyncio/tasks.py"
        ]
    },
```

jart:
Oh I see. That's a pretty cool formatter. We don't need to beautify the Python sources, since `pyobj.c` is operating on a parsed Python bytecode representation. We just need to make them a little less magical. If our goal is to turn a python source file into an elf object, then that python source file shouldn't have things like star import statements that effectively say, "sorry mario but the princess is in a different castle"

ahgamut:
just to clarify, yapf/isort only beautify the source, the JSON I created via a custom script

ahgamut:
https://gist.github.com/ahgamut/f0a6be870c4d1580fbda40d8079d01f7

(Runs with APE, run in `third_party/python`)

Ok, I think the below script takes care of "most" stdlib dependencies statically ("most" because there are conditional imports here and there which might be incorrectly included). 

All tests are skipped because they use freely dynamic imports.

PR #262, opened 2021-08-30
Log reorg

pkulchenko:
This PR renames `LOGF` to `INFOF` and `ANYF` to `LOGF`; `LOGF` now accepts log level, so it's a breaking change. I think I caught all the placed that need to be updated (including comments), but if there is a command to build all affected executables, I'll use it.

Along the same lines, should `LOGIFNEG1` and `LOGIFNULL` be named `WARNIFNEG1` and `WARNIFNULL`?

### DIFF
--- a/dsp/mpeg/mpeg1.c
+++ b/dsp/mpeg/mpeg1.c
@@ -591,24 +591,24 @@ void plm_video_decode_sequence_header(plm_video_t *self) {
 
   self->has_sequence_header = true;
 
-  LOGF("%s:\n"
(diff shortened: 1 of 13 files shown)


jart:
> if there is a command to build all affected executables, I'll use it.

That would be `make` :-)

> Along the same lines, should LOGIFNEG1 and LOGIFNULL be named WARNIFNEG1 and WARNIFNULL?

I'm indifferent. Probably not worth changing.

jart (approved):
LGTM

PR #264, opened 2021-09-03
speedups from pyston

ahgamut:
The pyston wiki has some [optimizations for speedups](https://github.com/pyston/pyston/wiki/Semantic-changes):

* Removing `tracemalloc`: see [this commit](https://github.com/pyston/pyston/commit/5bd9138a926c2b4b7c822734e7630c1ac39dc45f) for reference, and [this one as well](https://github.com/pyston/pyston/commit/f39d9b4741da7e67bfc7345f3380ca06ef65fdb4). `pyston` uses a dummy module when disabling `tracemalloc`, I just skip adding the module entirely, because both end up triggering an `ImportError` anyway.
* Removing `PyErr_BadInternalCall`: see [this commit](https://github.com/pyston/pyston/commit/2c325e95c764771478e89978a0d45672fce5f826) for reference. 

I have hardcoded a skip of three tests due to removing `PyErr_BadInternalCall`.
pyston uses a new attribute in the `sys` module to skip these tests, I think APE python can have a different approach: 

* a `_cosmo` C extension module in the stdlib which contains all the necessary info (+ potential speedup via shortcut functions)
* create a custom `_sysconfigdata` as a target in `python.mk` (after `STAGE1`) similar to how Python does with `getbuilddir.txt`


ahgamut:
Also removed `CheckFunctionResult`: [reference](https://github.com/pyston/pyston/commit/b052314d6c2b87f19926ac6318503ce4e9557657#diff-cf6158859b32f9548b010233c5c8cd5ef7631a8080a45f8f72f09c61aa898483)

I think it is also possible to remove Python's recursion limit checking like Pyston does. Let me see how that works.

ahgamut:
@jart I just put together a simple `_cosmo` module because I thought it would be nice to have things like:

* access Cosmopolitan-specific information in a standard way (compile-time constants like `MODE`, functions like `getHostOs()` etc.)
* compiled C versions of some function sequences for speed (for example `_w_long` and `_r_long` in `_bootstrap_external` or even a new ModuleFinder/Loader)
* access features like `OpenExecutable`, `chibicc`, `ftrace` etc. from Python. `ftrace` in particular would be interesting to use for certain blocks of Python code:

```python
import _cosmo

# some python code
with _cosmo.ftracer() as tracer:
    # call ftrace_install() internally because there is
    # a specific code block I want to examine
# have ftrace_disable() to avoid deluge of info

# more code
```

third_party/python/Python/pylifecycle.c
```diff
@@ -462,8 +477,10 @@ _Py_InitializeEx_Private(int install_sigs, int install_importlib)
     if (install_sigs)
         initsigs(); /* Signal handling stuff, including initintr() */
 
+#ifdef USE_TRACEMALLOC
     if (_PyTraceMalloc_Init() < 0)
```
jart, line 481: I would recommend writing these statements as:

```c
if (IsModeDbg() && _PyTraceMalloc_Init() < 0)
```

See https://github.com/jart/cosmopolitan/blob/master/libc/dce.h

third_party/python/Include/pymem.h
```diff
@@ -63,6 +64,11 @@ int _PyTraceMalloc_Untrack(
 PyObject* _PyTraceMalloc_GetTraceback(
     _PyTraceMalloc_domain_t domain,
     uintptr_t ptr);
+#else
+#define PyTraceMalloc_Track(domain, ptr, size) (-2)
```
jart, line 68: Did you forget the underscores here?

What you can do in this case is, something along the lines of:

```c
/* header */
int _PyTraceMalloc_Track(_PyTraceMalloc_domain_t, uintptr_t, size_t);
#ifdef MODE_DBG
#define _PyTraceMalloc_Track(domain, ptr, size) (-2)
#endif

/* implementation */
int (_PyTraceMalloc_Track)(_PyTraceMalloc_domain_t domain, uintptr_t ptr, size_t size) {
// ....
```

That's my favorite trick for overriding functions with a macro.

third_party/python/Include/cosmo.h
```diff
@@ -0,0 +1,10 @@
+#ifndef Py_COSMO_H
+#define Py_COSMO_H
+#include "third_party/python/Include/object.h"
+COSMOPOLITAN_C_START_
+/* clang-format off */
```
jart, line 5: Nit: not worth adding clang-format off if it's new code.

third_party/python/Modules/_testcapimodule.c
```diff
@@ -4112,7 +4112,7 @@ pyobject_malloc_without_gil(PyObject *self, PyObject *args)
 
     Py_RETURN_NONE;
 }
-
+#ifdef USE_TRACEMALLOC
```
jart, line 4115: Not worth ifdef'ing this out. We use `-ffunction-sections` so it'll be pruned by the linker.

jart (changes requested):
Thanks for sending! See recommendations for simplifying this change.

third_party/python/Python/pylifecycle.c
ahgamut: I've used `#ifdef MODE_DBG` for this and `_PyTraceMalloc_Fini`, which isn't in a conditional.

jart (approved):
Thank you!

PR #265, opened 2021-09-04
Add SetCookie method to redbean Lua

pkulchenko:


net/http/isvalidcookievalue.c
```diff
@@ -0,0 +1,39 @@
...
+ * (any ASCII excluding control char, whitespace,
+ * double quotes, comma, semicolon, and backslash).
+ *
+ * @param n if -1 implies strlen
+ */
+bool IsValidCookieValue(const char *s, size_t n) {
+  size_t i;
+  if (!n) return false;
+  if (n == -1) n = s ? strlen(s) : 0;
+  for (i = 0; i < n; ++i) {
```
jart, line 33: What about `DQUOTE *cookie-octet DQUOTE`? https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.1

net/http/isvalidcookievalue.c
```diff
@@ -0,0 +1,39 @@
...
+    if (!kHttpCookieValue[s[i] & 0xff]) {
```
jart, line 34: I'd recommend the following code since the generated assembly looks faster and tinier than a lookup table:

```c
static inline bool IsCookieOctet(unsigned char i) {
  return i > 0x20 && i < 0x7F && i != ',' && i != ';' && i != '\\' &&
         i != ' ' && i != '"';
}
```

net/http/khttpcookievalue.c
```diff
@@ -0,0 +1,41 @@
...
+// 1 is everything else in ASCII which is legal
+const char kHttpCookieValue[256] = {
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,  // 0x00
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,  // 0x10
+    0, 1, 0, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 1, 1, 1,  // 0x20
+    1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 1, 1, 1, 1,  // 0x30
+    1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,  // 0x40
+    1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 1, 1, 1,  // 0x50
+    1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,  // 0x60
+    1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,  // 0x70
```
jart, line 32: Don't forget about 0x7F.

tool/net/redbean.c
```diff
@@ -4163,6 +4163,117 @@ static int LuaSetHeader(lua_State *L) {
   return 0;
 }
 
+static void GrowCookieBuffer(char *buf, char *p, size_t *n, size_t increase) {
```
jart, line 4166: The buffering logic here is a little difficult to review.

Some background: I used stpcpy() heavily for the pure C code serving paths because they helped a lot with micro-optimizing every last cycle of performance, so we could hit a million qps. Something like a Lua API has some inherent overhead so it probably isn't worth living as dangerously with the micro-optimizations.

I would be thrilled if you used the new append*() library for this. The functions are all in https://github.com/jart/cosmopolitan/tree/master/libc/stdio They're extremely fast (not quite as fast as stpcpy), well-tested, and well-documented. I'm very proud of them. It took me two years to come up with an API this good. Since if you dig deeper you might notice that there's a whole bunch of other helpers I designed along the way, which turned out to be failures, and are still lingering around as technical debt waiting to be deleted. This new API is safe, easy to review, and easy to use. For example:

```c
char *b=0;
appendf(&b, "%.*s=%.*s", keylen, key, vallen, val);
free(b);
```

Or you could make it 2x faster:

```c
char *b=0;
appendd(&b, key, keylen);
appendw(&b, '=');
appendd(&b, val, vallen);
free(b);
```

jart (changes requested):

net/http/isvalidcookievalue.c
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Returns true if string is a valid cookie value
+ * (any ASCII excluding control char, whitespace,
+ * double quotes, comma, semicolon, and backslash).
+ *
+ * @param n if -1 implies strlen
+ */
+bool IsValidCookieValue(const char *s, size_t n) {
+  size_t i;
+  if (!n) return false;
```
jart, line 31: Empty value should be legal per the RFC.

net/http/isvalidcookievalue.c
pkulchenko: Yes, it's allowed, but only needs to be checked on the receiving side, as the method itself will generate non-quoted `*cookie-octet` (unless I misunderstood what you were suggesting).

net/http/isvalidcookievalue.c
pkulchenko: Good point; I'll add this to redbean.c directly and remove the added table/method files. Thanks!

pkulchenko: Unless you think it's still worthwhile to have `isvalidcookievalue.c `

net/http/khttpcookievalue.c
pkulchenko: This shouldn't apply anymore if `IsCookieOctet` is used

tool/net/redbean.c
pkulchenko: Ha, I saw this in several places, but went with the same logic you have in SetHeader for consistency (this is where the buffering logic is from). After looking at this in more detail, I love this `append*` API! I'll update the code to use it.

net/http/isvalidcookievalue.c
pkulchenko: Yes, fixed that already.

I wonder if I still need `& 0xff` in `!IsCookieOctet(s[i] & 0xff)`?

net/http/isvalidcookievalue.c
jart: Ah I see.

net/http/isvalidcookievalue.c
jart: redbean.c is a monster in terms of lines of code. If it can be easily made a separate file then I'd recommend doing so.

tool/net/redbean.c
jart: 😃

net/http/isvalidcookievalue.c
jart: The implicit type cast to `unsigned char` per the function parameter, will effectively do the same thing. So no. However, even if the parameter type of `IsCookieOctet` was `int` then it still wouldn't matter since the expression logic would still exclude negative numbers caused by the possibility of `char` == `signed char` sign-extension.

pkulchenko: Thank you for the explanation!

tool/net/redbean.c
pkulchenko: Updated; I agree, the code is shorter and easier to read.

pkulchenko:
@jart, don't merge it yet; will add HttpOnly option.

net/http/isvalidcookievalue.c
pkulchenko: Yes, I kept this logic separate.

pkulchenko:
Added HttpOnly. Should be good for merging if all `append*` usage looks good.

jart (approved):
LGTM

Issue #263, opened 2021-09-02
Add APE interpreter


lemaitre:
There is interest in keeping the the binary unaltered after execution, in order to keep it portable, and thus shareable. see: #82 #85 #146 #236

I have two ideas that would go in that direction.

### Repatch

Currently, the APE is a shell script that patches itself in order to be a valid ELF.

What could be done is at the start of the ELF binary, patch back the file to be a valid APE again. Then continue with normal execution.

It would still require write permission to the APE, but at least, it would still be portable afterwards.

### Loader

Another way would be to not patch the APE at all, and create a loader that would mmap the APE in memory, and jump at the starting address.
The loader would be a tiny ELF embedded in the APE, that would be copied into a temp file, and that temp ELF started passing the current APE as a parameter to the loader.
As the loader would be tiny, I think it would be an acceptable solution to copy it every time an APE is executed.

This would solve the write permission issue, but would require to get a temporary folder where a file could be both created and executed. It might be tricky to do in a portable way.

pkulchenko:
@lemaitre, isn't this what's already available in ape-no-modify-self.o according to https://github.com/jart/cosmopolitan/issues/82#issuecomment-815451975?

jart:
Yes we have ape-no-modify-self.o which enables the non-modification use case, at the cost of performance, with caveats. See ape.S for documentation on the matter, plus what's written in the issues.

I don't think we've discussed the idea of creating a loader payload yet. So I've added one to the examples folder. It's relatively straightforward to load an APE executable into memory and execute it. I hope that [examples/loader.c](https://github.com/jart/cosmopolitan/blob/master/examples/loader.c) will add some clarity. You can run it as follows:

```
m=tiny
make -j8 MODE=$m o/$m/examples
o/$m/examples/loader.elf o/$m/examples/printargs.com
```

jart:
Note: We should also probably write a formal specification for the APE format and then upstream patches to the Linux and BSD kernels so they can run it without the script step.

lemaitre:
> @lemaitre, isn't this what's already available in ape-no-modify-self.o according to [#82 (comment)](https://github.com/jart/cosmopolitan/issues/82#issuecomment-815451975)?

It is the same goal, but what I propose is different. As far as I understand, the solution in #82 is to copy the whole APE in a TMP dir, and have this copy patched. The problem here is that if the APE is large, then the copy will also be large.

[...]

PR #267, opened 2021-09-05
WIP: Embed APE loader inside APE

lemaitre:
This is an implementation of the loader proposal from #263.

The idea is to embed inside the APE a tiny ELF that loads the APE in memory, and jump into it to start it. The loader is copied into TMP folder and started from here. The loader is from Justine's example: 969174e155d4bddcec7a2c6c1f4e7e4d3c293f93.

Before considering merging, I would like some feedbacks, especially about how to organize the new code, and where to put it.
Also, I won't be able to create a MacOs loader, so I might need some help here.

### DIFF
--- a/ape/ape.S
+++ b/ape/ape.S
@@ -543,23 +543,21 @@ apesh:	.ascii	"'\n#'\"\n"			# sixth edition shebang
 	.ascii	"o=\"$(command -v \"$0\")\"\n"
 //	Try to use a system-wide APE loader.
 	.ascii	"type ape-loader >/dev/null 2>&1 && "
-	.ascii	"exec ape-loader \"$0\" \"$@\"\n"
+	.ascii	"exec ape-loader \"$o\" \"$@\"\n"
 #ifdef APE_LOADER
 //	There is no system-wide APE loader, but there is one
 //	embedded inside the APE. So if the system is not MacOs,
 //	extract the loader into a temp folder, and use it to
 //	load the APE without modifying it.
 	.ascii	"if [ ! -d /Applications ]; then\n"
-	.ascii	  "t=\"${TMPDIR:-/tmp}/ape-loader.$$\"\n"
-	.ascii	  "dd if=\"$o\" of=\"$t\" skip=\""
+	.ascii	  "t=\"${TMPDIR:-/tmp}/ape-loader\"\n"
+	.ascii	  "[ -x \"$t\" ] || {\ndd if=\"$o\" of=\"$t\" skip=\""
 	.shstub	    ape_loader_dd_skip,2
 	.ascii	    "\" count=\""
 	.shstub	    ape_loader_dd_count,2
-	.ascii	    "\" bs=\""
-	.shstub	    ape_loader_dd_bs,2
-	.ascii	    "\" 2>/dev/null\n"
-	.ascii	  "chmod 700 \"$t\"\n"
-	.ascii	  "exec \"$t\" \"$0\" \"$@\"\n"
+	.ascii	    "\" bs=64 2>/dev/null &&\n"
+	.ascii	  "chmod 700 \"$t\"\n}\n"
+	.ascii	  "exec \"$t\" \"$o\" \"$@\"\n"
 	.ascii	"fi\n"
 #endif
 #ifndef APE_NO_MODIFY_SELF
--- a/ape/ape.S
+++ b/ape/ape.S
@@ -540,11 +540,32 @@ apesh:	.ascii	"'\n#'\"\n"			# sixth edition shebang
 //	modify the binary to follow the local system's convention.
 //	There isn't a one-size-fits-all approach for this, thus we
 //	present two choices.
+	.ascii	"o=\"$(command -v \"$0\")\"\n"
+//	Try to use a system-wide APE loader.
+	.ascii	"type ape-loader >/dev/null 2>&1 && "
+	.ascii	"exec ape-loader \"$0\" \"$@\"\n"
+#ifdef APE_LOADER
+//	There is no system-wide APE loader, but there is one
+//	embedded inside the APE. So if the system is not MacOs,
+//	extract the loader into a temp folder, and use it to
+//	load the APE without modifying it.
+	.ascii	"if [ ! -d /Applications ]; then\n"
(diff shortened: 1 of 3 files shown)


ape/ape.lds
```diff
@@ -478,6 +485,17 @@ HIDDEN(ape_stack_filesz = 0);
 HIDDEN(ape_stack_memsz = STACKSIZE);
 HIDDEN(ape_stack_align = 16);
 
+HIDDEN(ape_loader_offset = ape_rom_offset + ape_loader - ape_rom_vaddr);
```
jart, line 488: You wouldn't need to define an ELF segment for the loader payload. All the linker script needs to do is something like this:

```
HIDDEN(ape_loader_b0 = RVA(ape_loader) % 10 + 48);
HIDDEN(ape_loader_b1 = RVA(ape_loader) < 10 ? 32 : RVA(ape_loader) / 10 % 10 + 48);
HIDDEN(ape_loader_b2 = RVA(ape_loader) < 100 ? 32 : RVA(ape_loader) / 100 % 10 + 48);
HIDDEN(ape_loader_b3 = RVA(ape_loader) < 1000 ? 32 : RVA(ape_loader) / 1000 % 10 + 48);
HIDDEN(ape_loader_b4 = RVA(ape_loader) < 10000 ? 32 : RVA(ape_loader) / 10000 % 10 + 48);
HIDDEN(ape_loader_b5 = RVA(ape_loader) < 100000 ? 32 : RVA(ape_loader) / 100000 % 10 + 48);
HIDDEN(ape_loader_b6 = RVA(ape_loader) < 1000000 ? 32 : RVA(ape_loader) / 1000000 % 10 + 48);
HIDDEN(ape_loader_b7 = RVA(ape_loader) < 10000000 ? 32 : RVA(ape_loader) / 10000000 % 10 + 48);
HIDDEN(ape_loader_b8 = RVA(ape_loader) < 100000000 ? 32 : RVA(ape_loader) / 100000000 % 10 + 48);
HIDDEN(ape_loader_b9 = RVA(ape_loader) < 1000000000 ? 32 : RVA(ape_loader) / 1000000000 % 10 + 48);
```

You'd also want to define a section... It should probably go right before `.zip` in the linker script, e.g.

```ld-script
  .payload . : {
    KEEP(*(.payload))
  }
```

ape/ape.S
```diff
@@ -645,6 +666,11 @@ apesh:	.ascii	"'\n#'\"\n"			# sixth edition shebang
 	.ascii	"fi\n"
 	.ascii	"exit $R\n"
 	.endobj	apesh
+#ifdef APE_LOADER
+        .section .ape.loader,"a",@progbits
+        .incbin APE_LOADER
```
jart, line 671: Nice

ape/ape.S
```diff
@@ -540,11 +540,32 @@ apesh:	.ascii	"'\n#'\"\n"			# sixth edition shebang
 //	modify the binary to follow the local system's convention.
 //	There isn't a one-size-fits-all approach for this, thus we
 //	present two choices.
+	.ascii	"o=\"$(command -v \"$0\")\"\n"
+//	Try to use a system-wide APE loader.
+	.ascii	"type ape-loader >/dev/null 2>&1 && "
+	.ascii	"exec ape-loader \"$0\" \"$@\"\n"
+#ifdef APE_LOADER
+//	There is no system-wide APE loader, but there is one
```
jart, line 548: 1. I would recommend having this only happen for the `ape-no-modify-self.o` build.
2. You might want to check `$(uname -s) = Linux`

jart (changes requested):
Welcome as a new contributor. We need to go through one quick hurdle before we can proceed. Could you please email Justine Tunney <jtunney@gmail.com> and say that you intend to assign her the copyright
to the changes you contribute to Cosmopolitan? See [CONTRIBUTING.md](CONTRIBUTING.md) for further details. We make the ask because it helps us ship the tiniest most legally permissive prim and proper binaries.
You also only need to do it one time, so any future changes you send us can be quickly merged.

ape/ape.lds
lemaitre: Isn't what `SHSTUB2` is for? Apart from that, I can remove other definitions.

Will do the separate `.payload ` section. I was just not really sure this would not mess up the computation of the offset inside the file, but that should be fine as long as I deal with LOADADDR, right?

ape/ape.S
lemaitre: 1.
I was in the impression that, if a system-wide ape-loader exist, it is always the right thing to do.
The copy of the embedded loader is done only if `APE_LOADER` is defined, which, I expect, happens only for `ape-no-modify-self.o`.

I need a separate macro, as `APE_LOADER` gives the path of the loader ELF.

2.
I just copied the test for Linux/MacOS which was already here. The check for the existence of a directory is faster than calling `uname` in a subshell.

lemaitre:
> Could you please email Justine Tunney [jtunney@gmail.com](mailto:jtunney@gmail.com) and say that you intend to assign her the copyright
> to the changes you contribute to Cosmopolitan?

I've just sent it to you from my personal email.

lemaitre:
I've ran some benchmarks about how fast the loader is:

|    native | direct loader | shell loader | shell loader + copy |
|-----------:|-----------------:|----------------:|--------------------------:|
| 1.52 ms |       1.57 ms |      8.07 ms |                   15.7 ms |

The binary tested is `hello.com` (with stdout redirected to `/dev/null`).
So the loader by itself is really fast and have a negligible impact.
However, the shell trick alone account for about 5-6 ms, and the copy of the loader takes another 7 ms.

All in all, I think it's reasonable, especially because this overhead does not increase with the size of the APE.

jart (approved):
LGTM Thank You!

> especially because this overhead does not increase with the size of the APE.

That's definitely the best part!

The 5ms constant factor can be using something like binfmt_misc or, more ideally, getting ape support merged into kernels.

lemaitre:
@jart I know you were eager to merge this, but I think it was not ready yet (hence the "WIP" prefix in the name).
To me there were a few questions that needed answers:
- Should we have a test for a system-wide `ape-loader` before trying anything else?
- When copying the loader in the temp folder, should we randomize the destination a bit?
- Should we keep the old code for copying the whole APE (in case the loader is not embedded)?

Maybe I could just create an issue to discuss this, and the another MR when we have answers.

PR #268, opened 2021-09-05
Add error reporting to redbean Slurp

pkulchenko:
This should allow `content = assert(Slurp(filename))` to work as
expected and report an error if file doesn't exist or can't be read.

@jart, I'm wondering if we can parse the Commentary field in `sysv/consts.sh` (or add a column) to get better error descriptions to use in `strerror/%m` instead of errno names (and keep errno names as the fallback)?. It'll slightly increase the exe size (maybe skip it for TINY), but should be quite manageable and make the error reports more user friendly.

### DIFF
--- a/tool/net/redbean.c
+++ b/tool/net/redbean.c
@@ -4639,12 +4639,18 @@ static int LuaEncodeLatin1(lua_State *L) {
 }
 
 static int LuaSlurp(lua_State *L) {
-  char *p;
+  char *p, *f;
   size_t n;
-  p = xslurp(luaL_checkstring(L, 1), &n);
-  lua_pushlstring(L, p, n);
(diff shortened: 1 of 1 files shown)


jart:
We could add all the errno descriptions to the redbean txt file.

jart:
There's also a finite number of errnos open() and read() are going to return. We could special case those in this function if you want.

jart (approved):

pkulchenko:
> There's also a finite number of errnos open() and read() are going to return. We could special case those in this function if you want.

True. but the same logic would need to be applied everywhere where a similar handling would be useful. I think what we have now is fine, but we may need to revisit if there is more demand.

PR #271, opened 2021-09-16
Fix SSE scalar ops

pkulchenko:
I'm not 100% sure, but I think the patch is correct according to my reading of the [Intel Intrinsics Guide](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#expand=3,4882,4891,4944,193,7003&text=_mm_sub_ss).


jart (approved):
Thanks for spotting this!

PR #275, opened 2021-09-18
Add Argon2 implementation

pkulchenko:
@jart, @ahgamut, this adds Argon2 implementation to cosmopolitan/redbean, but I did run into a couple of issues that I'm hoping you can help me with.

I added intrinsics to `bits/zmmintrin.internal.S` and function declarations to `bits/emmintrin.internal.h`. I also added `LIBC_BITS` to argon2.mk, but am still getting `error: _mm_mul_epu32 (o/rel/third_party/argon2/opt.o) not declared by direct deps of o/rel/third_party/argon2/argon2.a.pkg` and am not sure why, as I see `bits.a.pkg` on the list of packages in that error message and it should be enough to get `_mm_mul_epu32`, no? Any ideas on how to fix this?

@jart, also, I used existing .S files as a template, but got a message about unrecognized pseudo-ops (any that starts with `.`), even though they work just fine in all existing .S files. I ended up removing all of them, but am curious as to what I'm doing wrong.


pkulchenko:
@jart, I got Argon2 to compile with the SSE2 ops provided in 670e6d46; I'll run some more tests tomorrow and provide an update.

Based on my reading of the Argon2 code that uses these intrinsics, it expects that the parameters are not going to be affected (and this is how I implemented those that I added), but it loos like this is not the case with some of the existing ones (for example, `_mm_add_sd` will change the values of the first parameter, which is also being returned). Should the existing once be updated to not clobber their parameters?

pkulchenko:
@jart, I completed the updates and testing and everything looks good so far. I was trying to get the "optimized" version to work, but it was triggering SIGSERV in some cases for me, so I decided to go back to the reference implementation and haven't seen any issues there. I also checked the generated hashes on available test cases and everything looks good. The binary size somehow only increased by less than 50k (for redbean). I left the BITS module updates in, even though they are not used in the reference implementation (as those intrinsics may still be needed and should provide a blueprint for more; assuming they are correct).

pkulchenko:
Here is an example that can be included in redbean's code:

```
local argon2 = require "argon2"
local encoded = assert(argon2.hash_encoded("password", "somesalt"))
-- Log(kLogInfo, encoded)
-- should print "$argon2id$v=19$m=4096,t=3,p=1$c29tZXNhbHQ$qLml5cbqFAO6YxVHhrSBHP0UWdxrIxkNcM8aMX3blzU"

local ok, err = argon2.verify(encoded, "password")
if err then
  error("could not verify: " .. err)
end

if not ok then
  error("The password does not match the supplied hash")
end
```

pkulchenko:
Another quick update. The binary increase is actually only 20k and the password generation with default settings from Redbean takes about 15-17ms on my underpowered laptop.

jart (approved):
LGTM. Thank you!

PR #279, opened 2021-09-28
Add clock polyfill on Windows

pkulchenko:
This helps if there is no LOCK_PROCESS_CPUTIME_ID on Windows.

This patch works, but it works too well, as it shows kernel+user time, but `clock()` on windows actually shows the wall time ("The clock function tells how much wall-clock time has passed since the CRT initialization during process start." from https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/clock) and that's why I suggested `CLOCK_MONOTONIC_RAW` minus the same value saved during the process initialization. Should I update the implementation?


jart (approved):
Thank you!

PR #281, opened 2021-10-01
speedups from pyston again

ahgamut:
removed (ie hid behind `#if IsModeDbg()`) the recursion checking methods,
and the memory hooks + allocator methods. also updated all the
testcases that might involve these debugging features. 

References: 
* https://github.com/pyston/pyston/pull/7
* https://github.com/pyston/pyston/commit/9ed83c40855b57c10988f76770a4eb825e034cd8 because these parts have changed in later versions of CPython (more allocator-related stuff), so needed to check which parts could be removed

No new tests were added to `python.mk`.
All existing tests pass without error in `MODE=` and `MODE=asan`, but `MODE=dbg` is failing for some reason (will post backtrace).

### DIFF
--- a/third_party/mbedtls/mbedtls.mk
+++ b/third_party/mbedtls/mbedtls.mk
@@ -58,6 +58,10 @@ o/$(MODE)/third_party/mbedtls/everest.o:				\
 			OVERRIDE_CFLAGS +=				\
 				-O3
 
+o/$(MODE)/third_party/mbedtls/des.o:				\
+			OVERRIDE_CFLAGS +=				\
+				-DSTACK_FRAME_UNLIMITED
+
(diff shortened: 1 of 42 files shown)


ahgamut:
The build fails with `MODE=dbg` beacuse `pyobj.com` is unable to flush `stderr` properly when processing `_pyio.py` -- `Py_CheckFunctionResult` triggers the error. This didn't happen in the other two builds because `Py_CheckFunctionResult` is disabled there.

2021-10-02_02-25-45_1349x341


The same error also occurs when processing `random.py` and `plistlib.py`.

jart (approved):
Nice! Thank you! I'll look into the MODE=dbg issue right away and report back.

Issue #277, opened 2021-09-26
Design consideration for implementing pthread


lemaitre:
I am planning to implement the pthread interface for cosmopolitan.

I think I can make it work with a small memory footprint by having most functions using a portable implementation (centered around atomics). The few operations that could not be portable are basically: `thread_create`, `atomic_wait` (futex) and maybe `thread_detach` and `tls_init`.
All the other operations could be (in theory) implemented with portable atomics on top of those.

I already did a POC outside cosmopolitan and thanks to the `clone` syscall, the creation of a thread on Linux is ~30 instructions.

So now I have a few questions:
- Do we need to support TLS of dynamic modules? (loaded with dlopen)
- Where should go `.tdata` and `.tbss` sections?
- Is there any interest in having pthread cancellation points?
- Should we add signal masks and CPU affinities?
- Do we want fair locks?
- Do we want thread stack cache?
- Do we want a multithreaded malloc being a part of this interface?

lemaitre:
Here are my own answers to those questions:

> Do we need to support TLS of dynamic modules? (loaded with dlopen)

I would say yes, but I would prefer no.

> Where should go `.tdata` and `.tbss` sections?

Just after `.data`?

> Is there any interest in having pthread cancellation points?

I'm not sure to see what is the use of cancellation points. Maybe we could just kill the thread?

> Should we add signal masks and CPU affinities?

I don't think it is necessary at the beginning.

> Do we want fair locks?

I personally don't. I think most of the time, a fast lock is better than a fair lock. And fast locks are much simpler to implement than fair ones.

> Do we want thread stack cache?

I would say no: if you have a program that would benefit from such a cache, you are probably even better to just reuse your threads instead of creating/deleting them all the time.

> Do we want a multithreaded malloc being a part of this interface?

I would say yes, but I would appreciate some help on that. So maybe not for a start.

pkulchenko:
>>  Do we need to support TLS of dynamic modules? (loaded with dlopen)
> I would say yes, but I would prefer no.

There is no current dlopen support in cosmopolitan, so I'd say we can cross that bridge when we get there.

PR #282, opened 2021-10-03
[WIP] Threading

lemaitre:
This is the PR associated to #277.

The goal is to implement the pthread interface (or at least similar) for cosmopolitan.
The key observation that makes it possible to consider such a daunting task is the following:
It seems that there are really few part of the code that is OS dependent, namely OS thread creation, OS thread destruction, futex, and TLS.

The complex parts are built on top of those, and on top of atomics which are OS independent.

The goal of this PR is to focus on the portable abstraction, and make it works on Linux (because that's the platform I know).

Status:

- [x] thread creation
- [x] thread exit
- [x] detach
- [x] join
- [ ] kill
- [ ] cancel
- [ ] mutex
- [ ] rw_lock
- [ ] semaphore
- [ ] condition variable
- [x] TLS setup
- [ ] TLS allocation
- [ ] TLS for main thread
- [ ] dynamic TLS
- [x] custom stack size
- [ ] custom stack
- [ ] thread-safe libc
- [ ] actual pthread interface
- [ ] anything I forgot


jart (changes requested):
Could you relocate libc/runtime/clone.c into this new `LIBC_THREAD` package? Then we'll have Windows and NetBSD covered too. I like the choice of API naming. I would also recommend that we implement mutexes using https://github.com/google/nsync

lemaitre:
I can definitely move `libc/runtime/clone.c` into this new package, but I'm not sure what factorization could be done.

For mutexes, I was thinking more about what is used in webkit: https://webkit.org/blog/6161/locking-in-webkit/
Especially the futex "emulation" in userspace part which could greatly reduce the os dependency of this code (apparently not all systems have a nice drop-in to futex).

jart:
Nsync was handwritten by the guy who built all the locking infrastructure at Google. Before that he invented Altavista. Before that he invented Bzip2. This guy really doesn't evangelize his work so not many people know about it. 

OS dependencies are toilsome but it isn't something I fear, since that's our bread and butter. Cosmopolitan libraries let users to do things like copy a file twice as fast as their system's `cp` command because we did all the hard work of figuring out and abstracting the optimum systems integrations. On Linux I'm reasonably certain that for mutexes that entails futex. Even if we only support just Linux threads for the present we can iterate on it and improve. I just want to be sure that whatever we do choose to do, is the best.

jart:
Thanks for adding futexes!

I'd like to move forward with merging since even though it's a work in progress it's a step in the right direction. I can provide some assistance and collaboration on this effort. You may see me making follow up changes to what you've contributed. Simple things like polyfills, style, and integration so that your project has a good home in this codebase. Think of it as me volunteering for your project. However I want it to be *your* project and I want to be able to tell anyone who asks about Cosmopolitan Threads that you're leading effort.

PR #292, opened 2021-10-15
minimize Python startup imports

ahgamut:
attempt to shave off a few more ms in startup time:

avoiding sysconfig is similar to [bpo-29585](https://bugs.python.org/issue29585) ie https://github.com/python/cpython/commit/a8f8d5b4bd30dbe0828550469d98f12d2ebb2ef4

the modification to the stat/os modules is one of the hidden advantages of Actually Portable Python: since Cosmopolitan provides a POSIX interface, some of the platform checks done by Python can be moved internally to C, or avoided entirely. I'm pretty sure similar shortcuts exist for other modules as well.

### DIFF
--- a/third_party/python/Lib/os.py
+++ b/third_party/python/Lib/os.py
@@ -38,7 +38,7 @@ def _exists(name):
     return name in globals()
 
 def _get_exports_list(module):
-    return getattr(module, "__all__", (n for n in dir(module) if n[0] != '_'))
+    return list(getattr(module, "__all__", (n for n in dir(module) if n[0] != '_')))
 
 name = 'posix'
 linesep = '\n'


ahgamut:
@jart there is an error in `frozen.c` if I try to build `MODE=tiny` without first building `MODE=`

With this PR, the following modules are the only Python files loaded during startup:

```
import 'codecs' # <_frozen_importlib_external.SourcelessFileLoader object at 0x1000800e4390>
import 'encodings.aliases' # <_frozen_importlib_external.SourcelessFileLoader object at 0x100080167cf8>
import 'encodings' # <_frozen_importlib_external.SourcelessFileLoader object at 0x1000800d9eb8>
import 'encodings.utf_8' # <_frozen_importlib_external.SourcelessFileLoader object at 0x100080175b70>
import 'encodings.latin_1' # <_frozen_importlib_external.SourcelessFileLoader object at 0x1000801785f8>
import '_weakrefset' # <_frozen_importlib_external.SourcelessFileLoader object at 0x100080181550>
import 'abc' # <_frozen_importlib_external.SourcelessFileLoader object at 0x100080178ba8>
import 'io' # <_frozen_importlib_external.SourcelessFileLoader object at 0x1000801787f0>
import '_bootlocale' # <_frozen_importlib_external.SourcelessFileLoader object at 0x10008018abe0>
import 'stat' # <_frozen_importlib_external.SourcelessFileLoader object at 0x1000801a9dd8>
import 'genericpath' # <_frozen_importlib_external.SourcelessFileLoader object at 0x1000801bb550>
import 'ntpath' # <_frozen_importlib_external.SourcelessFileLoader object at 0x1000801b3588>
import 'posixpath' # <_frozen_importlib_external.SourcelessFileLoader object at 0x1000801ae7b8>
import '_collections_abc' # <_frozen_importlib_external.SourcelessFileLoader object at 0x1000801bbba8>
import 'os' # <_frozen_importlib_external.SourcelessFileLoader object at 0x10008019a4a8>
import '_sitebuiltins' # <_frozen_importlib_external.SourcelessFileLoader object at 0x10008019a400>
import 'site' # <_frozen_importlib_external.SourcelessFileLoader object at 0x10008018aef0>
```

Are further speedups possible?

* `codecs`/`encodings` have no direct speedups
* `weakrefset` can possibly be ported to C (add to `_weakref`?)
* `abc` can be ported to C. I think this should be done, because Python3.7 did it: see [bpo-31333](https://bugs.python.org/issue31333)
* `io` has no direct speedups
* `_bootlocale` has just one function
* `stat` is almost empty at this point, possibly it can be removed and just rename `_stat` to `stat`
* `*path` modules not sure what can be removed, need to examine
* `_collections_abc` uses `abc` heavily, and is imported by `os` for just `MutableMapping`. I tried to avoid the import and use `dict` instead of `MutableMapping`, but that didn't work (some unrelated `optparse` test failed).
* `_sitebuiltins` is separate from `site` due to a refcount thing, can possibly move into C if it is a major slowdown
* `site` itself has some code that is ignored because of APE features, so maybe those can be removed

There are some `try/except` blocks in many places that are mostly useless within APE Python, removing those ought to help a little.

PR #306, opened 2021-10-27
python 3.7 backports

ahgamut:
python/cpython@7fc252adfbedece75f2330bcfdadbf84dee7836f

(mixed up the symbols in the commit message)


jart (approved):
LGTM.

Compatibility is nice to have but having a better Python is more important. We don't need to carry the same burdens as the core CPython team yet since we're still a new project. We should enjoy that freedom while it lasts. Once we have the website online we can also document the things we changed, the things we backported, etc.

PR #308, opened 2021-11-01
Redbean getstatus and more

pkulchenko:
@jart, I added and renamed a couple of redbean methods. Feel free to remove renaming commits, but I think it looks more consistent with the new names. I marked those doing renames with "breaking change" comment in the commit message.

### DIFF
--- a/tool/net/help.txt
+++ b/tool/net/help.txt
@@ -709,6 +709,10 @@ FUNCTIONS
    GetScheme() → str
            Returns scheme from Request-URL, if any.
 
+   GetStatus() → int
+           Returns current status (as set by an earlier SetStatus call) or
+           `nil` if the status hasn't been set yet.
+
    GetTime() → seconds:number
            Returns current time as a UNIX timestamp with 0.0001s precision.
 
--- a/tool/net/help.txt
+++ b/tool/net/help.txt
@@ -592,6 +592,10 @@ FUNCTIONS
            Turns integer like 0x01020304 into a string like 1.2.3.4. See also
            ParseIp for the inverse operation.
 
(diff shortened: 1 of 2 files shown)


jart (changes requested):
I'm happy to do this, but only if old names continue to be defined so existing code doesn't break. Also I intend to leave feedback on your design design ideas shortly in #97. Thanks you so much for your patience!

jart:
With regard to breaking changes, please note the logging macros we changed earlier weren't publicly documented on the website, so they were strictly speaking an internal api and the commit message was intended as a courtesy to anyone who might have forked the cosmopolitan repo for their own internal needs, as a matter of courtesy. On the other hand APIs like `GetPayLoad()` are publicly and obviously documented on https://redbean.dev/ and are likely to be used. In the past, when renaming redbean functions like popcnt() to Popcnt() I kept the old ones around since there's little cost to doing so and it's a clear win for the user. Ultimately these things boil down to using our best judgement and consensus.

pkulchenko:
Good point; I restored the two renamed methods and marked as "deprecated" in the docs.

jart (approved):
Looks good. It would also be sufficient to simply say ` (old name: GetPayload)`, but even that isn't strictly necessary. So long as old code written against publicly documented apis continues to run, we've done our job in terms of securing the user's time investments and can therefore evolve the best practices as much as we wish.

jart:
Also for reference, here are links to the old versions of the docs maintained on the website:

- https://redbean.dev/1.3.html
- https://redbean.dev/0.4.html
- https://redbean.dev/0.3.html

I intend to keep doing that for each release to supplement the version control information.

PR #311, opened 2021-11-01
Don't pass unused args to printf in brailledump

jokva:
Justine,

I ported bd.c (as an exercise) and found a printf with too many arguments. I'm not sure if it would ever cause problems, or if it is gracefully handled in cosmopolitan (glibc didn't seem to give me any issues at least, but to be fair I didn't try to poke at it much).

--

The final print does not print any content, only the byte-offset of the
end-block, which makes the A, B params unused. From gcc:

    bd.c:77:17: warning: too many arguments for format [-Wformat-extra-args]
        if (o) printf("%08x\n", o, A, B);

### DIFF
--- a/tool/viz/bd.c
+++ b/tool/viz/bd.c
@@ -77,6 +77,6 @@ int main(int argc, char *argv[]) {
     }
     o += n;
   }
-  if (o) printf("%08x\n", o, A, B);
+  if (o) printf("%08x\n", o);
   return !feof(f);
 }


jart (approved):
LGTM. Thank you for spotting this and taking the time to send a PR. The extra args are harmless from an ABI standpoint but it's confusing from a readability standpoint, so this fix is good to have.

Also, welcome as a new contributor. We need to go through one quick hurdle before we can proceed. Could you please email Justine Tunney <jtunney@gmail.com> and say that you intend to assign her the copyright to the changes you contribute to Cosmopolitan? See [CONTRIBUTING.md](CONTRIBUTING.md) for further details. We make the ask because it helps us ship the tiniest most legally permissive prim and proper binaries. You also only need to do it one time, so any future changes you send us can be quickly merged.

Thanks!

jokva:
I have sent you an email.

jart:
Cool confirmed. Thanks again for the fix.

PR #312, opened 2021-11-02
change noinline to dontinline

ahgamut:
third-party libraries often use __attribute__((noinline))
which clashes with the defined macro noinline in libc/integral/c.inc.
this commit changes the noinline macro to dontinline (similar to
wontreturn) to avoid the clash.

the chibicc parser checks for attributes by discarding the underscores
first. the only place where noinline shouldn't be replaced.


jart (approved):
Looks good. Thank you for sending this!

PR #317, opened 2021-11-04
Backporting METH_FASTCALL from Python 3.7

ahgamut:
All commits have reference commit from cpython linked in the message.

A couple of backports are partial (IIRC the ones with OrderedDict and RecursiveCall) because the linked commits are not isolated changes; they build on some features tangentially related to METH_FASTCALL.

I'm still wary about compatibility (dictionary lookup has changed internally from 3.6 to 3.7, got segfaults when I attempted to port python/cpython@7dc6a5f50aa). However, all existing tests pass under `MODE=`, `MODE=dbg`, `MODE=opt`, and `MODE=optlinux`, so I expect backporting `METH_FASTCALL` is still ok.

`python.com` under `MODE=optlinux` is now 1.05x faster than vanilla Python 3.6.15.

| Benchmark               | py36-15  | APE-optlinux       | APE-optlinux-fastcall  |
|-------------------------|:--------:|:----------------------:|:----------------------:|
| Geometric mean          | (ref)    | 1.02x faster           | 1.05x faster           |
| chaos                   | 247 ms   | 242 ms: 1.02x faster   | 239 ms: 1.03x faster   |
| crypto_pyaes            | 214 ms   | 212 ms: 1.01x faster   | 213 ms: 1.01x faster   |
| deltablue               | 17.2 ms  | 16.3 ms: 1.05x faster  | 16.0 ms: 1.07x faster  |
| fannkuch                | 997 ms   | 1.58 sec: 1.59x slower | 1.02 sec: 1.03x slower |
| float                   | 228 ms   | 221 ms: 1.03x faster   | 213 ms: 1.07x faster   |
| go                      | 555 ms   | 524 ms: 1.06x faster   | 515 ms: 1.08x faster   |
| hexiom                  | 22.7 ms  | 22.3 ms: 1.02x faster  | 22.1 ms: 1.03x faster  |
| json_dumps              | 24.7 ms  | 27.2 ms: 1.10x slower  | 27.0 ms: 1.09x slower  |
| json_loads              | 51.5 us  | 48.3 us: 1.07x faster  | 47.2 us: 1.09x faster  |
| logging_format          | 23.7 us  | 27.0 us: 1.14x slower  | 26.2 us: 1.11x slower  |
| logging_silent          | 745 ns   | 697 ns: 1.07x faster   | 692 ns: 1.08x faster   |
| logging_simple          | 20.4 us  | 23.4 us: 1.15x slower  | 22.8 us: 1.12x slower  |
| meteor_contest          | 199 ms   | 202 ms: 1.02x slower   | 201 ms: 1.01x slower   |
| nbody                   | 245 ms   | 249 ms: 1.02x slower   | 244 ms: 1.01x faster   |
| nqueens                 | 206 ms   | 199 ms: 1.03x faster   | 202 ms: 1.02x faster   |
| pathlib                 | 39.5 ms  | 36.9 ms: 1.07x faster  | not significant        |
| pickle                  | 20.1 us  | 17.3 us: 1.16x faster  | 17.9 us: 1.12x faster  |
| pickle_dict             | 57.5 us  | 39.5 us: 1.45x faster  | 39.4 us: 1.46x faster  |
| pickle_list             | 8.10 us  | 6.19 us: 1.31x faster  | 6.20 us: 1.31x faster  |
| pickle_pure_python      | 1.07 ms  | 994 us: 1.08x faster   | 978 us: 1.10x faster   |
| pidigits                | 293 ms   | 293 ms: 1.00x faster   | 292 ms: 1.00x faster   |
| pyflate                 | 1.45 sec | not significant        | 1.47 sec: 1.01x slower |
| python_startup          | 17.3 ms  | 17.4 ms: 1.01x slower  | 15.9 ms: 1.09x faster  |
| python_startup_no_site  | 10.8 ms  | 10.5 ms: 1.03x faster  | 10.4 ms: 1.04x faster  |
| raytrace                | 1.23 sec | 1.14 sec: 1.07x faster | 1.14 sec: 1.08x faster |
| regex_compile           | 384 ms   | 364 ms: 1.06x faster   | 365 ms: 1.05x faster   |
| regex_dna               | 281 ms   | 288 ms: 1.02x slower   | 286 ms: 1.02x slower   |
| regex_effbot            | 4.97 ms  | 5.00 ms: 1.01x slower  | 4.86 ms: 1.02x faster  |
| regex_v8                | 43.1 ms  | 43.5 ms: 1.01x slower  | 41.9 ms: 1.03x faster  |
| richards                | 171 ms   | 163 ms: 1.05x faster   | 161 ms: 1.06x faster   |
| scimark_fft             | 624 ms   | 617 ms: 1.01x faster   | 632 ms: 1.01x slower   |
| scimark_lu              | 486 ms   | 473 ms: 1.03x faster   | 440 ms: 1.11x faster   |
| scimark_monte_carlo     | 222 ms   | 219 ms: 1.01x faster   | 215 ms: 1.04x faster   |
| scimark_sor             | 466 ms   | 456 ms: 1.02x faster   | 445 ms: 1.05x faster   |
| scimark_sparse_mat_mult | 7.57 ms  | 6.97 ms: 1.09x faster  | not significant        |
| spectral_norm           | 254 ms   | 258 ms: 1.01x slower   | 261 ms: 1.03x slower   |
| telco                   | 14.3 ms  | 13.2 ms: 1.08x faster  | 10.9 ms: 1.30x faster  |
| unpack_sequence         | 85.1 ns  | 91.1 ns: 1.07x slower  | 90.9 ns: 1.07x slower  |
| unpickle                | 28.2 us  | 25.7 us: 1.10x faster  | 24.7 us: 1.14x faster  |
| unpickle_list           | 7.27 us  | 7.10 us: 1.02x faster  | 7.09 us: 1.03x faster  |
| unpickle_pure_python    | 796 us   | 750 us: 1.06x faster   | 705 us: 1.13x faster   |
| xml_etree_parse         | 277 ms   | 263 ms: 1.05x faster   | 262 ms: 1.05x faster   |
| xml_etree_iterparse     | 222 ms   | 213 ms: 1.04x faster   | 204 ms: 1.09x faster   |
| xml_etree_generate      | 223 ms   | 208 ms: 1.07x faster   | 204 ms: 1.09x faster   |
| xml_etree_process       | 178 ms   | 174 ms: 1.02x faster   | 171 ms: 1.04x faster   |

### DIFF
--- a/third_party/python/Include/abstract.h
+++ b/third_party/python/Include/abstract.h
@@ -19,11 +19,12 @@ PyObject *_PyStack_AsTupleSlice(PyObject **stack, Py_ssize_t nargs,
 
 PyObject *_PyStack_AsDict(PyObject **values, PyObject *kwnames);
 
-PyObject ** _PyStack_UnpackDict(
+int _PyStack_UnpackDict(
     PyObject **args,
     Py_ssize_t nargs,
     PyObject *kwargs,
-    PyObject **kwnames,
+    PyObject ***p_stack,
+    PyObject **p_kwnames,
     PyObject *func);
 
 PyObject *_PyObject_FastCallDict(PyObject *func, PyObject **args,
--- a/third_party/python/Include/abstract.h
+++ b/third_party/python/Include/abstract.h
@@ -14,6 +14,9 @@ PyObject *PyObject_Call(PyObject *callable_object, PyObject *args,
 
 #ifndef Py_LIMITED_API
 PyObject *_PyStack_AsTuple(PyObject **stack, Py_ssize_t nargs);
+PyObject *_PyStack_AsTupleSlice(PyObject **stack, Py_ssize_t nargs,
+                                Py_ssize_t start, Py_ssize_t end);
+
 PyObject *_PyStack_AsDict(PyObject **values, PyObject *kwnames);
 
 PyObject ** _PyStack_UnpackDict(
--- a/third_party/python/Include/abstract.h
+++ b/third_party/python/Include/abstract.h
@@ -42,6 +42,14 @@ PyObject *_PyObject_FastCallKeywords(PyObject *func, PyObject **args,
 PyObject *_PyObject_Call_Prepend(PyObject *func, PyObject *obj, PyObject *args,
                                  PyObject *kwargs);
 
+#define _PY_FASTCALL_SMALL_STACK 5
+
+PyObject *_PyObject_FastCall_Prepend(
+    PyObject *callable,
+    PyObject *obj,
+    PyObject **args,
+    Py_ssize_t nargs);
+
 #if IsModeDbg()
 PyObject *_Py_CheckFunctionResult(PyObject *func, PyObject *result,
                                   const char *where);
--- a/third_party/python/Include/abstract.h
+++ b/third_party/python/Include/abstract.h
@@ -24,8 +24,7 @@ int _PyStack_UnpackDict(
     Py_ssize_t nargs,
     PyObject *kwargs,
     PyObject ***p_stack,
-    PyObject **p_kwnames,
-    PyObject *func);
+    PyObject **p_kwnames);
 
 PyObject *_PyObject_FastCallDict(PyObject *func, PyObject **args,
                                  Py_ssize_t nargs, PyObject *kwargs);
--- a/third_party/python/Include/methodobject.h
+++ b/third_party/python/Include/methodobject.h
@@ -93,6 +93,12 @@ typedef struct {
     PyObject    *m_module; /* The __module__ attribute, can be anything */
     PyObject    *m_weakreflist; /* List of weak references */
 } PyCFunctionObject;
+PyObject * _PyMethodDef_RawFastCallDict(
+    PyMethodDef *method,
+    PyObject *self,
+    PyObject **args,
+    Py_ssize_t nargs,
+    PyObject *kwargs);
 #endif
 
 int PyCFunction_ClearFreeList(void);
--- a/third_party/python/Include/modsupport.h
+++ b/third_party/python/Include/modsupport.h
@@ -47,7 +47,8 @@ PyObject * Py_BuildValue(const char *, ...);
 PyObject * _Py_BuildValue_SizeT(const char *, ...);
 
 #ifndef Py_LIMITED_API
-int _PyArg_NoKeywords(const char *funcname, PyObject *kw);
+int _PyArg_NoKeywords(const char *funcname, PyObject *kwargs);
+int _PyArg_NoStackKeywords(const char *funcname, PyObject *kwnames);
 int _PyArg_NoPositional(const char *funcname, PyObject *args);
 #endif
 PyObject * Py_VaBuildValue(const char *, va_list);
--- a/third_party/python/Include/modsupport.h
+++ b/third_party/python/Include/modsupport.h
@@ -57,6 +57,12 @@ int _PyArg_UnpackStack(
 int _PyArg_NoKeywords(const char *funcname, PyObject *kwargs);
 int _PyArg_NoStackKeywords(const char *funcname, PyObject *kwnames);
 int _PyArg_NoPositional(const char *funcname, PyObject *args);
+#define _PyArg_NoKeywords(funcname, kwargs) \
+    ((kwargs) == NULL || _PyArg_NoKeywords((funcname), (kwargs)))
+#define _PyArg_NoStackKeywords(funcname, kwnames) \
+    ((kwnames) == NULL || _PyArg_NoStackKeywords((funcname), (kwnames)))
+#define _PyArg_NoPositional(funcname, args) \
+    ((args) == NULL || _PyArg_NoPositional((funcname), (args)))
 #endif
 PyObject * Py_VaBuildValue(const char *, va_list);
 
--- a/third_party/python/Include/modsupport.h
+++ b/third_party/python/Include/modsupport.h
@@ -66,12 +66,15 @@ typedef struct _PyArg_Parser {
 } _PyArg_Parser;
 #ifdef PY_SSIZE_T_CLEAN
 #define _PyArg_ParseTupleAndKeywordsFast  _PyArg_ParseTupleAndKeywordsFast_SizeT
-#define _PyArg_ParseStack  _PyArg_ParseStack_SizeT
+#define _PyArg_ParseStack _PyArg_ParseStack_SizeT
+#define _PyArg_ParseStackAndKeywords  _PyArg_ParseStackAndKeywords_SizeT
 #define _PyArg_VaParseTupleAndKeywordsFast  _PyArg_VaParseTupleAndKeywordsFast_SizeT
 #endif
 int _PyArg_ParseTupleAndKeywordsFast(PyObject *, PyObject *,
                                                  struct _PyArg_Parser *, ...);
-int _PyArg_ParseStack(PyObject **args, Py_ssize_t nargs, PyObject *kwnames,
+int _PyArg_ParseStack(PyObject **args, Py_ssize_t nargs,
+                    const char *format, ...);
+int _PyArg_ParseStackAndKeywords(PyObject **args, Py_ssize_t nargs, PyObject *kwnames,
                                   struct _PyArg_Parser *, ...);
 int _PyArg_VaParseTupleAndKeywordsFast(PyObject *, PyObject *,
                                                    struct _PyArg_Parser *, va_list);
--- a/third_party/python/Include/modsupport.h
+++ b/third_party/python/Include/modsupport.h
@@ -47,6 +47,13 @@ PyObject * Py_BuildValue(const char *, ...);
 PyObject * _Py_BuildValue_SizeT(const char *, ...);
 
 #ifndef Py_LIMITED_API
+int _PyArg_UnpackStack(
+    PyObject **args,
+    Py_ssize_t nargs,
+    const char *name,
+    Py_ssize_t min,
+    Py_ssize_t max,
+    ...);
 int _PyArg_NoKeywords(const char *funcname, PyObject *kwargs);
 int _PyArg_NoStackKeywords(const char *funcname, PyObject *kwnames);
 int _PyArg_NoPositional(const char *funcname, PyObject *args);
--- a/third_party/python/Lib/pickle.py
+++ b/third_party/python/Lib/pickle.py
@@ -1465,12 +1465,16 @@ class _Unpickler:
     def load_appends(self):
         items = self.pop_mark()
         list_obj = self.stack[-1]
-        if isinstance(list_obj, list):
-            list_obj.extend(items)
+        try:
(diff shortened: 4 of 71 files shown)


ahgamut:
Using `METH_FASTCALL` instead of `METH_VARARGS` or `METH_VARARGS | METH_KEYWORDS` is a simple way to improve function call performance in Python 3.7. There are some commits to Python3.7 which just change existing function signatures to use `METH_FASTCALL`.

Examples of ports that are yet to be added to the repo: python/cpython@b813a0e9 python/cpython@c9ea9335 python/cpython@41baebd8

There are still some small improvements to be gained.

ahgamut:
A couple of interesting observations: 

* According to my measurements, `python.com` under `MODE=optlinux` now starts up slightly faster than Anaconda Python 3.6. This is interesting because Conda Python binaries use LTO+PGO, and somehow the APE starts up faster. Should probably measure again on a separate system to be sure. LTO+PGO Python wins out overall though.

| Benchmark               | pyconda36-13 | APE-optlinux-fastcall  |
|-------------------------|:------------:|:----------------------:|
| Geometric mean          | (ref)        | 1.04x slower           |
| pickle_dict             | 54.7 us      | 39.4 us: 1.39x faster  |
| sqlite_synth            | 4.66 us      | 3.37 us: 1.38x faster  |
| telco                   | 13.9 ms      | 10.9 ms: 1.27x faster  |
| pickle_list             | 7.61 us      | 6.20 us: 1.23x faster  |
| unpickle_list           | 8.38 us      | 7.09 us: 1.18x faster  |
| pickle                  | 19.7 us      | 17.9 us: 1.10x faster  |
| pidigits                | 313 ms       | 292 ms: 1.07x faster   |
| python_startup          | 16.8 ms      | 15.9 ms: 1.06x faster  |
| xml_etree_parse         | 274 ms       | 262 ms: 1.05x faster   |
| xml_etree_iterparse     | 207 ms       | 204 ms: 1.01x faster   |
| scimark_fft             | 638 ms       | 632 ms: 1.01x faster   |

* The build+test system for APE Python made it so much easier to do these backports! Debugging was straightforward, the wait times for testing were far less bothersome due to running in parallel.

ahgamut:
@jart It ought to be worth it to change frequently called `METH_VARARGS` functions to `METH_FASTCALL`; I did it for a few functions in the [above commit](https://github.com/jart/cosmopolitan/pull/317/commits/ed44f3b14dc0e295789d07fc224130d77dd1442e). The same can be done for a lot of functions in `Modules/*.c`. There's probably a more automatic way to do it (like an editor macro). Python provides the [Argument Clinic tool](https://docs.python.org/3.7/howto/clinic.html) to generate `.inc` files from docstrings, but the editor macro approach seems easier.

`METH_FASTCALL` is faster than `METH_VARARGS` because:

* it avoids allocating some temporary tuple objects
* it uses `static _PyArg_Parser` objects instead of manually parsing (the static objects get deallocated during `_PyArg_Fini`)
* it skips a few function calls


`METH_FASTCALL` is not fully backported yet, this current PR does not have the functions necessary for `METH_FASTCALL | METH_KEYWORDS` (python/cpython@6969eaf4682beb), which looks like it streamlines things further.

jart:
> According to my measurements, python.com under MODE=optlinux now starts up slightly faster than Anaconda Python 3.6. This is interesting because Conda Python binaries use LTO+PGO, and somehow the APE starts up faster

Nice result! Also think about the fact that we're doing all the heavy lifting of decompressing the sources we load. So even with the added complexity of running zlib deflate() on each file, we're still outperforming the alternatives. It really goes to show how much potential our model of execution has versus the traditional way of doing things.

> I'm still wary about compatibility (dictionary lookup has changed internally from 3.6 to 3.7, got segfaults when I attempted to port python/cpython@7dc6a5f).

Sometimes the dependency chain of an upstream commit isn't clear, but I think it's fair to assume if the tests pass we've should be clear, since having the tests ported buys us the luxury of knowing that for certain. Even though porting the tests and making them go fast wasn't strictly necessary, I like that it brought us the freedom to make these kinds of changes.

jart (approved):

ahgamut:
`MODE=opt` uses `-march=native`, and I've run these comparisons on a Haswell CPU, so maybe a more modern CPU could show better results? There is still LTO + PGO/`UNLIKELY` to test at the compiler level.

I remember a blog post somewhere that compared performance and startup times for Python 2.7, 3.6, and 3.7.
3.6 was by far the slowest, and 3.7 was only slightly slower than 2.7 despite doing a lot more work (like the unicode loading at start). 

It would be interesting to port Python3.7 just to see how much the design of Python affects overall speed. Perhaps the simpler design of CPython3.6 (slower parsing, fewer type-specific shortcuts, more internal function calls, fewer VM/bytecode optimizations, old code) masks some performance improvements of APE Python. The `METH_FASTCALL` set of backports target one category for speedup (internal function calls). A similar set of commits should exist for other categories; [This](https://faster-cpython.readthedocs.io/optimizations.html) is a useful reference.

Also Python is now changing the C API to aid optimization, which breaks compatibility and some C extensions (see [here](https://vstinner.github.io/c-api-abstract-pyobject.html) and [here](https://vstinner.github.io/c-api-python3_10-changes.html)).

jart:
The METH_FASTCALL stuff sounds promising. But I want to start switching gears towards getting it published. For example, we need an HTML website of some kind that looks good, and it should include the auto-generated Python documentation, since some of our modules are nonstandard and it'll make life easier for people if they can get help on the website. Another thing we might consider is naming the project Basilisk since it's even more fearsome than Anaconda. We could keep going with Actually Portable Python, since after all we're likely to also have an Actually Portable Lua and and Actually Portable Scheme too. What do you think?

ahgamut:
> Another thing we might consider is naming the project Basilisk since it's even more fearsome than Anaconda. We could keep going with Actually Portable Python, since after all we're likely to also have an Actually Portable Lua and and Actually Portable Scheme too. What do you think?

Basilisk Python sounds awesome, but to me the name doesn't explain the magic as well as bestline or sectorlisp.
I don't have a specific name in mind though, so I'm okay with any name. I just think of it as Actually Portable Python because that's what I called it in the blog post when I ported Lua.

The [Python documentation](https://docs.python.org/3/faq/general.html) says:

>  Finally, Python is portable: it runs on many Unix variants including Linux and macOS, and on Windows.

Now, Python isn't actually portable, especially once we write extensions in C or C++.
Actually Portable Python is: with the Makefile build system we can build/test/debug extensions on Linux and run anywhere. If I add OpenSSL 1.1.1k, `numpy`, `greenlet` etc. to Python, test it across all `MODE`s, the code can run on 7 operating systems. And that's marvelous.

jart:
I'm so glad you brought it up. The Python team has always worked very hard to give Python excellent source portability. The binary portability we've added is a big improvement, but that source portability they already achieved is worth respecting. There's a much different story behind the name "Actually Portable Executable" which pokes fun at PE for not living up to its name, but that's more likely to flatter Microsoft if anything, since it reminds everyone of how they won their contest with IBM OS/2. If you know someone on the Python dev team, then it'd be simpler if we just asked them.

ahgamut:
> But I want to start switching gears towards getting it published.

* If possible I'd prefer to finish the `FASTCALL` backport; there is one more block of commits I need to examine before backporting is complete. 
* Any third-party Python modules to add? pip, setuptools, bottle etc.
* A big question many of us will have with Actually Portable Python is "but what about C extensions?" It would be nice to have a PR or two (don't need to merge) that port well-known CPython extensions into Cosmopolitan, as a proof of concept. If thats too much we can always show SQLite as an example, or point to https://github.com/ahgamut/cpython/tree/cosmo_py36.
* I'll check which tests can be added to the build.

> For example, we need an HTML website of some kind that looks good, and it should include the auto-generated Python documentation, since some of our modules are nonstandard and it'll make life easier for people if they can get help on the website.

I looked at the dependencies for [Sphinx](https://github.com/sphinx-doc/sphinx), and it looks like `markupsafe` is the only one with a C extension. Let me see if Sphinx runs on Actually Portable Python.

jart:
Waiting on backports is fine. It'll give me more time to automate the remote test runner for Python tests.

Package management is totally up to you. My primary interest is making sure everything defined by the monorepo works well. 

The GUI work @jacereda is doing with dlopen() would be synergistic with Python extensions. I'd still want to integrate chibicc since there's a big value prop to having a solution to building C modules, at all, in a way that behaves exactly the same across platforms. Since half as fast as GCC/MSVC is still 10x faster than Python. But if you don't mind bloating the VA space a little bit schlepping in platform  specific libraries, and trading away some portability convenience, then it's a great story to say you've got dlopen() as an option too. It'd be nice to have at least one for launch.

We have enough SSL support that people can use the urllib http client. So long as we're transparent about the things we're still working towards, I'm sure people will forgive an alpha for not having perfectly recreated OpenSSL. I've noticed recently that some maintainers aren't happy right now with OpenSSL 3.0 so there's likely to be more openness to alternatives.

ahgamut:
> Package management is totally up to you. My primary interest is making sure everything defined by the monorepo works well.

The current build system in the monorepo works really well: fork the repo, add necessary files from wheels to either `third_party/python/Lib` or `third_party/python/Modules`, change `python.mk`, build/test, and use built `python.com`.

I expect this ought to be enough for adding packages like `bottle`, `requests`, or even `sphinx`.


* both `pip` and `setuptools` vendor other libraries internally
* `pip install` calls the internet if package is not available locally + setuptools allows arbitrary scripts in `setup.py`, may complicate build process in the repo (testing `pip` requires internet?)
* `pip` downloads require Python's OpenSSL wrapper, and I have not ported OpenSSL to the monorepo. wrapping MbedTLS might be better (even if more difficult).
* Is `OpenExecutable` working everywhere? Installing pure-python wheels is basically unzipping them into the correct location (`/zip/.python/mymodule/`)
* Installing extensions from source will not be as good as the current build system until after chibicc is added to `python.com`

jart:
Oh definitely don't run it as part of the build or test process. It was hard enough getting CJK tests to not talk to the Internet. I'm simply saying I don't care if the pip whl blob is in the python.com file, provided it's community supported and doesn't get in the way. I'm not interested in using it myself. I just don't want to say no to user expectations.

OpenExecutable is a work in progress. Currently focusing on an APE execve() function which is very similar. Once those changes go in you'll have more tools at your disposal if you wish to modify pip so that it installs packages directly into the executable. However you would need to find some way to unblob that whl file Python2.6 has checked-in to their source tree, so changes are reviewable. Can't imagine that would be easy. As for SSL have you tried it since I wrote that barebones Mbedtls HTTPS wrapper?

ahgamut:
> However you would need to find some way to unblob that whl file Python2.6 has checked-in to their source tree, so changes are reviewable. Can't imagine that would be easy. 

Pure-python wheels are just zip files? I don't understand where Python2.6 is coming from.

If you are referring to the `pip` and `setuptools` wheels in the 3.6 source tree, I just unzip them and add the folders into the APE.
`pip` and `setuptools` for Python 3.6 have no binary blobs AFAIK.

PR #320, opened 2021-11-07
Extend GetZipPaths to accept an optional prefix to filter based on

pkulchenko:
This allows pre-filtering returned paths instead of post-filtering them.


jart (approved):
Looks good. Thanks for sending this!

PR #321, opened 2021-11-07
Add redbean function to calculate various HMAC codes

pkulchenko:
@jart, this implements all HMAC and regular digests currently in mbedTLS. I didn't mark any of the existing Md5 and Sha* functions, but they can all be replaced with proper `GetCryptoHash` calls. This implementation also include `BLAKE2B256`, which is not present as one of the standalone function.

Let me know if you want to mark them all as "Deprecated; use GetCryptoHash instead".

### DIFF
--- a/tool/net/redbean.c
+++ b/tool/net/redbean.c
@@ -4653,6 +4653,31 @@ static int LuaSha512(lua_State *L) {
   return LuaHasher(L, 64, mbedtls_sha512_ret_512);
 }
 
+static noinline int LuaGetCryptoHash(lua_State *L) {
+  size_t hl, pl, kl;
+  uint8_t d[64];
+  mbedtls_md_context_t ctx;
+  // get hash name, payload, and key
+  void *h = luaL_checklstring(L, 1, &hl);
+  void *p = luaL_checklstring(L, 2, &pl);
+  void *k = luaL_optlstring(L, 3, "", &kl);
+
+  const mbedtls_md_info_t *digest = mbedtls_md_info_from_string(h);
+  if (!digest) return luaL_argerror(L, 1, "unknown hash type");
+
+  if (kl == 0) {
+    // no key provided, run generic hash function
+    if ((digest->f_md)(p, pl, d))
+      return luaL_error(L, "bad input data");
+  }
+  else if (mbedtls_md_hmac(digest, k, kl, p, pl, d))
+    return luaL_error(L, "bad input data");
+
+  lua_pushlstring(L, (void *)d, digest->size);
+  mbedtls_platform_zeroize(d, sizeof(d));
+  return 1;
+}
+
 static int LuaGetRandomBytes(lua_State *L) {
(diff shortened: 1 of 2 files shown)


pkulchenko:
@jart, just a quick note on `return luaL_error(L, "bad input data");`. I realize that you use `luaL_error(L, "bad input data"); unreachable;` instead, but `return luaL_error...` is [idiomatic Lua code described in the Lua manual](https://www.lua.org/manual/5.4/manual.html#luaL_error). I can also update the existing code at the next opportunity if you're on board with that.

jart:
Could you resolve the conflict from the previous merge?

I fully support us doing things the idiomatic way. The upstream Lua sources needed to conform to a much stricter interpretation of the ANSI standard than is needed for us in redbean. I tried at one point to bolt `wontreturn` onto the Lua APIs which would have been the most elegant solution in terms of code generation, but it's sadly incompatible with the `return` idiom which is baked in pretty deep. So it's better that we just do things the Lua way.

tool/net/redbean.c
```diff
@@ -4638,6 +4638,31 @@ static int LuaSha512(lua_State *L) {
...
+
+  const mbedtls_md_info_t *digest = mbedtls_md_info_from_string(h);
+  if (!digest) return luaL_argerror(L, 1, "unknown hash type");
+
+  if (kl == 0) {
+    // no key provided, run generic hash function
+    if ((digest->f_md)(p, pl, d))
+      return luaL_error(L, "bad input data");
+  }
+  else if (mbedtls_md_hmac(digest, k, kl, p, pl, d))
```
jart, line 4658: Oh nice, I forgot that there was a terser API than what I recommended earlier offline.

tool/net/redbean.c
```diff
@@ -4638,6 +4638,31 @@ static int LuaSha512(lua_State *L) {
...
+static noinline int LuaGetCryptoHash(lua_State *L) {
+  size_t hl, pl, kl;
+  uint8_t d[64];
+  mbedtls_md_context_t ctx;
+  // get hash name, payload, and key
+  void *h = luaL_checklstring(L, 1, &hl);
+  void *p = luaL_checklstring(L, 2, &pl);
+  void *k = luaL_optlstring(L, 3, "", &kl);
+
+  const mbedtls_md_info_t *digest = mbedtls_md_info_from_string(h);
```
jart, line 4650: It's interesting how this worked even with the constness in the type signature. Strictly speaking you'd normally want to call md_init() and then md_setup() where you index into the result of md_info_from_string(). But looking at the code the way you did things should be fine, and would have been caught anyway by the `.rodata` protection had it been an issue.

jart (approved):

jart:
Thanks!

pkulchenko:
Rebased; thank you for the feedback!

jart:
Nice, six second jinx.

PR #322, opened 2021-11-13
Add encode json and encode Lua functions to redbean

pkulchenko:
These functions only do encoding. Lua decoding can be done using available methods (load/loadstring) and JSON decoding needs to be added. I've benchmarked JSON processing against lua-cjson and this implementation beats it handily on array encoding with numbers, is a bit slower on arrays with strings, and about 3x slower on objects with strings. I think this is caused by `EscapeJsStringLiteral` and memory management in calls on 1-2 char strings to append* functions. cjson is using strbuf implementation to manage dynamically allocated string buffers (somewhat similar to what append* functions do). We may come up with a subset of what `EscapeJsStringLiteral` does, as some of that escaping is optional for JSON encoding (or maybe make it an option).

I tried to improve the performance a bit by pre-allocating a buffer with `appendr` and then working directly with that buffer, but ran into issues with resizing the buffer to the actual size after the processing (as it didn't resize on some occasions, possibly because `malloc_usable_size` was picking the "wrong" (larger) memory fragment that was already resized to a smaller one, not sure).

I think it's already fairly complete and fast, but would appreciate a review and any suggestions to improve. I'll report any issues I find, as I plan to use both of these methods shortly.

### DIFF
--- a/tool/net/redbean.c
+++ b/tool/net/redbean.c
@@ -4704,9 +4704,10 @@ static int EncodeJsonData(lua_State *L, char **buf, int level, char *numformat) 
   int t = lua_type(L, idx);
   if (level < 0) return luaL_argerror(L, 1, "too many nested tables");
   if (LUA_TSTRING == t) {
-    appendf(buf, "\"%s\"", gc(EscapeJsStringLiteral(lua_tostring(L, idx), -1, 0)));
+    appendw(buf, '"');
+    appends(buf, gc(EscapeJsStringLiteral(lua_tostring(L, idx), -1, 0)));
+    appendw(buf, '"');
   } else if (LUA_TNUMBER == t) {
-    // TODO: what if int64_t isn't representable as double
     appendf(buf, numformat, lua_tonumber(L, idx));
   } else if (LUA_TBOOLEAN == t) {
     appends(buf, lua_toboolean(L, idx) ? "true" : "false");
@@ -4716,10 +4717,10 @@ static int EncodeJsonData(lua_State *L, char **buf, int level, char *numformat) 
     isarray = tbllen > 0 || // integer keys present
       (lua_pushnil(L), lua_next(L, -2) == 0) || // no non-integer keys
       (lua_pop(L, 2), false); // pop key/value pushed by lua_next
-    appendd(buf, isarray ? "[" : "{", 1);
+    appendw(buf, isarray ? '[' : '{');
     if (isarray) {
       for (int i = 1; i <= tbllen; i++) {
-        if (i > 1) appendd(buf, ",", 1);
+        if (i > 1) appendw(buf, ',');
         lua_rawgeti(L, -1, i); // table/-2, value/-1
         EncodeJsonData(L, buf, level-1, numformat);
         lua_pop(L, 1);
@@ -4730,47 +4731,45 @@ static int EncodeJsonData(lua_State *L, char **buf, int level, char *numformat) 
       while (lua_next(L, -2) != 0) {
         if (!lua_isstring(L, -2))
           return luaL_argerror(L, 1, "expected number or string as key value");
-        if (i++ > 1) appendd(buf, ",", 1);
+        if (i++ > 1) appendw(buf, ',');
+        appendw(buf, '"');
         if (lua_type(L, -2) == LUA_TSTRING) {
-          appendf(buf, "\"%s\":",
-            gc(EscapeJsStringLiteral(lua_tostring(L, -2), -1, 0)));
+          appends(buf, gc(EscapeJsStringLiteral(lua_tostring(L, -2), -1, 0)));
         } else {
           // we'd still prefer to use lua_tostring on a numeric index, but can't
           // use it in-place, as it breaks lua_next (changes numeric key to a string)
           lua_pushvalue(L, -2); // table/-4, key/-3, value/-2, key/-1
-          appendf(buf, "\"%s\":", lua_tostring(L, idx)); // use the copy
+          appends(buf, lua_tostring(L, idx)); // use the copy
           lua_remove(L, -1); // remove copied key: table/-3, key/-2, value/-1
         }
+        appendw(buf, '"' | ':' << 010);
         EncodeJsonData(L, buf, level-1, numformat);
         lua_pop(L, 1); // table/-2, key/-1
       }
       // stack: table/-1, as the key was popped by lua_next
     }
-    appendd(buf, isarray ? "]" : "}", 1);
-  } else if (LUA_TNIL == t)
+    appendw(buf, isarray ? ']' : '}');
+  } else if (LUA_TNIL == t) {
     appendd(buf, "null", 4);
-  else {
+  } else {
     return luaL_argerror(L, 1, "can't serialize value of this type");
   }
   return 0;
 }
 
 static void EscapeLuaString(char *s, size_t len, char **buf) {
-  static char OCT[8] = "01234567";
-  char escape[4] = "\\000";
-  appendd(buf, "\"", 1);
+  appendw(buf, '"');
   for (size_t i = 0; i < len; i++) {
     if (s[i] == '\\' || s[i] == '\"' || s[i] == '\n' ||
         s[i] == '\0' || s[i] == '\r') {
-      escape[1] = OCT[(s[i] >> 6) & 7];
-      escape[2] = OCT[(s[i] >> 3) & 7];
-      escape[3] = OCT[(s[i] >> 0) & 7];
-      appendd(buf, escape, 4);
+      appendw(buf, '\\' | 'x' << 010 |
+        "0123456789abcdef"[(c & 0xF0) >> 4] << 020 |
+        "0123456789abcdef"[(c & 0x0F) >> 0] << 030);
     } else {
       appendd(buf, s+i, 1);
     }
   }
-  appendd(buf, "\"", 1);
+  appendw(buf, '"');
 }
 
 static int EncodeLuaData(lua_State *L, char **buf, int level, char *numformat) {
@@ -4784,12 +4783,11 @@ static int EncodeLuaData(lua_State *L, char **buf, int level, char *numformat) {
     s = lua_tolstring(L, idx, &slen);
     EscapeLuaString(s, slen, buf);
   } else if (LUA_TNUMBER == t) {
-    // TODO: what if int64_t isn't representable as double
     appendf(buf, numformat, lua_tonumber(L, idx));
   } else if (LUA_TBOOLEAN == t) {
     appends(buf, lua_toboolean(L, idx) ? "true" : "false");
   } else if (LUA_TTABLE == t) {
-    appendd(buf, "{", 1);
+    appendw(buf, '{');
     int i = 0;
     lua_pushnil(L); // push the first key
     while (lua_next(L, -2) != 0) {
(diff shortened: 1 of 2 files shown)


jart:
appendr was intended more for resetting the size of the buffer. I never tried it for the use case of preallocating, which generally isn't an optimization I'd recommend. Functions that don't use append library, such as EscapeJsStringLiteral() usually do it by knowing the largest possible size the buffer could be, in which case it just needs malloc at the beginning and realloc at the end. In terms of its performance, it's more generalized and has smaller code size than the lua-cjson implementation. There are ways we could speed it up for ascii without changing the code much, by using simd.

tool/net/redbean.c
```diff
@@ -4697,6 +4697,162 @@ static int LuaGetHttpReason(lua_State *L) {
...
+static int EncodeJsonData(lua_State *L, char **buf, int level, char *numformat) {
+  size_t idx = -1;
+  size_t tbllen, buflen;
+  bool isarray;
+  int t = lua_type(L, idx);
+  if (level < 0) return luaL_argerror(L, 1, "too many nested tables");
+  if (LUA_TSTRING == t) {
+    appendf(buf, "\"%s\"", gc(EscapeJsStringLiteral(lua_tostring(L, idx), -1, 0)));
+  } else if (LUA_TNUMBER == t) {
+    // TODO: what if int64_t isn't representable as double
```
jart, line 4709: Two possible options. One is make it NAN. The other is raise an error. Saturation clamping we could do but it wouldn't feel right in the general case. It'd help to know what other json impls do.

tool/net/redbean.c
```diff
@@ -4697,6 +4697,162 @@ static int LuaGetHttpReason(lua_State *L) {
...
 }
 
+static int EncodeJsonData(lua_State *L, char **buf, int level, char *numformat) {
+  size_t idx = -1;
+  size_t tbllen, buflen;
+  bool isarray;
+  int t = lua_type(L, idx);
+  if (level < 0) return luaL_argerror(L, 1, "too many nested tables");
+  if (LUA_TSTRING == t) {
+    appendf(buf, "\"%s\"", gc(EscapeJsStringLiteral(lua_tostring(L, idx), -1, 0)));
```
jart, line 4707: Optionally, if string performance is a concern, you could nudge out a bit more of it here by breaking it down into `appendw('"');appendd(...);appendw('"');` That way you avoid spinning up the `__fmt()` engine and avoid the strlen scan.

tool/net/redbean.c
```diff
@@ -4697,6 +4697,162 @@ static int LuaGetHttpReason(lua_State *L) {
...
+  } else if (LUA_TNUMBER == t) {
+    // TODO: what if int64_t isn't representable as double
+    appendf(buf, numformat, lua_tonumber(L, idx));
+  } else if (LUA_TBOOLEAN == t) {
+    appends(buf, lua_toboolean(L, idx) ? "true" : "false");
+  } else if (LUA_TTABLE == t) {
+    tbllen = lua_rawlen(L, idx);
+    // encode tables with numeric indices and empty tables as arrays
+    isarray = tbllen > 0 || // integer keys present
+      (lua_pushnil(L), lua_next(L, -2) == 0) || // no non-integer keys
```
jart, line 4717: Lua will clean up the stack on function return if this conditional expression short-circuits, correct?

tool/net/redbean.c
```diff
@@ -4697,6 +4697,162 @@ static int LuaGetHttpReason(lua_State *L) {
...
+      (lua_pop(L, 2), false); // pop key/value pushed by lua_next
+    appendd(buf, isarray ? "[" : "{", 1);
```
jart, line 4719: `appendw(buf, isarray ? '[' : '{');`

tool/net/redbean.c
```diff
@@ -4697,6 +4697,162 @@ static int LuaGetHttpReason(lua_State *L) {
...
+          appendf(buf, "\"%s\":", lua_tostring(L, idx)); // use the copy
+          lua_remove(L, -1); // remove copied key: table/-3, key/-2, value/-1
+        }
+        EncodeJsonData(L, buf, level-1, numformat);
+        lua_pop(L, 1); // table/-2, key/-1
+      }
+      // stack: table/-1, as the key was popped by lua_next
+    }
+    appendd(buf, isarray ? "]" : "}", 1);
+  } else if (LUA_TNIL == t)
```
jart, line 4750: [nit] could you add curlies here?

tool/net/redbean.c
```diff
@@ -4697,6 +4697,162 @@ static int LuaGetHttpReason(lua_State *L) {
...
+}
+
+static void EscapeLuaString(char *s, size_t len, char **buf) {
+  static char OCT[8] = "01234567";
+  char escape[4] = "\\000";
+  appendd(buf, "\"", 1);
+  for (size_t i = 0; i < len; i++) {
+    if (s[i] == '\\' || s[i] == '\"' || s[i] == '\n' ||
+        s[i] == '\0' || s[i] == '\r') {
+      escape[1] = OCT[(s[i] >> 6) & 7];
```
jart, line 4765: Aren't Lua strings decimal? In situations like this, you could also use appendw if you want. For example, 

```c
appendw(&b, '\\' | 
    (((s[i] & 0300) >> 6) + '0') << 010 |
    (((s[i] & 0070) >> 3) + '0') << 020 |
    (((s[i] & 0007) >> 0) + '0') << 030);
```

Obviously change if the intent is decimal. String indexing technique is mostly useful if intent is hex. It's fine to duplicate the string literal since the linker de-dupes it.

tool/net/redbean.c
```diff
@@ -4697,6 +4697,162 @@ static int LuaGetHttpReason(lua_State *L) {
...
+    lua_getfield(L, 2, "useoutput");
+    // ignore useoutput outside of request handling
+    if (ishandlingrequest && lua_isboolean(L, -1)) useoutput = lua_toboolean(L, -1);
+    lua_getfield(L, 2, "maxdepth");
+    maxdepth = luaL_optinteger(L, -1, maxdepth);
+    lua_getfield(L, 2, "numformat");
+    numformat = luaL_optstring(L, -1, numformat);
+  }
+  lua_settop(L, 1); // keep the passed argument on top
+  (void)Encoder(L, useoutput ? &outbuf : &p, maxdepth, numformat);
```
jart, line 4837: Isn't the `(void)` statement thing only really necessary if we're compiling with stricter compiler warnings? Within the monorepo .c files we officially speaking only need to cater to one vendored compiler with one defined configuration. Code with an ultra-source-portable audience is probably better regularly exported to a separate repo, e.g. bestline.

tool/net/redbean.c
```diff
@@ -4697,6 +4697,162 @@ static int LuaGetHttpReason(lua_State *L) {
...
+
+  if (useoutput) {
+    lua_pushnil(L);
+  } else {
+    lua_pushstring(L, p);
+    free(p);
```
jart, line 4843: Shouldn't this `free(p)` be moved to the next line?

jart (changes requested):

tool/net/redbean.c
pkulchenko: NAN is not acceptable in (standard) JSON, although there are some JSON implementation that may use it. Raising an error should be fine. What's a good/fast way to check for that condition?

tool/net/redbean.c
pkulchenko: Yes, I'll switch to that.

tool/net/redbean.c
pkulchenko: Right; if the second line returns true, then there is nothing left on the stack and if it continues to the third line, then the stack is corrected with lua_pop there.

tool/net/redbean.c
pkulchenko: Updated.

tool/net/redbean.c
pkulchenko: Updated.

tool/net/redbean.c
pkulchenko: Updated. This is almost exactly what I tried initially, but I couldn't make it to produce a correct result for some reason (I even tried the `READ64LE` form). I'll report back if I notice any issues after I run my tests.

tool/net/redbean.c
pkulchenko: Removed.

tool/net/redbean.c
pkulchenko: No, because the former branch is when the output buffer is used, so it doesn't need to be freed.

tool/net/redbean.c
pkulchenko: @jart, should I simply compare the integer with `0x7FEFFFFFFFFFFFFF` per this SO answer: https://stackoverflow.com/a/39783981/1442917?

pkulchenko:
Everything has been addressed, pushed, and tested except the integer comparison.

tool/net/redbean.c
jart: The right thing to do is probably to do nothing. Quoth ANSI X3.159-1988 § 3.2.1.4

```
   When a double is demoted to float or a long double to double or
float, if the value being converted is outside the range of values
that can be represented, the behavior is undefined. 
```

If I write a program like this:

```c
  volatile int64_t x = INT64_MAX;
  printf("%ld\n%.19g\n", INT64_MAX, (double)x);
  return 0;
}
```

It appears to output an approximation:

```
9223372036854775807
9223372036854775808
```

jart: Oh whoops wrong comment.

tool/net/redbean.c
pkulchenko: Got it; based on your other comment I removed the TODO comments to keep things clear.

jart: Honestly I think for a general purpose abstraction like JSON the right thing to do is probably to do nothing. See comment I left in the other thread. Maybe leave a comment saying `/* precision loss possible */` in case we ever find reason to make the behavior configurable. Please note undefined behavior has the following definition per ANSI C standard:

```
 * Undefined behavior --- behavior, upon use of a nonportable or
   erroneous program construct, of erroneous data, or of
   indeterminately-valued objects, for which the Standard imposes no
   requirements.  Permissible undefined behavior ranges from ignoring the
   situation completely with unpredictable results, to behaving during
   translation or program execution in a documented manner characteristic
   of the environment (with or without the issuance of a diagnostic
   message), to terminating a translation or execution (with the issuance
   of a diagnostic message).
```

So it would be legal to raise an error if you want. But whatever you choose to do, it's probably worth noting in the docs.

tool/net/redbean.c
jart: Oh I see. I didn't notice the ternary. It'd still work if the free was in the same logical scope, since p is initialized to 0.

tool/net/redbean.c
```diff
@@ -4697,6 +4697,161 @@ static int LuaGetHttpReason(lua_State *L) {
...
 
+static int EncodeJsonData(lua_State *L, char **buf, int level, char *numformat) {
+  size_t idx = -1;
+  size_t tbllen, buflen;
+  bool isarray;
+  int t = lua_type(L, idx);
+  if (level < 0) return luaL_argerror(L, 1, "too many nested tables");
+  if (LUA_TSTRING == t) {
+    appendw(buf, '"');
+    appends(buf, gc(EscapeJsStringLiteral(lua_tostring(L, idx), -1, 0)));
```
jart, line 4708: [optional] if you want performance that's good to the last drop here:

```c
    tmpstr = EscapeJsStringLiteral(lua_tostring(L, idx), lua_string_len, &tmpstrlen);
    appendd(buf, tmpstr, tmpstrlen);
    free(tmpstr);
```

It's obviously less elegant than the thing you did before. I only make the suggestion since performance was mentioned. That's the best possible way you can get it without optimizing the underlying APIs.

tool/net/redbean.c
```diff
@@ -4697,6 +4697,161 @@ static int LuaGetHttpReason(lua_State *L) {
...
+  }
+  return 0;
+}
+
+static void EscapeLuaString(char *s, size_t len, char **buf) {
+  appendw(buf, '"');
+  for (size_t i = 0; i < len; i++) {
+    if (s[i] == '\\' || s[i] == '\"' || s[i] == '\n' ||
+        s[i] == '\0' || s[i] == '\r') {
+      appendw(buf, '\\' |
```
jart, line 4765: For what it's worth, if you want to do hex, since I'm pretty sure lua strings aren't octal:

```c
      appendw(buf, '\\' | 'x' << 010 |
          "0123456789abcdef"[(c & 0xF0) >> 4] << 020 |
          "0123456789abcdef"[(c & 0x0F) >> 0] << 030);
```

tool/net/redbean.c
pkulchenko: Right; I think it may be best to keep it the way it is, but it's a good suggestion to keep in mind.

tool/net/redbean.c
pkulchenko: You're right; it shouldn't be octal; fixed.

jart (approved):

PR #324, opened 2021-11-13
Fix incorrect reference in Lua encoding

pkulchenko:
@jart, this is a follow-up to #322, as I marked the comment as resolved before the compilation was done and there was a typo that needs to be fixed. My apologies.


jart (approved):
No worries!

Issue #315, opened 2021-11-04
lua popen returns partial output / fread() doesn't read (everything)


coderofsalvation:
fread() seems to return partial (buffered?) output when indirectly called by lua-code (redbean.c), I tried to reproduce it here:

https://gist.github.com/coderofsalvation/c60d4aa466d1cb137027e743f18bd924

On IRC RhodiumToad mentioned the possible culprit:

```
<RhodiumToad> | it does only one readv() call, it makes no attempt to loop if insufficient data was read from the pipe
<RhodiumToad> | lua interprets a short read result from fread() as meaning end-of-file
<RhodiumToad> | fread() is not permitted by spec to return a short result in the absence of error or EOF
```
I've tried to compare it to the musl-version of [fread.c](https://git.musl-libc.org/cgit/musl/tree/src/stdio/fread.c) hoping that I could somehow shove in some for-loop, but I'm afraid I'm not the right person to do so (modifying fread is risky).

pkulchenko:
I think this is a very interesting point and only @jart can shed some light on the implementation details for fread().

> fread() is not permitted by spec to return a short result in the absence of error or EOF

It looks like Lua aborts read_all (which is what *a is using) as soon as there is a partial return, so its implementation seems to be consistent with this interpretation.

I don't see what in the [fread spec](https://man7.org/linux/man-pages/man3/fread.3.html) may lead to this interpretation though. There is the following text -- "If an error occurs, or the end-of-file is reached, the return value is a short item count (or zero)." -- but I don't think it should be interpreted as if it's the *only* valid case when a short item count should be returned.

The musl implementation you referenced uses `__toread` to check if there is anything available before reading, but cosmo `fread` doesn't do that.

[...]

PR #325, opened 2021-11-14
added example (currently with sideeffects)

coderofsalvation:
```
$ o/third_party/lua/lua.com examples/helloshell.lua | sort | uniq
 124060
 130754
 2924
 2943
 2944
```

more info see https://github.com/jart/cosmopolitan/issues/315

### DIFF
--- a/examples/helloshell.lua
+++ b/examples/helloshell.lua
@@ -1,0 +1,19 @@
+-- o/third_party/lua/lua.com helloshell.lua | sort | uniq
+-- output:
+--   124060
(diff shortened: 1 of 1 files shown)


mrdomino:
Closing as this does not seem intended to be merged; the example stands regardless, although I'm unclear if the bug is still relevant.

pkulchenko:
@mrdomino, this is just a minimal example for the issue discussed in #315, which still appears to be relevant. I agree that the example is sufficient though.

PR #326, opened 2021-11-16
Redbean StoreAsset fix and lua cli

pkulchenko:
@jart, this patch implements three (related) changes:

(removed this one, as it's handled in #360) 1. Moves Fetch SSL check to a later spot and makes the dependency more clear
2. Fixes an issue with StoreAsset when the asset with the same name is already present (the base for the rest of the central dictionary was off, which was creating an invalid archive)
3. Adds -A and -e options to add asset(s) from the comment line and executing Lua commands from the command line. I realize that StoreAsset is not available on Windows and some other systems, but I find it very convenient to have a self-reliant system and expect StoreAsset to become available on Windows as well. The documentation/convenience would be improved as well, as it wouldn't require using system zip and archive renaming.

Feel free to cherry-pick if you disagree with some of these items proposed.

### DIFF
--- a/tool/net/help.txt
+++ b/tool/net/help.txt
@@ -34,6 +34,7 @@ FLAGS
   -b        log message bodies
   -a        log resource usage
   -g        log handler latency
+  -e        run specified Lua command
   -j        enable ssl client verify
   -k        disable ssl fetch verify
   -f        log worker function calls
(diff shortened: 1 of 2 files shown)


pkulchenko:
@jart, I removed SSL check, as it's handled in #360. I think this is good to be merged and I'll handle adding a directory separately (after these changes are in).

jart (approved):
You had me at self-reliant system.

PR #327, opened 2021-11-19
Fix Content-Type being generated twice after setting content type

pkulchenko:
@jart, this fixes a couple of issues with the current content-type header generation:

1. `SetStatus(200); Write("foo")` does not send content-type at all
2. `SetHeader('Content-Type', "some value")` sets two content-type headers (as the first one is being set by `GetLuaResponse`) and the second is set by the SetHeader itself, which is not correct according to the spec

I confirmed that the sequence `SetHeader('Content-Type', "some value"); SetStatus(200); Write("foo")` correctly sets the default header value (as the existing one is reset after `SetStatus()`). There shouldn't be any interaction with all Serve* methods (as long as the content type is set using `AppendContentType`).

I also added a check for the content length being non-zero, as the content type is not needed otherwise.

### DIFF
--- a/tool/net/redbean.c
+++ b/tool/net/redbean.c
@@ -353,6 +353,7 @@ static bool connectionclose;
 static bool hasonworkerstop;
 static bool hasonworkerstart;
 static bool hasonhttprequest;
+static bool hascontenttype;
 static bool ishandlingrequest;
 static bool keyboardinterrupt;
 static bool listeningonport443;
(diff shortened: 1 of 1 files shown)


pkulchenko:
This still doesn't prevent someone to call `SetHeader("Content-Type", ...)` multiple times.

jart (approved):
Looks good. Thanks!

PR #328, opened 2021-11-19
Finish backporting METH_FASTCALL

ahgamut:
I think all relevant commits have been backported. the remaining easy ones are mostly adding `METH_FASTCALL` to built-in modules and functions.

I also back-ported two opcodes from Py37 `LOAD_METHOD` and `CALL_METHOD`.
(I thought the hard part would be running `freeze` and re-building `importlib.inc`, but the build system handled that.)
Now the bytecode magic number for APE Python is `3390` (Python 3.7a1), from `3379`.

All existing tests pass in `MODE=`, `MODE=dbg`, `MODE=tiny`, `MODE=opt`, and `MODE=optlinux`.
I'll run the benchmark and confirm that these commits do help performance (the tests seem to complete faster), but otherwise I think the backporting should be enough for now.


ahgamut:
I tried moving `abc.py` to the C implementation. It compiles, and all tests pass, except 4-6 that use `typing.py`. 

`typing.py` is quite difficult to change. Specifically, adding `_abc.c` from Py37 will eventually lead to errors with `test_typing`, such as `isinstance(b'', typing.BytesString)` evaluating to `False` instead of `True`. I'll put the changes up on a separate branch if anyone has an idea to fix this.

ahgamut:
@jart the following tests fail in `MODE=tiny`, so they are commented out in the `python.mk`: 

```make
# The below tests fail in MODE=tiny
# third_party/python/Lib/test/test_trace.py				\
 	third_party/python/Lib/test/test_signal.py			\
 	third_party/python/Lib/test/test_property.py				\
 	third_party/python/Lib/test/test_logging.py				\
 	third_party/python/Lib/test/test_pydoc.py				\
	third_party/python/Lib/test/test_doctest2.py				\
	third_party/python/Lib/test/test_dynamicclassattribute.py		\

```

these tests pass in `MODE=`, `MODE=dbg`, `MODE=opt`, and `MODE=optlinux`, so the failures are likely related to docstrings being missing or the `.py` files not added to the ZIP store.

Is it possible to have the `.py` files added to `pythontester.com` for `MODE=tiny`?

ahgamut:
I think the change to OrderedDict in #317 needs to be reverted, because all tests are crashing. I'll check it out later.

jart:
Could you sync this to head so I can review and merge?

ahgamut:
@jart the tests in `python.mk` pass in `MODE=`, `MODE=tiny`,  `MODE=dbg`, `MODE=opt`, `MODE=optlinux`, and `MODE=asan`. (A few tests failed in `MODE=asan` but worked when I ran the whole thing again.)

There are some tests which fail only in `MODE=tiny`; I've commented them out for now, I'll add a `unittest.skipIf` guard to those files. I've also commented out `test_cmdline_script` and `test_doctest`, which are currently failing for some reason.

The backports are a lot of new third-party code; if they play well with the existing stuff, we can update other CPython methods to also use the `METH_FASTCALL` format. 

Something fun to do would be to build APE Python from each commit since `MODE=optlinux` was introduced, and see how performance has changed across all these commits.

ahgamut:
Apart from making more builtin-methods use `METH_FASTCALL` I don't think there are many other avenues for speedups. There are some compiler-level tricks possible, like [nudging gcc to produce better code in `ceval.c`](https://www.corsix.org/content/micro-optimisations-can-speed-up-cpython) which I tried [here](https://github.com/ahgamut/cosmopolitan/tree/ceval-mod), but they don't seem to improve overall performance significantly. Perhaps testing on an empty system might show the difference.

CPython's heavy indirection offsets the improvements offered by Cosmopolitan Libc, and barring some magic or heavy modifications to the internals, I don't expect any more large speedups from the Python side. Part of the slowness lies in the fact that Python only started focusing on speed from Py37 onwards, which is why I attempted these backports.

If major speed improvements are desired, we can replace APE Python 3.6 in the repo with [Python 3.9](https://github.com/ahgamut/cpython/tree/cosmo_py39) (or higher!). I think such a change would be better once threads are available, because right now APE Python 3.9 spoofs `_dummy_thread` at the C level, which is a trick I would prefer to avoid.

jart (approved):

jart:
https://github.com/ahgamut/cosmopolitan/commit/d4be3083408ec405811f0ea61b01c109ed299247 looks like a very solid compiler optimization that would avoid a great many superfluous `movslq` instructions being generated, which I've noticed before are very significant from a microoptimization standpoint. See also https://github.com/jart/cosmopolitan/tree/master/tool/plinko I would totally merge that change.

PR #336, opened 2021-12-07
Fix asset read permissions (closes #335)

### LINKED ISSUE #335: asset should be readable by user or group, not just others
https://github.com/jart/cosmopolitan/blob/d6a039821f927cc81b71e11f21599ff93e1fed62/tool/net/redbean.c#L6203

I would have put `if ((m = GetMode(a)) & 0444) {` instead.

It shooted me in the foot as I had a file in `rw-------` and got `403 Forbidden`.

pkulchenko:


jart (changes requested):
As mentioned in #335 I recommend we have better logs and documentation to help people navigate this gotcha.

pkulchenko:
@jart, I updated the documentation and added a comment; the log message was already present, so I haven't changed that.

jart (approved):
Looks good. Thanks!

PR #338, opened 2021-12-08
Add optional network mask to redbean FormatIp

pkulchenko:


pkulchenko:
Closing this per discussion with @jart, as this can be achieved using bitops. For earlier Lua versions that don't have them, an alternative will be explored when they become available with redbean.

PR #340, opened 2021-12-28
Add antirez picol

octetta:
As a prelude to possibly adding Jim Tcl, here's a teensy Tcl clone from antirez's thought experiment ("...a Tcl interpreter in 550 lines of C code ") http://oldblog.antirez.com/page/picol.html

### DIFF
--- a/examples/examples.mk
+++ b/examples/examples.mk
@@ -140,6 +140,18 @@ o/$(MODE)/examples/hello.com.dbg:				\
 		$(APE_NO_MODIFY_SELF)
 	@$(APELINK)
 
+o/$(MODE)/examples/picol.o:					\
+		OVERRIDE_CPPFLAGS +=				\
+			-DSTACK_FRAME_UNLIMITED
+
+o/$(MODE)/examples/picol.com.dbg:				\
+		$(EXAMPLES_DEPS)				\
+		o/$(MODE)/examples/picol.o			\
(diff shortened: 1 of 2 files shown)


examples/examples.mk
```diff
@@ -140,6 +140,18 @@ o/$(MODE)/examples/hello.com.dbg:				\
 		$(APE_NO_MODIFY_SELF)
 	@$(APELINK)
 
+o/$(MODE)/examples/picol.o:					\
```
octetta, line 143: I added this rather hastily... is there a logical place this should have been placed (alpha order, etc?).

examples/picol.c
```diff
@@ -0,0 +1,649 @@
...
+ */
+
+/*
+ * Original from http://oldblog.antirez.com/page/picol.html
+ * Changes on 2021-12-27 by octetta :
+ * . Use Cosmopolitan's headers.
+ * . Formatted as per Cosmopolitan's standards.
+ */
+
+#include <libc/log/log.h>
```
octetta, line 35: I just nuked antirez' original includes in favor of the ones that I thought covered the library use here.

jart (approved):
I can't believe how simple this interpreter is. Let me know when it's synced.

examples/examples.mk
octetta: I sent the copyright assignment.

Issue #180, opened 2021-06-02
Does Cosmopolitan work on Windows with Clang?


Keithcat1:
I downloaded cosmopolitan.zip and tried to create me a portable executable using Clang. I thought that it should work since Clang can generate code for any CPU as long as you have the correct headers and libs ETC.
C:\py\cosmopolitan>clang -g -O -o hello.com.dbg hello.c                      -static -fno-pie -no-pie -mno-red-zone -fno-omit-frame-pointer -nostdlib -nostdinc -Wl,--gc-sections -Wl,-z,max-page-size=0x1000       -Wl,-T,ape.lds -include cosm
opolitan.h crt.o ape.o cosmopolitan.a --target=x86_64-pc-linux-gnu
ld.lld: error: cosmopolitan.a(zip.o):(.zip.5+0x10): relocation R_X86_64_32 out o
f range: 18446744073705357312 is not in [0, 4294967295]
clang-12: error: linker command failed with exit code 1 (use -v to see invocation)
It would be nice if this was supported, because just about everyone has clang on their computer.

jart:
Could you try the 0.2 release? A regression slipped into the project at some point since then regarding LLVM support. We'll likely need to git bisect the repo to figure out exactly what caused it.

I would recommend building on Linux if at all possible. That's the recommended and most supported approach. There's also this: https://justine.lol/cosmopolitan/windows-compiling.html

Keithcat1:
0.2 works, or at least gives the impression of working since it spat
out some kind of thing.
However, this doesn't work:

C:\py\cosmopolitan>llvm-objcopy.exe -SO binary hello.com.dbg hello.com
llvm-objcopy.exe: error: too many positional arguments

I had to do:
C:\py\cosmopolitan>llvm-objcopy.exe hello.com.dbg hello.com

But when I ran hello.com, I got:
C:\py\COSMOP~1>hello
Program too big to fit in memory

jart:
Try doing `llvm-objcopy -S -O binary hello.com.dbg hello.com`. Here's what I did and it worked with the 0.2 release.

```
[...]

PR #343, opened 2022-01-01
Fix Clang/LLD woes!

nocylah:
Several changes to get Clang and LLD working again (GH-180):

- LLD will always place non-allocated sections at LMA(0x0). Flagged the .zip
  section as 'a' to fix this.
- LLD interprets DEFINED(x) in a subtly-different way, where any symbol
  declaration counts as defining. Changed several default-value
  definitions that used DEFINED to instead use PROVIDE_HIDDEN.
- Due to the same LLD/DEFINED behavior, EfiMain was always getting
  linked in. Updated to use a separate flagging variable (USE_EFI).
- Clang needed a few new flags to match GCC frame/stack behavior.


jart:
Thank you but after giving it some thought I've decided that I want to delete the commented Clang/LLD code for the time being. Even when it was working it caused a lot of issues for Linux users.

PR #359, opened 2022-02-20
Fix redbean browser launch before server starts

### LINKED ISSUE #337: redbean: LaunchBrowser not working on macos
redbean 1.5
macos big sur 11.6

If I add in an index.html file then it works fine and I get the server running, but as soon as I add the command `LaunchBrowser()` to the `.init.lua` file, it fails with:

```
error: Uncaught SIGSEGV (SEGV_MAPERR) on Mikas-Mac-Pro pid 5793
  ./redbrean.com.zip
  ENOENT[No such file or directory][2]
```

I've tried with paths `"/"` and `"/index.html"` and `"index.html"` but they all fail with the same issue.

pkulchenko:
Fixes #337.

### DIFF
--- a/tool/net/redbean.c
+++ b/tool/net/redbean.c
@@ -2566,13 +2566,19 @@ static void LaunchBrowser(const char *path) {
   const char *u, *prog;
   sigset_t chldmask, savemask;
   struct sigaction ignore, saveint, savequit;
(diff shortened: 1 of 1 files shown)


jart (approved):
Nice. Thanks for the fix. LGTM

PR #366, opened 2022-03-10
Add support for storing folders in redbean from command line

pkulchenko:
@jart, changes to review:

- added ability to package folders recursively; `-A path/` adds all files in that folder (including `path/` in their path)
- changed `GetAsset` to `GetAssetZip`, as I think this correctly handles the case of the asset existing in the file system (stagedirs), but not in the archive: https://github.com/jart/cosmopolitan/compare/master...pkulchenko:redbean-add-dir?expand=1#diff-a937cc5818818785e079eab14ff76fa5db080fe9bd902d1ab21aa546aa79c7d1R3232
- added storing asset message with INFO level: https://github.com/jart/cosmopolitan/compare/master...pkulchenko:redbean-add-dir?expand=1#diff-a937cc5818818785e079eab14ff76fa5db080fe9bd902d1ab21aa546aa79c7d1R3202

### DIFF
--- a/tool/net/help.txt
+++ b/tool/net/help.txt
@@ -48,7 +48,7 @@ FLAGS
   -R /X=/Y  rewrites X to Y                   [repeatable]
   -K PATH   tls private key path              [repeatable]
   -C PATH   tls certificate(s) path           [repeatable]
-  -A PATH   add asset with path               [repeatable]
+  -A PATH   add assets with path (recursive)  [repeatable]
   -M INT    tunes max message payload size    [def. 65536]
   -t INT    timeout ms or keepalive sec if <0 [def. 60000]
   -p PORT   listen port                       [def. 8080; repeatable]
--- a/tool/net/redbean.c
+++ b/tool/net/redbean.c
@@ -2130,58 +2130,6 @@ static wontreturn void PrintUsage(FILE *f, int rc) {
   exit(rc);
 }
 
-static void GetOpts(int argc, char *argv[]) {
-  int opt;
-  bool storeasset = false;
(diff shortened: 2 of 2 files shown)


jart (approved):
Looks good. In the future we may want to update this to use nftw(). https://github.com/jart/cosmopolitan/blob/master/examples/walk.c

pkulchenko:
Sounds good; the current logic is based on https://github.com/jart/cosmopolitan/blob/master/examples/lstime.c#L41

ahgamut:
`zip x.com ./a.txt` does the same thing as `zip x.com a.txt` 

I expect `StoreFile("./a.txt")` to do the same thing as `StoreFile("a.txt")`, but it doesn't.

pkulchenko:
@ahgamut, are you suggesting the path to be normalized before being stored?

pkulchenko:
@ahgamut, what does your wrapper do? Just drops leading `./`?

@jart, are you okay to drop leading `./` and check for `IsAcceptablePath()` after that, warn and skip?

ahgamut:
@pkulchenko I wrote it like `StoreFile(path, target)` which reads data from `path` and calls `StoreAsset(target, strlen(target), ...)` [here](https://github.com/ahgamut/cosmopolitan/blob/storeasset-py/third_party/python/Python/cosmomodule.c#L232). If `target` is not provided, I try to construct it from `path`.

I like being able to choose/validate the target location inside the zip; it's useful in Python, but the tradeoffs may be different for redbean.

pkulchenko:
@ahgamut, thank you for the suggestion! I submitted #373 that does this (among other things).

PR #368, opened 2022-03-20
Fix EncodeUrl when parameters without values are used

pkulchenko:
@jart, this fixes an issue when provided table has incomplete list of elements (because that's how the user provided it or because it's the result of parsing URL with parameters without values).

This can be seen on redbean demo if click on "Try clicking here" link and then on "redbean.lua?magic" link on the current master.


jart (approved):
LGTM

PR #369, opened 2022-03-20
make.com can build the Cosmopolitan repo

ahgamut:
I was testing out a build when I discovered this!


ahgamut:
On second thought, perhaps it might be better if `make.com` was added to the repo by @jart.
Either way, this is really exciting!

PR #370, opened 2022-03-20
Redbean stream yield implementation

pkulchenko:
@jart, the first commit simplifies the fix that you applied (I moved all the processing inside `LuaCallWithTrace`; the second commit implements streaming/flush using `coroutine.yield`. All my tests work and I haven't seen any issues so far.


pkulchenko:
@jart, one question: should I bump minor version number, as I may need to check for streaming available, but it's not easily discoverable from the Lua code?

tomberek:
Works as expected, thanks!

Issue #358, opened 2022-02-19
redbean: allow symlinks for -D path


mischief:
hi, i tried to run redbean like
```
redbean.com -p 80 -D /data/git
```

but i see
```
E2022-02-19T14:07:40+000016:tool/net/redbean.c:921:redbean:1] (cfg) error: not a directory: "/data/git"
```

because /data/git is indeed a symlink.

in https://github.com/jart/cosmopolitan/blob/master/tool/net/redbean.c#L914, it seems to disallow them. is there a particular reason why?

jart:
Thanks for the report. That's unintended. Should be an easy fix.

mischief:
@jart it looks like there is `isdirectory` and `issymlink`, but is there a call to check if a path is a directory *or* a symlink to one that can be used? or (maybe less ideal), don't check isdirectory and let stats/opens fail later?

pkulchenko:
@mischief, I noticed that `examples/cp.c` simply checks for `(endswith(dst, "/") || isdirectory(dst))`; we can do the same here. The other option would be to something like the following:

```
struct stat st;
fstatat(AT_FDCWD, path, &st, 0) != -1 && S_ISDIR(st.st_mode);
```

This is what `isdirectory` does (at least an equivalent of, per the comment there), just with `AT_SYMLINK_NOFOLLOW`. I think adding `endswith` may be sufficient.

mischief:
@pkulchenko it seems like trying to call fstatat directly might break some behavior because current `isdirectory` can check the embedded zip.

i could either copy isdirectory into a modified function in redbean or add a new argument to isdirectory to allow symlink. which is preferred?

pkulchenko:
I probably would just add `endswith(dst, "/")` check. This is what [I did in the PR I just submitted](https://github.com/pkulchenko/cosmopolitan/commit/f4cd70a9fff8ce59474c72cfd27cbb7860d6de0b#diff-a937cc5818818785e079eab14ff76fa5db080fe9bd902d1ab21aa546aa79c7d1R3529) to cover a similar case in StoreFile.

pkulchenko:
I also considered adding new argument to isdirectory, but it would be an API change and require updating all instances of that call.

@jart, should issymlink return `st_mode` value of the target (as the "true" value) to allow testing it for its mode values? We could then just use `S_ISDIR(issymlink(path))` or something along those lines and it would still be backward compatible (assuming `st_mode` doesn't allow/have 0 as a valid value)?

[...]

PR #372, opened 2022-03-20
redbean: allow symlinks for -D path

mischief:
previously ProgramDirectory called isdirectory which returns false for
symlinks. instead just check directory with stat, and allow user to pass
`-D /tmp/foo/`, where /tmp/foo might be a symlink.

### DIFF
--- a/tool/net/redbean.c
+++ b/tool/net/redbean.c
@@ -938,12 +938,16 @@ static bool HasString(struct Strings *l, const char *s, size_t n) {
 static void ProgramDirectory(const char *path) {
   char *s;
   size_t n;
+  struct stat st;
+
+  if(stat(path, &st) == -1 || !S_ISDIR(st.st_mode)) {
+    DIEF("(cfg) error: not a directory: %`'s", path);
+  }
+
   s = strdup(path);
   n = strlen(s);
-  while (n && (s[n - 1] == '/' || s[n - 1] == '\\')) s[--n] = 0;
-  if (!n || !isdirectory(s)) {
-    DIEF("(cfg) error: not a directory: %`'s", s);
-  }
+
+  INFOF("(cfg) program directory: %s", s);
   AddString(&stagedirs, s, n);
 }
 


mischief:
fixes #358

cc @jart @pkulchenko

pkulchenko:
Thank you for taking time to submit the PR. Just a small comment: the message should probably include "(cfg)" prefix for consistency with other configuration messages.

Welcome as a new contributor. We need to go through one quick hurdle before we can proceed. Could you please email Justine Tunney [jtunney@gmail.com](mailto:jtunney@gmail.com) and say that you intend to assign her the copyright to the changes you contribute to Cosmopolitan? See [CONTRIBUTING.md](https://github.com/jart/cosmopolitan/pull/CONTRIBUTING.md) for further details. We make the ask because it helps us ship the tiniest most legally permissive prim and proper binaries. You also only need to do it one time, so any future changes you send us can be quickly merged. Thanks!

mischief:
done and done.

jart (commented):
That's such an unfortunate gotcha that `isdirectory` needed to have `AT_SYMLINK_NOFOLLOW` by definition but that it ultimately detracted from its original intent which was to be easy. 

Is there any reason why we wouldn't want to just do this?

```
struct stat st;
isdir = stat(path, &st) != -1 && S_ISDIR(st.st_mode);
```

That should follow symlinks and catch the general case.

mischief:
> That's such an unfortunate gotcha that `isdirectory` needed to have `AT_SYMLINK_NOFOLLOW` by definition but that it ultimately detracted from its original intent which was to be easy.
> 
> Is there any reason why we wouldn't want to just do this?
> 
> ```
> struct stat st;
> isdir = stat(path, &st) != -1 && S_ISDIR(st.st_mode);
> ```
> 
> That should follow symlinks and catch the general case.

does this not break zip access?

jart:
It's a good question. I don't think the tohught occurred to me until now to do something like `redbean-demo.com -D /zip/tool/net/demo`. I'll do some investigation after this is merged and see if there's any other bugs that need to be addressed. ZIP also *should* have UNIX mode bits or the NT file attribute mode, e.g. `kNtFileAttributeDirectory`.

jart (approved):

PR #373, opened 2022-03-20
Redbean fixes for ProgramUniprocess and storing symlinked folders with -A

pkulchenko:
This makes several small improvements/fixes:
- Allows -A to store symlinked directories
- Fixes ProgramUniprocess to enable it on Windows (since forking is supported there)
- Improves StoreFile to drop leading `./` in the path for the stored file and to check for reasonable path (also free memory).
- Adds freeing memory in LuaEncodeUrl (which was reported as unfreed with debug build)

### DIFF
--- a/tool/net/demo/maxmind.lua
+++ b/tool/net/demo/maxmind.lua
@@ -297,14 +297,14 @@ local function main()
       if geo then
          Dump(geo)
       else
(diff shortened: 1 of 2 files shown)


jart (approved):
Looks real good. Thanks for sending this.

PR #376, opened 2022-03-28
Fix EncodeUrl redbean function processing parameters without values

pkulchenko:
@jart, I found one more issue with EncodeUrl, so this patch fixes it. I also simplified the logic and updated it to conserve the Lua stack, as it only has 255 values by default, so it's only enough for 82-85 parameters or so. The current logic doesn't depend on the number of parameters, so should handle an arbitrary number (memory allowing).

### DIFF
--- a/third_party/lua/luaencodeurl.c
+++ b/third_party/lua/luaencodeurl.c
@@ -25,9 +25,10 @@
 int LuaEncodeUrl(lua_State *L) {
   size_t size;
   struct Url h;
-  int i, j, k, n, m;
+  int i, j, n;
   const char *data;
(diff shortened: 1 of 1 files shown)


jart (approved):
Looks good. Are there any assertions we could have added that would have caught this?

pkulchenko:
I don't think so, but a test for round-trip between ParseUrl and EncodeUrl should work; something like the following:

```
url = "/path?c&d[1][2]=file&a=1&b="
parsed = ParseUrl(url)
table.insert(parsed.params, {}) -- this is optional, but covers a possible user-generated case as well
EncodeUrl(parsed) == url
```

PR #386, opened 2022-04-17
Build on macOS/Darwin

ghaerr:
Adds the ability to build Cosmopolitan on macOS.

Prerequisites:
[Rich Felker's gcc cross-compiler](https://github.com/FiloSottile/homebrew-musl-cross).
[Gnu Make v4.3](https://stackoverflow.com/questions/43175529/updating-make-version-on-mac).
These can quickly be installed using:
```
brew install filosottile/musl-cross/musl-cross
brew install make --with-default-names
```
Note: Test with `make --version` and make sure NOT v3.8. Using make v3.8 will result in very strange errors, as the $(file...) function is silently ignored.

After installation of the cross compiler and make, the repo can be compiled using:
```
make MODE=darwin
```

@jart, the file build/compiler.mk was added as the CC= variable must be set prior to including build/definitions.mk, or the Linux bootstrapped versions of the cross compiler end up being called due to SPAWNER and COMPILE macros. I thought adding this new file was the best solution, and it is also the easiest to understand and/or modify for future cross-compilers. Ideally, $(MODE) isn't checked in compiler.mk, but instead `uname -s` run and compared with "Darwin". This would separate the cross compiler used from the MODE= options in the future.

The options I selected for MODE=darwin in config.mk are just copied from the Tiny Metallic Unix Mode. Ultimately, we shouldn't have to specify MODE=darwin, but instead be allowed to use any MODE=, just like Linux. This would work if `uname -s` were used in compiler.mk. I don't know enough about the various MODEs to feel comfortable changing that yet.

Of course, build/bootstrap/compile.com will need to be updated; this PR does not supply a binary compile.com.

If accepted, I can change README.md for macOS build instructions.

The libc tests aren't done in test/libc/release/emulate.sh, which will need further work. The build was completing up to the chibicc compile, which then tries to run chibicc.dbg, which of course is an ELF binary and won't run on macOS. That can be looked into later.

On pulling the latest repo to test this PR, the build is failing on test/libc/runtime/mprotect_test.c, for which new repo code must have changed something. I haven't had time to look into that yet. Here is the full error:
```
make MODE=darwin
♥cosmo
error:test/libc/runtime/mprotect_test.c:202: mprotect_testBadProt_failsEinval() on zorg
	EXPECT_EQ(-1, mprotect(p, 0, -1))
		need -1	(or 0xffffffffffffffff) =
		 got 0	(or 0 or ' ')
	EUNKNOWN
	o/darwin/test/libc/runtime/mprotect_test.com.tmp.14409 @ zorg
error:test/libc/runtime/mprotect_test.c:203: mprotect_testBadProt_failsEinval() on zorg
	EXPECT_EQ(EINVAL, errno)
		need 22	(or 0x16 or '▬') =
		 got 0	(or 0 or ' ')
	EUNKNOWN
	o/darwin/test/libc/runtime/mprotect_test.com.tmp.14409 @ zorg
2 / 44 tests failed

`make MODE=darwin -j4 o/darwin/test/libc/runtime/mprotect_test.com.runs` exited with 2:
o/darwin/test/libc/runtime/mprotect_test.com
```
I will look further into the regression tests with fixes, if required, in a separate PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -85,6 +85,7 @@ PKGS =
 
 -include ~/.cosmo.mk
 include build/functions.mk			#─┐
+include build/compiler.mk			#─┐
 include build/definitions.mk			# ├──META
 include build/config.mk				# │  You can build
 include build/rules.mk				# │  You can topologically order
--- a/build/compiler.mk
+++ b/build/compiler.mk
@@ -12,7 +12,8 @@ PKG = build/bootstrap/package.com
 MKDEPS = build/bootstrap/mkdeps.com
 ZIPOBJ = build/bootstrap/zipobj.com
 
-ifeq ($(MODE), darwin)
+ifeq ($(shell uname), Darwin)
+
 TOOLPREFIX=/usr/local/bin/x86_64-linux-musl-
 AS = $(TOOLPREFIX)as
 CC = $(TOOLPREFIX)gcc
--- a/build/compiler.mk
+++ b/build/compiler.mk
@@ -1,0 +1,43 @@
+#-*-mode:makefile-gmake;indent-tabs-mode:t;tab-width:8;coding:utf-8-*-┐
+#───vi: set et ft=make ts=8 tw=8 fenc=utf-8 :vi───────────────────────┘
+#
+# SYNOPSIS
+#
+#   Cosmopolitan Compiler and Toold Definitions
+#
+
+# see build/compile, etc. which run third_party/gcc/unbundle.sh
+AR = build/bootstrap/ar.com
+PKG = build/bootstrap/package.com
+MKDEPS = build/bootstrap/mkdeps.com
+ZIPOBJ = build/bootstrap/zipobj.com
+
+ifeq ($(MODE), darwin)
+TOOLPREFIX=/usr/local/bin/x86_64-linux-musl-
+AS = $(TOOLPREFIX)as
+CC = $(TOOLPREFIX)gcc
+CXX = $(TOOLPREFIX)g++
+CXXFILT = $(TOOLPREFIX)c++filt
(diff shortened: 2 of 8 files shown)


jart:
> Note: Test with make --version and make sure NOT v3.8. Using make v3.8 will result in very strange errors, as the $(file...) function is silently ignored.

Out of curiosity does `o//third_party/make/make.com` work for you on Darwin?

ghaerr:
> Out of curiosity does o//third_party/make/make.com work for you on Darwin?

Hmmm... it isn't being built by default. Trying `make MODE=darwin o/darwin/third_party/make/make.com` however, builds make.com. Then running `./make.com MODE=darwin` does seem to work! The build ends up with the same libc/test/mprotect.c error as before, so I would say it works, but needs to be built or be supplied in build/bootstrap.

ghaerr:
(I didn't try building the whole tree with make.com... would you like to know the result of that?)

ghaerr:
Yes, make.com builds the whole tree using `./make.com MODE=darwin`.

ghaerr:
Hello @jart,

I have simplified the macOS build by eliminating the need for `make MODE=darwin`, which used MODE= as an operating system specifier rather than a build type qualifier. The recent push now allows building on macOS (after having installed Felker's cross-compiler via homebrew) by just running `make`, and special builds using any of the usual `make MODE=` mechanisms.

Essentially, the two changes required for macOS builds to proceed are to consolidate the CC=, AS= etc tool paths into a single compiler.mk file, and to eliminate compile.com's addition of `-no-canonical-prefixes` when running on XNU. There are a couple of other small changes to stop unbundling of the toolchain for Linux, and allow macOS builds in sanitycheck.

Thank you!

PR #387, opened 2022-04-17
Makefile: incorrect executable location

ProducerMatt:
`o/dbg/` only contains `.o` files from what I can see. The desired `.com` is in `o/examples/`


ProducerMatt:
I sent the copyright assignment.

jart (approved):
Looks good. Technically speaking though, the path is correct if you build using `make MODE=dbg`. But this will probably be less surprising. Another cool thing is the canonical way to write `o/examples/crashreport.com` would be `o//examples/crashreport.com`

PR #389, opened 2022-04-21
fix basename error with cp.com

ahgamut:
when the destination is a folder, cp.com should take the basename of the
source file and add it to the destination. cp.c was instead using the
basename function itself as a string (?), and not calling basename.


jart (approved):
Well that's an interesting bug. Thanks for the good catch. I wish we could use the GCC checks for printf arguments. But unfortunately they provide no way of configuring extensions.

Issue #81, opened 2021-02-28
Universal shared libraries?


jammm:
First of all, huge kudos for the amazing work on comopolitan.

While I'm still studying how it works, I hope you can pardon me for asking a noob question here. I was wondering if it's possible to compile cross-platform shared libraries (e..g, .so+.dll for Windows/Linux)?

jart:
I would be willing to merge a change that makes it possible to build Python extensions using Cosmopolitan. 

1. The assembly code is mostly PIC compatible. There are a few places where out of laziness I didn't use the `ezlea` macro that would need to be updated.
2. The APE bootloader would need a linker overridable symbol that can change ET_EXEC in https://github.com/jart/cosmopolitan/blob/master/ape/ape.S
3. A new entrypoint would need to be defined that can set the values of symbols such as `hook_malloc` appropriately.
4. A small fragment of Python loading code would need to be generated which would run during the setup.py phase, which overwrites the first 64-bytes of the shared object binary so it has an MZ header on Windows.
5. A program would need to be written to replace the `objcopy -S -O binary foo.com.dbg foo.com` step with something that would read the external symbol definitions from the .com.dbg binary and then insert them somehow into the .so

If all that doesn't discourage you so far, then contributions are welcome.

pkulchenko:
@jart, why limit this to Python extensions? I'd like to be able to build Lua extensions as well and load them into redbean or Lua executable. I'm good with re-compiling Lua modules using cosmopolitan if needed. Also, Lua support should be fairly simple, as it only relies on being able to load the library and find `luaopen_libname` function in the library.

Now that I think more about it, would this allow the same dynamic library to be loaded on all supported platforms? That would be very interesting...

jart:
Because Python extensions have a .py file that gets run before the native code is loaded and we can put the monkey patching code in there, which turns APE blob into an SO/DLL/DYLIB. Without a script it's not possible to polyglot macho/pe/elf.

pkulchenko:
We can do the same thing with Lua modules: load .lua file first, which will tweak the blob and then load it.

pkulchenko:
@jart, is there a cosmopolitan-provided method to detect the system it's running on? Since the Lua code may need to patch the code differently depending on the platform, what's the best way to find it? Lua provides its own mechanism for this, but it's a compile-time one and since it's compiled on one platform, but runs on multiple ones, I don't think it's going to help here.

jart:
[...]

PR #391, opened 2022-04-24
load C extensions of external packages correctly

ahgamut:
At present, all C extension modules loaded in the APE Python are
built-in modules -- they are compiled into the interpreter instead of
being separate shared objects in some folder.

In the Python3.6 stdlib, the modules usually follow a naming convention
like: X imports a module called _X, which is built from Modules/_X.c.
(example: json imports _json, which is built from Modules/_json.c).

However, external Python packages follow a different naming convention:
X imports a module called ._Y, which is available as a shared object
located in the same directory as X.

(example: markupsafe imports from ._speedups, expecting a _speedups.so
to be available within the markupsafe folder).

This change makes the interpreter load the required module as a builtin
(if present) if the requisite shared object/python file is not found.
The benefit of this is that the external packages can keep the same
naming convention for their C extensions, and once built, the APE will
load those extensions without error.


ahgamut:
@jart this PR is a solution for the [relative imports problem](https://github.com/jart/cosmopolitan/issues/81#issuecomment-905089180) that I mentioned in #81 a while back:

> The only issue is step 4. Since Python allows relative imports inside a package, splitting the C and Python parts of a library means I have to ensure that such imports are handled correctly. This usually means creating .pyi-like stub files, and redirecting the imports to a correctly named C extension. It's easy to do by hand for something like greenlet, numpy would just require more files to be changed.

It doesn't change anything right now, but will help when building external packages into the APE becomes an automatic task.
Perhaps it can be merged later, once we have a bunch of external Python packages that are useful to add.

ahgamut:
proof of concept: https://github.com/ahgamut/cosmopolitan/tree/import-mod-test

The above repo adds the external Python package [Markupsafe](https://github.com/pallets/markupsafe/tree/2.0.1) to APE Python.

The benefit of this PR is that _zero_ glue code was required in of the `.py` files in Markupsafe.
The interpreter handles the issue of importing `markupsafe._speedups` without complaint.

If there is a way to do the changes in https://github.com/ahgamut/cosmopolitan/commit/7f9cef5b2c221ef2153e9a75fa1cbe1671efb318 (and there should be, hidden somewhere in the depths of `setuptools`), then it would be possible to add external python packages with C extensions to the APE Python build.

jart (approved):

PR #393, opened 2022-04-26
Update redbean to stringify error objects

pkulchenko:


jart (approved):
Looks good. Thanks! Before I merge could you push a quick change modifying the `README.cosmo` file so we can keep a record of our local changes?

pkulchenko:
I didn't modify any Lua code, only one of our extensions (`LuaCallWithTrace`) and you appear to be tracking changes to the actual Lua code in `README.cosmo`.

PR #403, opened 2022-05-07
add underscore to baud constants

ahgamut:
some macros in third-party libraries have these constants used as variables which causes clashes


ahgamut:
The BLIS library had some segfaults with the `__thread` specifier, which I just disabled.
The baud constants were the only "change" required to compile the BLIS source code, and they're not really related to the functioning of BLIS.

Everything else was handled by the modifying the build system of BLIS.

jart:
Could we solve this by deleting `libc/sysv/consts/baud.h` and relocating its contents to `libc/isystem/termios.h` instead?

ahgamut:
The clash is just with the `B0` variables being available in the global namespace.
If properly namespaced (via double underscores or some other name), we can keep the files as is.

jart:
Yes but we didn't define these. 

```
jart@nightmare:~/cosmo2$ grep -PR '\bB0\b' /usr/include
/usr/include/asm-generic/termbits.h:#define  B0 0000000         /* hang up */
/usr/include/bits/termios.h:#define B0       0000000
```

There's other precedents for things we only define under `libc/isystem/` and therefore exclude from `cosmopolitan.h`. For example:

```
complex.h:5:#define I _Complex_I
iso646.h:4:#define and &&
stdnoreturn.h:6:#define noreturn wontreturn
sys/dir.h:4:#define direct dirent
```

jart:
If you do this, then I'd say also relocate the contents of libc/calls/ttydefaults.h into libc/isystem/termios.h

ahgamut:
> Yes but we didn't define these.

If that's the case then BLIS should complain even when compiling with the system `/usr/include` ... wait, why would a linear algebra library complain about tty constants?!

Oh, the clash with BLIS happens only because `-include cosmopolitan.h` puts the information of *all* the headers available for compiling each file. It would not happen if we did `-isystem libc/isystem`. 

Ignore this PR, the build issue is because I used the amalgamation header.

ahgamut:
My takeaway is that using `-isystem libc/isystem`, or headers rolled up like `/usr/include` rather than a single amalgamation would prevent these crossfires.

I'll find out how `libc/isystem` can be used as a drop-in replacement to `/usr/include`.

ghaerr:
> I'll find out how libc/isystem can be used as a drop-in replacement to /usr/include.

That would be great. I used a similar scheme when compiling Microwindows outside the Cosmopolitan tree but ended up temporarily using a local copy of the directory (with some headers including cosmopolitan.h) because I didn't have the time or knowledge to track down the correct internal headers needed. I'd be happy to test again after updates, if that's one of the intended uses of libc/isystem.

PR #405, opened 2022-05-07
Redbean fetch headers

pkulchenko:
The second commit fixes a small issue with redbean crashing if there is no listening socket and no host has been provided (in which case `servers.n` is 0).

### DIFF
--- a/tool/net/redbean.c
+++ b/tool/net/redbean.c
@@ -3805,10 +3805,11 @@ static int LuaFetch(lua_State *L) {
       port = usessl ? "443" : "80";
     }
   } else {
-    ip = ntohl(servers.p[0].addr.sin_addr.s_addr);
+    ip = servers.n ? ntohl(servers.p[0].addr.sin_addr.s_addr) : INADDR_LOOPBACK;
     host =
         gc(xasprintf("%hhu.%hhu.%hhu.%hhu", ip >> 24, ip >> 16, ip >> 8, ip));
-    port = gc(xasprintf("%d", ntohs(servers.p[0].addr.sin_port)));
+    port =
+        gc(xasprintf("%d", servers.n ? ntohs(servers.p[0].addr.sin_port) : 80));
   }
   if (!IsAcceptableHost(host, -1)) {
     luaL_argerror(L, 1, "invalid host");
--- a/tool/net/redbean.c
+++ b/tool/net/redbean.c
@@ -3694,6 +3694,11 @@ static int LuaFetch(lua_State *L) {
   struct HttpUnchunker u;
   const char *urlarg, *request, *body, *method;
   char *conlenhdr = "";
+  char *headers = 0;
+  char *hosthdr = 0;
+  char *agenthdr = brand;
+  char *key, *val, *hdr;
+  size_t keylen, vallen;
   size_t urlarglen, requestlen, paylen, bodylen;
   size_t g, i, n, hdrsize;
   int numredirects = 0, maxredirects = 5;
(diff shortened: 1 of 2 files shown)


tool/net/redbean.c
```diff
@@ -3720,6 +3725,41 @@ static int LuaFetch(lua_State *L) {
...
+          if (!IsValidHttpToken(key, keylen))
+            return luaL_argerror(L, 2, "invalid header name");
+
+          val = lua_tolstring(L, -1, &vallen);
+          if (!(hdr = gc(EncodeHttpHeaderValue(val, vallen, 0))))
+            return luaL_argerror(L, 2, "invalid header value encoding");
+
+          // Content-Length and Connection will be overwritten;
+          // skip them to avoid duplicates
+          if (strcasecmp(key, "content-length") &&
```
jart, line 3746: The high performance redbean way to do this, using perfect hash tables, is:

```c
int h;
if ((h = GetHttpHeader(key, keylen)) != -1 &&
    h == kHttpContentLength) {
  // ...
}
```

One thing this allows is, if the HTTP header is hard-coded (i.e. GetHttpHeader returns non -1) then you can round-trip it through `GetHttpHeaderName(h)` if you want to normalize the case, e.g. turning `"content-type"` into `"Content-Type"`.

tool/net/redbean.c
```diff
@@ -3720,6 +3725,41 @@ static int LuaFetch(lua_State *L) {
...
+              strcasecmp(key, "connection")) {
+            if (!strcasecmp(key, "user-agent")) {
+              agenthdr = gc(strdup(hdr));
+            } else if (!strcasecmp(key, "host")) {
+              hosthdr = gc(strdup(hdr));
+            } else {
+              appendd(&headers, key, keylen);
+              appendd(&headers, ": ", 2);
```
jart, line 3754: Rather than explicit length compiler can't check, consider instead either `appendw(&headers, READ16LE(": "))` or `appends(&headers, ": ")`

tool/net/redbean.c
```diff
@@ -3720,6 +3725,41 @@ static int LuaFetch(lua_State *L) {
...
+          val = lua_tolstring(L, -1, &vallen);
+          if (!(hdr = gc(EncodeHttpHeaderValue(val, vallen, 0))))
+            return luaL_argerror(L, 2, "invalid header value encoding");
+
+          // Content-Length and Connection will be overwritten;
+          // skip them to avoid duplicates
+          if (strcasecmp(key, "content-length") &&
+              strcasecmp(key, "connection")) {
+            if (!strcasecmp(key, "user-agent")) {
+              agenthdr = gc(strdup(hdr));
```
jart, line 3749: strdup shouldn't be necessary here because gc() is function scoped, rather than block scoped.

tool/net/redbean.c
```diff
@@ -3765,15 +3805,18 @@ static int LuaFetch(lua_State *L) {
       port = usessl ? "443" : "80";
     }
   } else {
-    ip = ntohl(servers.p[0].addr.sin_addr.s_addr);
+    ip = servers.n ? ntohl(servers.p[0].addr.sin_addr.s_addr) : INADDR_LOOPBACK;
     host =
         gc(xasprintf("%hhu.%hhu.%hhu.%hhu", ip >> 24, ip >> 16, ip >> 8, ip));
-    port = gc(xasprintf("%d", ntohs(servers.p[0].addr.sin_port)));
+    port =
```
jart, line 3811: Can the type of port be uint16_t instead of string? You can then use `%d` instead of `%s` below when formatting it into larger values.

jart (approved):
Looks good. Some suggestions on fine tuning the coding. Please take a look.

tool/net/redbean.c
pkulchenko: I considered this, but was under the impression that `GetHttpHeader(key, keylen)` will check against the request buffer, which is not the case here, as it's a header provided by user to be sent as part of a Fetch request. Am I wrong?

tool/net/redbean.c
pkulchenko: Will update.

tool/net/redbean.c
pkulchenko: oh, right; I can simply do `agenthdr = hdr` and `hosthdr = hdr`, since `hdr` values will be `gc` collected anyway.

tool/net/redbean.c
pkulchenko: I considered updating this, but it's used in `getaddinfo`, which requires conversion back to string anyway.

tool/net/redbean.c
jart: No `GetHttpHeader` is a pure function. It doesn't reference any globals.

pkulchenko: Right; updated.

jart (approved):

PR #406, opened 2022-05-13
METH_FASTCALL backport + compiler optimisation

ahgamut:
@jart this is #328 synced again to master branch + https://github.com/ahgamut/cosmopolitan/commit/d4be3083408ec405811f0ea61b01c109ed299247

I built `python.com` using `make -j4` on Debian Linux 11. All tests pass in `MODE=`, `MODE=opt`, `MODE=optlinux`, `MODE=tiny`, `MODE=asan`, `MODE=dbg` (`test_decimal` is slow but eventually succeeds). Let me know if there are any tests that need to be checked or excluded in specific situations.


jart (approved):
Looks good. I'm still encountering some of the same build flakes. But I believe I have a solution this time that I can push shortly as a follow-up commit. Thank you!

PR #408, opened 2022-05-14
move importlib functions to within C

ahgamut:
@jart this moves some of the frequently-used functions from Python code in `_bootstrap_external.py` to C in `import.c`.

All tests pass in `MODE=`, `MODE=tiny`.

The idea is that almost every line of Python code can cause a memory allocation, and so functions that are frequently called end up wasting memory repeatedly. By moving the Python code to C, it becomes more clear how the function calls occur, and we can attempt more optimizations at the lower C level.

For example: during APE startup, `_validate_bytecode_header` and `code_to_bytecode` are called back-to-back for every `.pyc` import. This leads to a repeated sequence like below

```python
    def get_code(self, fullname):
        # small UnicodeObject allocation for path
        path = self.get_filename(fullname)
        # allocate a large BytesObject from the `.pyc` file
        data = self.get_data(path)
        # allocate another large BytesObject because slicing data[:12] after validating bytecode header
        bytes_data = _validate_bytecode_header(data, name=fullname, path=path)
        # allocate a CodeObject after evaluating the BytesObject
        return _compile_bytecode(bytes_data, name=fullname, bytecode_path=path)
```

In C we can mostly avoid the `BytesObjects` allocations and many function calls by just calling `PyMarshal_ReadObjectFromFile` with a valid `FILE *`.

It's a small speedup, but it adds up if all tests run slightly faster than before.


ahgamut:
Testing on `MODE=dbg` is showing some ubsan errors.

ahgamut:
```diff
diff --git a/libc/zipos/read.c b/libc/zipos/read.c
index 2767dd10c..4fc80fa4d 100644
--- a/libc/zipos/read.c
+++ b/libc/zipos/read.c
@@ -43,7 +43,7 @@ ssize_t __zipos_read(struct ZiposHandle *h, const struct iovec *iov,
   x = y = opt_offset != -1 ? opt_offset : h->pos;
   for (i = 0; i < iovlen && y < h->size; ++i, y += b) {
     b = min(iov[i].iov_len, h->size - y);
-    memcpy(iov[i].iov_base, h->mem + y, b);
+    if(iov[i].iov_base) memcpy(iov[i].iov_base, h->mem + y, b);
   }
   if (opt_offset == -1) h->pos = y;
   return y - x;
```

silences the warnings but a couple of tests have errors.

ahgamut:
The following change enabled me to fix my code that was causing the ubsan warning:

```diff
diff --git a/libc/intrin/ubsan.c b/libc/intrin/ubsan.c
index ec357de0f..c65eb4a88 100644
--- a/libc/intrin/ubsan.c
+++ b/libc/intrin/ubsan.c
@@ -216,6 +216,7 @@ static void __ubsan_warning(const struct UbsanSourceLocation *loc,
                             const char *description) {
   kprintf("%s:%d: %subsan warning: %s is undefined behavior%s\n", loc->file,
           loc->line, SUBTLE, description, RESET);
+  __die();
 }
 
 dontdiscard __ubsan_die_f *__ubsan_abort(const struct UbsanSourceLocation *loc,
```

ahgamut:
Now all tests pass in `MODE=`, `MODE=tiny`, and `MODE=dbg`.

ahgamut:
@jart https://github.com/jart/cosmopolitan/pull/408/commits/e043046e43d3aa9f73f749f55cbcb3f734050c93 is all the code that needs to be changed for locking during imports

jart:
```diff
-    memcpy(iov[i].iov_base, h->mem + y, b);
+    if(iov[i].iov_base) memcpy(iov[i].iov_base, h->mem + y, b);
```

To avoid undefined behavior there, you'd like want:

```diff
-    memcpy(iov[i].iov_base, h->mem + y, b);
+    if(b) memcpy(iov[i].iov_base, h->mem + y, b);
```

jart:
> It's a small speedup, but it adds up if all tests run slightly faster than before.

I'd be interested in seeing the numbers. Startup latency is something I value. It's difficult to do with a change like this, because the system calls themselves usually have a lot of overhead. For instance, a stat() system call can be on the order of a microsecond. Whereas a crossover between Python and C my best guess (not having measured it) would probably be around 100ns. There's also some latency when assets are pulled out and inflated from the zip.

ahgamut:
So I implemented this because I read a [blog post about Python startup times](https://gregoryszorc.com/blog/2018/12/28/faster-in-memory-python-module-importing/), and the question I had was "How can we use the benefits of APE ZIP store + Cosmopolitan Libc to speedup imports in APE Python?"

I tried to do this earlier when submitting #248, but I guess I wasn't as good at writing CPython with Cosmopolitan Libc's support.

Performance is measured in that blog post by just checking `time` for importing all available stdlib packages. 
Simple gist from that post modified for APE Python: [`import_stdlib.py`](https://gist.github.com/ahgamut/b7f4ef4967e9c2316427686df4d8c525)

`time` with current HEAD commit and `MODE=optlinux`:
```
real	0m0.390s
user	0m0.378s
sys	0m0.013s
```

`time` with current HEAD commit + this PR merged and `MODE=optlinux`:
```
real	0m0.366s
user	0m0.350s
sys	0m0.017s
```

If you like I also can run the pyperformance benchmark for startup times. My guess is this PR currently shaves off a few ms per startup (modulo jitter) which saves around a second when it comes to running all the tests.

We can also look at `python.com -m cProfile import_stdlib.py` to see how the calls are distributed.

> For instance, a stat() system call can be on the order of a microsecond. Whereas a crossover between Python and C my best guess (not having measured it) would probably be around 100ns. There's also some latency when assets are pulled out and inflated from the zip.

By moving things like this into C, `ftrace`/`strace` logs become clearer when trying imports, and so this part of import process (most used during startup) will be easier to optimize in C than in Python. (Debugging the C is easier too thanks to the wonderful `ubsan` runtime :heart:)

For example, we can add a bool into the `SourcelessFileLoader` struct to avoid an unnecessary `stat` call, which is a change that would be difficult to observe+do/not be as effective if done in Python. I imagine other ZIP store/Cosmo-specific tricks also become viable.

jart (approved):

PR #410, opened 2022-05-20
os.realname variable for correct USER_SITE value

ahgamut:
In site.py, Python uses os.name to decide where the USER_SITE (ie the
folder containing the user's locally installed packages) is located.
With cosmo we have set os.name as "posix" always, so we use a new
os.realname to decide the USER_SITE location.

VirusTotal screenshot showing that `python.com` picks the right `site-packages` location:

2022-05-20_20-32-59_965x575

### DIFF
--- a/third_party/python/Lib/os.py
+++ b/third_party/python/Lib/os.py
@@ -42,6 +42,7 @@ def _get_exports_list(module):
 
 name = 'posix'
 linesep = '\n'
+realname = "nt" if cosmo.kernel == "nt" else "posix"
 from posix import *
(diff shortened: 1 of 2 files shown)


ahgamut:
@jart The general question to consider is: when Python asks for something OS-related (path separators, environment variables etc.), do we provide it from the OS and handle at the Python level, or do we provide from cosmo and handle inside the libc?

I'll have a better idea once I run Python unit tests on Windows also. It may be that this new variable `os.realname` is not needed.

jart (approved):

PR #416, opened 2022-06-03
Fix GDT limit in GetDescriptor

sjothen:
The GDT limit field should be 8N - 1 where N is the number of segment descriptors. So if you have 3 descriptors, the total size is 24 bytes but should be stored as 23 bytes:

From Intel SDM Vol. 3 Section 3.5.1:

> Because segment descriptors are always 8 bytes long, the GDT limit should always be one less than an integral multiple of eight (that is, 8N – 1)

This fixes a protection fault I got when using the third descriptor which should range from 16 to 24 bytes offset but GetDescriptor returns -1 because 24 bytes is greater than the 23 byte limit. So a quick fix is to just compare with limit + 1.

BTW, thanks for the library!


G4Vi:
Closing as @tkchia mentions this has been moved to Blink and fixed there.

PR #422, opened 2022-06-15
add remaining complex functions from musl

ahgamut:
* from https://git.musl-libc.org/cgit/musl/snapshot/musl-1.2.2.tar.gz
* only added functions that cosmo did not have already (for e.g. `cabs` and `csqrt` were not added)
* the musl license has been added to the start of each C file
* some macros from `musl-1.2.2/src/internal/libm.h` have been added `libc/tinymath/complex.internal.h`
* the `cpow` family of functions need `-ffast-math` otherwise at the linker stage it says:
    * `cpow` needs `__muldc3`
    * `cpowf` needs `__mulsc3`
    * `cpowl` needs `__mulxc3`

repo builds without error in `MODE=`.


jart (approved):
Looks good. Thank you!

Issue #423, opened 2022-06-15
`(vc)scanf` ignores field width for numbers


lourkeur:
# Description

This subverts expectations in some niche cases, e.g. parsing HTML RGB literals. Consider:
```c
int main() {
	const char *color = "#323030";
	int r, g, b;

	sscanf(color, "#%2x%2x%2x", &r, &g, &b);
	printf("%x %x %x\n", r, g, b);

	return 0;
}
```
Expected output: `32 30 30`
Actual output: `323030 4016cb 7ffe`

# Workaround
scan into 1 int and bitfiddle

# Impact
Desktop background got a green tinge. :nauseated_face: 
![hp4qf5ngzbkgk3m4diijlnsrfgp4mz7j-background](https://user-images.githubusercontent.com/15657735/173928117-79ca1b1a-f109-4f21-bcf2-006bc8249237.png)

PR #424, opened 2022-06-16
scan fixed-width integer format in vcscanf

ahgamut:
when a format string like %2x is provided, the width parameter was read
correctly as 2, but it was not used when decoding the number from the
input string (ie instead of reading 2 characters from the input string,
vcscanf read all the characters).

this change uses the value of width within the number decoding loop to
read fixed number of digits correctly. if width is zero (not provided),
the default of width is set as bits.

also added a test for this specific use case.

### DIFF
--- a/libc/fmt/vcscanf.c
+++ b/libc/fmt/vcscanf.c
@@ -177,9 +177,11 @@ int vcscanf(int callback(void *), int unget(int, void *), void *arg,
       DecodeNumber:
         if (c != -1) {
           number = 0;
+          width = !width ? bits : width;
           do {
             diglet = kBase36[(unsigned char)c];
             if (1 <= diglet && diglet <= base) {
+              width -= 1;
               number *= base;
               number += diglet - 1;
             } else if (thousands && diglet == ',') {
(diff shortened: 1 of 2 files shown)


ahgamut:
@jart this PR is wrt #423

lourkeur (approved):
All I know is it works on my program, lourkeur/distro@71e4deb836bf26cd1ac48195b26ea8e31b9248fa

lourkeur:
Tricky case: if it goes to `ConsumeBasePrefix`, should it count towards width?

ahgamut:
@lourkeur can you provide an example of the tricky case?

lourkeur:
@ahgamut Yes, the following program reads 32 into `r`

```c
int main() {
        int r = 0xdeadbeef;
  sscanf("0x20", "%2x", &r);
  printf("%d\n", r);
  return 0;
}
```
The equivalent GLibC program reads 0 into `r`. I believe that's closer to the standard, but I'm not 100% sure.

ahgamut:
Hmm so glibc does read only 0. I wonder what musl does for this case.

purinchu:
> Hmm so glibc does read only 0. I wonder what musl does for this case.

I just tried it. For

```c
/* musl outputs -559038737 (i.e. 0xdeadbeef) */
sscanf("0x20", "%2x", &r)
```

Seems musl doesn't even read the zero?

For

```c
/* musl outputs 32 */
sscanf("0x20", "0x%2x", &r)
```

jart (approved):
Thanks for patience. This is an outstanding change. I absolutely LOVE the fact that you wrote a unit test for this change. TDD is awesome and one of the things this codebase aims to make easy.

PR #425, opened 2022-06-16
add sys.meta_path entry for APE zip store

ahgamut:
@jart this PR adds a `CosmoImporter` as the first entry in `sys.meta_path`, thereby allowing a glance (and room for optimizations) into another part of Python's import process at the C level.

Entries in `sys.meta_path` return a `ModuleSpec` instance if they can find a particular module, otherwise they return `None`.
With `CosmoImporter` it is now possible to do things like: 

* have a static `std::map<modulename, location_in_zip_store>` to avoid `stat` syscalls for modules that are part of the startup process
* prevent unnecessary syscalls/Python operations because we know stdlib modules have to be in the APE zip store

With `CosmoImporter`, the `SourcelessFileLoader` added in #408 now has an additional attribute to avoid a `stat` syscall.

For a rough benchmark I again use the [`import_stdlib.py`](https://gist.github.com/ahgamut/b7f4ef4967e9c2316427686df4d8c525) from #408, which imports all available packages in the APE stdlib.

with `MODE=optlinux` and current HEAD commit https://github.com/jart/cosmopolitan/commit/c06ffd458c984077f0aab2a324f4cb7df3530663:

```
real    0m0.307s
user    0m0.282s
sys     0m0.025s
```

with `MODE=optlinux` and current HEAD commit merged with this PR:
```
real    0m0.297s
user    0m0.289s
sys     0m0.009s
```

Not much, but by writing better versions of `is_builtin` and `find_frozen` and a lookup table for zip store modules, it can be reduced further.


ahgamut:
All tests pass in `MODE=` and `MODE=optlinux`. Running tests in other `MODE`s...

ahgamut:
Ok I implemented a simple lookup table with binary search to check for modules.
(logarithmic, but each comparison is a `strcmp` with indirection involved).

It doesn't seem to have improved on the `import_stdlib.py` benchmark that much, but here's something interesting:

`python.com` built in `MODE=` Current HEAD commit https://github.com/jart/cosmopolitan/commit/67b28b9af199f3ba9118c3e1196e92c0d3b3a0d8

```
o//third_party/python/python.com --strace -c "2+2" | grep fstat | wc -l
69
```

with merging this PR it becomes

```
o//third_party/python/python.com --strace -c "2+2" | grep fstat | wc -l
17
```

The table `ZipCdir_Lookup` that I added for zip store modules contains entries for those modules used in APE startup, so we avoid stat calls during startup completely.

ahgamut:
All `python.com` tests still pass in MODE=, MODE=tiny, MODE=dbg, MODE=opt,MODE=asan, and MODE=optlinux on my Debian 11 system.

ahgamut:
`CosmoImporter` speeds up APE startup by 5% in the default case, as measured via `pyperformance`:

| Benchmark              | APE-optlinux-beforeimpmod | APE-optlinux-afterimpmod |
|------------------------|:-------------------------:|:------------------------:|
| python_startup_no_site | 9.54 ms                   | 9.06 ms: 1.05x faster    |
| python_startup         | 14.5 ms                   | 13.9 ms: 1.04x faster    |
| Geometric mean         | (ref)                     | 1.05x faster             |

Another interesting thing I noticed is that APE Python startup time is now competitive with Anaconda Python 3.7:

| Benchmark              | pyconda37-11 | APE-optlinux-afterimpmod |
|------------------------|:------------:|:------------------------:|
| python_startup_no_site | 9.79 ms      | 9.06 ms: 1.08x faster    |
| python_startup         | 14.0 ms      | 13.9 ms: 1.00x faster    |
| Geometric mean         | (ref)        | 1.04x faster             |

Would be nice to get a separate computer to measure benchmarks without jitter.

jart (approved):

Issue #431, opened 2022-06-18
Migrate from Travis CI to another CI (e.g., GitHub Actions)?


mbrukman:
This repo currently has a [`travis.yml`][travis-yml] and no other CI config file, but Travis CI has [changed its pricing and support model for OSS projects a while back][travis-ended-oss-support], and both of these URLs are now displaying errors:

* https://www.travis-ci.org/jart/cosmopolitan
* https://www.travis-ci.com/jart/cosmopolitan

This means that the project is currently operating with a CI that runs automatically on every PR, whereas it used to before.

Given that the entire Travis CI config is very simple, amounting to `make -j4 V=0`:

https://github.com/jart/cosmopolitan/blob/67b28b9af199f3ba9118c3e1196e92c0d3b3a0d8/.travis.yml#L1-L6

it should be pretty straight-forward to replicate with GitHub Actions. They don't support the `virt: lxd` parameter that Travis CI supported, but I'm guessing this was an optimization to have faster spin-up while still having `sudo` access? Overall, I've found GitHub Actions to be pretty fast, so this may not be an issue.

GitHub actions has [a lot of docs][ga-docs], but to get started quickly, you can probably adapt a variation of this [default blank template that GitHub provides][ga-template] and get up and running very quickly. I'm using something [very similar][my-ga-config].

[travis-yml]: https://github.com/jart/cosmopolitan/blob/67b28b9af199f3ba9118c3e1196e92c0d3b3a0d8/.travis.yml
[travis-ended-oss-support]: https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing
[ga-docs]: https://docs.github.com/en/actions
[ga-template]: https://github.com/jart/cosmopolitan/new/master?workflow_template=blank
[my-ga-config]: https://github.com/mbrukman/c-stdlib/blob/main/.github/workflows/main.yml

jart:
Could you help us do that? I haven't had time to learn how to use GitHub actions. The lxd thing was for a more lightweight environment if I remember. Our build doesn't require sudo access. It doesn't even require that the make command be installed anymore, since you can use `build/bootstrap/make.com`.

thetanil:
started on this. It's absolutely basic, but i would like some feedback and a green build before I move too much further.

i put the badge under the badger. i can move it, open to suggestions.

the build is failing on testcves on my local debian as well as gh runners, so I guess it's correct that there are failing tests ATM.

[...]

PR #440, opened 2022-06-22
Issues 431 gh actions

thetanil:
first step on #431 - basic build with github actions. now running green.

tests disabled:
- test/libc/calls/execve_test.c : HasMzHeader  todo(jart) added
- third_party/python/python.mk: disabled os and ioctl tests 
    these tests use tty. they have "SkipTest" handling when tty is not a real tty, but that also causes the tests to fail. Not sure why.


.github/workflows/build.yml
```diff
@@ -0,0 +1,32 @@
+name: build
+
+on:
+  push:
+    branches: 
+      - "master"
+      - "issues-431-gh-actions"
```
jart, line 7: Shouldn't this line be removed before merging?

third_party/python/python.mk
```diff
@@ -2610,9 +2610,9 @@ o/$(MODE)/third_party/python/Lib/test/test_pyexpat.py.runs: \
 		o/$(MODE)/third_party/python/pythontester.com
 	@$(COMPILE) -ACHECK -tT$@ $(PYHARNESSARGS) $< -m test.test_pyexpat $(PYTESTARGS)
 
-o/$(MODE)/third_party/python/Lib/test/test_ioctl.py.runs: \
-		o/$(MODE)/third_party/python/pythontester.com
-	@$(COMPILE) -ACHECK -tT$@ $(PYHARNESSARGS) $< -m test.test_ioctl $(PYTESTARGS)
+#o/$(MODE)/third_party/python/Lib/test/test_ioctl.py.runs: \
```
jart, line 2613: Could you comment out the test code itself and leave a todo? I'm surprised this doesn't break the build, since these tests are still listed under `THIRD_PARTY_PYTHON_CHECKS` via `$(THIRD_PARTY_PYTHON_PYTEST_PYMAINS:%=o/$(MODE)/%.runs)`

jart (changes requested):

pkulchenko:
@thetanil, btw, you can avoid all the merge commits by rebasing against master and then force-pushing. Force-pushing is fine for feature branches and rebase allows to keep the history clean with only changes relevant for the PR.

.github/workflows/build.yml
thetanil: done

third_party/python/python.mk
thetanil: done

thetanil:
follow up PR which is not a mess already created.

PR #450, opened 2022-06-26
Redbean report write lock failure

pkulchenko:
@jart, I ran into several cases where StoreAsset or -A option doesn't work because a write lock can't be acquired (for example, because of setcap permissions) and this allows a message to be logged to continue (instead of crashing the process).

I think we'll want to modify StoreAsset to report true/false on success instead of just logging a message in the future (especially when it's available on all platforms).

Also did a bit of a log message clean up for consistency.


jart (approved):
Looks good. Could we also have the Lua function report that `StoreAsset` failed?

Issue #429, opened 2022-06-17
Hello.com, redbean.com fail to run on MacOS M1 under Rosetta


pkulchenko:
hello.com is built on linux following instructions published here: https://justine.lol/cosmopolitan/ and then executed on a MacBook Pro with the M1 Max chip (arm64) running macOS Monterey 12.4 under Rosetta translation.

```sh
#To confirm that it's running under Rosetta:
❯ arch -arch x86_64 ./[hello.com](http://hello.com/)
[1]    79473 segmentation fault  arch -arch x86_64 ./[hello.com](http://hello.com/)

❯ sh ./[hello.com](http://hello.com/)
[1]    73377 segmentation fault  sh ./[hello.com](http://hello.com/)
❯ sh ./[hello.com](http://hello.com/) --assimilate
❯ ./[hello.com](http://hello.com/)
[1]    73395 segmentation fault  ./[hello.com](http://hello.com/)
❯ lldb --file ./[hello.com](http://hello.com/)
(lldb) target create "./[hello.com](http://hello.com/)"
Current executable set to '/Users/user/cosmo/[hello.com](http://hello.com/)' (x86_64).
(lldb) r
Process 73409 launched: '/Users/user/cosmo/[hello.com](http://hello.com/)' (x86_64)
Process 73409 stopped
* thread #1, stop reason = EXC_BAD_ACCESS (code=1, address=0x1f)
    frame #0: 0x000000000040239f [hello.com](http://hello.com/)
->  0x40239f: movl   (%rsp), %ebx
    0x4023a2: leaq   0x8(%rsp), %rsi
    0x4023a7: leaq   0x10(%rsp,%rbx,8), %rdx
    0x4023ac: movq   %rsp, 0x1e97d(%rip)
Target 0: ([hello.com](http://hello.com/)) stopped.
(lldb) bt
* thread #1, stop reason = EXC_BAD_ACCESS (code=1, address=0x1f)
  * frame #0: 0x000000000040239f [hello.com](http://hello.com/)
(lldb)
❯ file [hello.com](http://hello.com/)
[hello.com](http://hello.com/): Mach-O 64-bit executable x86_64
```

Full crash report:

```json
[...]

PR #453, opened 2022-06-27
Work around Rosetta clobbering startup registers on M1 Macs (issue #429)

dkulchenko:
Rosetta doesn't correctly respect the startup registers as defined in LC_UNIXTHREAD
which makes platform detection go awry. But at least Rosetta appears to consistently
set rbx to 0x00000000ffffffff and rdx to 0x0000000000000001 at startup for every
x64 executable I could get my hands on. So we use that to detect Rosetta's presence
and set up the correct registers for XNU.

Thanks to @pkulchenko and @jart for pointers on getting this done :)

### DIFF
--- a/ape/loader-elf.S
+++ b/ape/loader-elf.S
@@ -213,7 +213,18 @@ macho:	.long	0xFEEDFACE+1
 //
 //	@see	APE_LOADER_ENTRY
 //	@see	ape/loader.h
-_start:	mov	%rsp,%rsi
+_start:
+
+//	Hack for detecting M1 Rosetta environment.
+//	https://github.com/jart/cosmopolitan/issues/429#issuecomment-1166704377
+	cmp	$-1,%ebx
+	jne	0f
+	cmp	$+1,%edx
+	jne	0f
+	mov	$XNU,%dl
+	mov	$0,%rcx
+
+0:	mov	%rsp,%rsi
 	jmp	ApeLoader
 	.endfn	_start,globl
 
--- a/ape/loader-elf.S
+++ b/ape/loader-elf.S
@@ -222,7 +222,7 @@ _start:
 	cmp	$+1,%edx
 	jne	0f
 	mov	$XNU,%dl
-	mov	$0,%rcx
+	xor	%ecx,%ecx
 
 0:	mov	%rsp,%rsi
 	jmp	ApeLoader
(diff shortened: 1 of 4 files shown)


libc/crt/crt.S
```diff
@@ -31,6 +31,18 @@
...
 
+#if SupportsXnu()
+//	Hack for detecting M1 Rosetta environment.
+//	https://github.com/jart/cosmopolitan/issues/429#issuecomment-1166704377
+	cmp	$-1,%ebx
+	jne	0f
+	cmp	$+1,%edx
+	jne	0f
+	mov	$XNU,%cl
+	mov	$0,%rdi
```
jart, line 42: Is there any reason you can't avoid the need for clearing RDI by moving this piece of code below the FreeBSD check?

ape/loader-elf.S
```diff
@@ -213,7 +213,18 @@ macho:	.long	0xFEEDFACE+1
...
+_start:
+
+//	Hack for detecting M1 Rosetta environment.
+//	https://github.com/jart/cosmopolitan/issues/429#issuecomment-1166704377
+	cmp	$-1,%ebx
+	jne	0f
+	cmp	$+1,%edx
+	jne	0f
+	mov	$XNU,%dl
+	mov	$0,%rcx
```
jart, line 225: You could save 5 bytes of code size here by encoding the operation as follows:

```asm
   0:   48 c7 c1 00 00 00 00    mov    $0x0,%rcx
   7:   31 c9                   xor    %ecx,%ecx
```

With x86-64 assembly, operations on the 32-bit versions of registers (i.e. `%e*` vs. `%r*`) will always clear the top bits as a side-effect. The 32-bit versions of instructions are usually shorter too, because they don't need to use a REX byte (i.e. 0x4x).

jart (changes requested):
Great hack. Only a few small suggestions. Welcome as a new contributor! Per CONTRIBUTING.md could you please send a quick email to me (jtunney@gmail.com) stating that you intend to assign me the copyright to the changes you contribute to Cosmopolitan?

libc/crt/crt.S
jart: Thanks for explaining! What you did sounds good then.

jart (approved):

PR #469, opened 2022-07-09
ttyraster: reset ANSI attributes after each line

wchargin:
This fixes a bug that appears when using `printimage` in my `xterm` or
`gnome-terminal`. When printing a line causes the viewport to scroll,
the last color painted would extend to the last column of the window
rather than the last column of the image. (This is especially noticeable
when rendering a narrow image in a wide window.) A similar artifact
would occur with the last line, which would appear in the color of the
bottom-right corner of the image.

By slapping an `\e[0m` before each newline and before the end of the
image, we avoid these artifacts.

![Screenshots of the output of `printimage` before and after this
change. Each screenshot shows `lemurs.png` printed twice; the first copy
is only partially visible because the window has scrolled. In the
"before" picture, the second copy has extra bands of color off to the
right of the image starting about halfway down, plus a gray band beneath
the image. The "after" picture has no such artifacts.][ss]

[ss]: https://user-images.githubusercontent.com/4317806/178096633-9a9e46a7-360b-49c1-8569-e6b401b81d09.png

wchargin-branch: ttyraster-reset-each-line

### DIFF
--- a/dsp/tty/ttyraster.c
+++ b/dsp/tty/ttyraster.c
@@ -775,7 +775,7 @@ char *ttyraster(char *v, const struct TtyRgb *c, size_t yn, size_t xn,
   struct Glyph glyph;
   struct TtyRgb chun[4], lastchunk[4];
   for (y = 0; y < yn; y += 2, c += xn) {
-    if (y) *v++ = '\r', *v++ = '\n';
+    if (y) v = stpcpy(v, "\e[0m\r\n");
     for (x = 0; x < xn; x += 2, c += 2) {
       CopyChunk(chun, c, xn);
       if (ttyquant()->alg == kTtyQuantTrue) {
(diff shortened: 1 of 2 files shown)


wchargin:
I'm far from an ANSI wizard. This patch seemed reasonable enough to me
(if it's okay to set the attributes, it feels okay to reset them), but
maybe this should be `\e[0;10m` instead or maybe it shouldn't be an ANSI
code at all or maybe something else. Feel free to flag; happy to revise.

jart (approved):
LGTM. Thank you for spotting this and taking the time to send a PR.

Welcome as a new contributor. We need to go through one quick hurdle before we can proceed. Could you please email Justine Tunney <jtunney@gmail.com> and say that you intend to assign her the copyright to the changes you contribute to Cosmopolitan? Please send the email under your legal name rather than anonymously. See [CONTRIBUTING.md](CONTRIBUTING.md) for further details. We make the ask because it helps us ship the tiniest most legally permissive prim and proper binaries. You also only need to do it one time, so any future changes you send us can be quickly merged.

Thanks!

jart:
Thanks for sending the copyright email. Build failure appears unrelated, due to an issue I solved in the last hour. I'll go ahead and merge. Thanks!

PR #470, opened 2022-07-09
Redbean fix listen leak

pkulchenko:
@jart, this should release the resources allocated to closed sockets. Can you review and check to see if you're getting any leak reports with these changes? I do not, but I'm not sure if I'm testing under the same conditions you are.


jart (approved):

jart:
Let's try it. Thanks.

PR #473, opened 2022-07-10
add json.org tests for DecodeJson

ahgamut:
* https://www.json.org/JSON_checker/test.zip
* downloaded and extracted all JSON test cases
* converted `*.json` into `*.lua` by pasting as multiline string (added spaces around `[` and `]` to avoid Lua issues with `[[` and `]]`)

the current `ljson.c` parser is lenient according to the json.org test suite -- all the `pass.lua` testcases pass, but all the `fail.lua` test cases do not fail.


jart (approved):
Looks real nice. Thank you. One thing you might consider trying, if you haven't considered it already, in there's a convention I've been using more and more (per Paul's recommendation) where Lua errors can be propagated elegantly as overloaded return values rather than pcall. It's used extensively in the UNIX APIs. Take a look and let me know if you think that might be better for the JSON APIs. CC: @pkulchenko

jart:
Looks like there's GitHub Actions errors. Once those are resolved I can merge.

ahgamut:
@jart the Github action errors are related to the strictness of `ljson.c`, which currently allows some classes of input that the JSON spec and python's `json.loads` would error out as invalid.

This PR adds some code to `ljson.c`only to make it follow the JSON spec in terms of strictness -- without these changes the `ljson.c` would still accept everything that is valid JSON (ie the `pass.lua` test cases work without any change).

Here's an example failure case that is now handled due to this PR, but perhaps not needed:

```json
```
{unquoted_key: "keys must be quoted"}
```
JSON says keys can only be quoted strings (not integers, even though Lua allows that for tables).

Here's a failure case that is yet to be handled:

```json
[ "Extra closing ]"] ] 
```

The parser correctly picks the needed list and exits without error.

PR #476, opened 2022-07-11
add depth limit for JSON

ahgamut:
- json depth limit set as 1024
- STACK_FRAME_UNLIMITED for ljson.c
- every call of Parse checks if the limit will be crossed

alternative: check stack at the start of Parse everytime, and adjust
when necessary.

### DIFF
--- a/tool/net/ljson.c
+++ b/tool/net/ljson.c
@@ -27,6 +27,8 @@
 #include "third_party/lua/ltests.h"
 #include "third_party/lua/lua.h"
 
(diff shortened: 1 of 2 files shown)


jart (approved):
Thanks. It'd be great to have those tests. Ask the guy who filed the issue how he managed to incorporate that test suite. I'd love to have it in the repo. Especially if a script could be written to turn their great big HTML file full of tests, into just one .lua file test.

PR #482, opened 2022-07-12
make ljson more strict

ahgamut:
```
- json spec says only characters after 0x20 are allowed in strings
- add new context ARRAY_SINGLE for [singleton] without commas
- add new context ARRAY_END for [] without commas
- add new context OBJECT_END for {} without commas
- call Parse() once before looping for array or object so that the empty
  case is handled easily ([,] should now be wrong)
- narrow use of OBJECT_KEY -- now it is only used to confirm that the
  key is a string
- rename ARRAY_VAL to AFTER_VALUE -- a comma will only be present after
  a value ie either [1,2] or {"a":1,"b":2}, so the context for comma is
  always after value
- reset context when AFTER_VALUE and OBJECT_VAL to avoid extra ,:
- use bitfields instead of equality to check the context value
- OBJECT_KEY may be unnecessary, can be checked in object loop
- TOP_LEVEL may be unnecessary, can be checked via depth value
```

result of above updates:

- all but one test from `json.org` now behave as expected
- test cases with extra punctuation from `JSONTestSuite` now behave as expected
- ljson is now only lenient when it comes to poorly formatted numbers and missing comma/colon


jart (approved):
Nice change. Thank you!

ahgamut:
@jart ljson now is only lenient for:

- missing commas or colons
- formatting of numbers

for missing commas or colons we can have another `UNLIKELY` check at the start of each value, like below

```c
case '"':
  if (UNLIKELY(0 != (context & (OBJECT_VAL | AFTER_VALUE)))) goto MissingPunctuation;
  // carry on as usual
```

jart (approved):
Hmm hmm hmm. Very good. Approved you change is.

PR #483, opened 2022-07-13
throw parsing errors for missing commas and colons

ahgamut:
context will be set as zero if a comma is found `AFTER_VALUE` or a colon
is found before `OBJECT_VAL`, so we add an `UNLIKELY` check before parsing
each value to confirm that these two fields are set to zero in context.

this change now means that `ljson` complies with all the tests from
json.org, and only deviates from JSONTestSuite's expected behavior when
it comes to parsing numbers.


jart (approved):
Very impressive. Do you think we should ask them to include us in their JSON compliance chart now?

PR #486, opened 2022-07-15
typo

neuhaus:


neuhaus:
i don't think swapping two bytes is copyrightable.

jart (approved):
LGTM. Thank you for spotting this and taking the time to send a PR.

Welcome as a new contributor. We need to go through one quick hurdle before we can proceed. Could you please email Justine Tunney <jtunney@gmail.com> and say that you intend to assign her the copyright to the changes you contribute to Cosmopolitan? Please send the email under your legal name rather than anonymously. See [CONTRIBUTING.md](CONTRIBUTING.md) for further details. We make the ask because it helps us ship the tiniest most legally permissive prim and proper binaries. You also only need to do it one time, so any future changes you send us can be quickly merged.

I understand your viewpoint on the matter. We can't make exceptions to the process because we want to be fair.

Thanks!

PR #492, opened 2022-07-16
Update Redbean help.txt

w13b3:
Hey,

I have been using Lua functions that are set in redbean.c but are not mentioned in the documentation.
For example: IsHiddenPath was already available in Redbean 0.3 but until now not documented.
So I took the time to fix some typos and write the documentation for the functions that were missing.


jart (approved):
Thank you for doing this! Approved.

jart:
Website updated with this change. Thanks again!

PR #493, opened 2022-07-18
Unveil improvements

The-King-of-Toasters:
- Merge Multiple masks to just one.
- Add documentation to for sys_unveil.
- Inline the chomp function in the unveil tool.

### DIFF
--- a/libc/mem/unveil.c
+++ b/libc/mem/unveil.c
@@ -32,43 +32,44 @@
 #include "libc/sysv/consts/s.h"
 #include "libc/sysv/errfuns.h"
 
+#define UNVEIL_READ                                             \
+  (LANDLOCK_ACCESS_FS_READ_FILE | LANDLOCK_ACCESS_FS_READ_DIR | \
+   LANDLOCK_ACCESS_FS_REFER)
+#define UNVEIL_WRITE (LANDLOCK_ACCESS_FS_WRITE_FILE)
+#define UNVEIL_EXEC  (LANDLOCK_ACCESS_FS_EXECUTE)
+#define UNVEIL_CREATE                                             \
+  (LANDLOCK_ACCESS_FS_MAKE_CHAR | LANDLOCK_ACCESS_FS_MAKE_DIR |   \
+   LANDLOCK_ACCESS_FS_MAKE_REG | LANDLOCK_ACCESS_FS_MAKE_SOCK |   \
+   LANDLOCK_ACCESS_FS_MAKE_FIFO | LANDLOCK_ACCESS_FS_MAKE_BLOCK | \
+   LANDLOCK_ACCESS_FS_MAKE_SYM)
+
 #define FILE_BITS                                                 \
   (LANDLOCK_ACCESS_FS_READ_FILE | LANDLOCK_ACCESS_FS_WRITE_FILE | \
    LANDLOCK_ACCESS_FS_EXECUTE)
 
-/*
+/**
  * Long living state for landlock calls.
- * The bits are set at runtime to handle future API additions.
+ * fs_mask is set to use all the access rights from the latest landlock ABI.
+ * On init, the current supported abi is checked and unavailable rights are
+ * masked off.
+ *
  * As of 5.19, the latest abi is v2.
  *
  * TODO:
- *  - Documentation for sys_unveil.
  *  - Integrate with pledge and remove the file access?
  *  - Stuff state into the .protected section?
  */
 _Thread_local static struct {
-  int abi;
+  uint64_t fs_mask;
   int fd;
-  uint64_t read;
-  uint64_t write;
-  uint64_t exec;
-  uint64_t create;
 } State = {
-    .abi = 2,
-    .read = LANDLOCK_ACCESS_FS_READ_FILE | LANDLOCK_ACCESS_FS_READ_DIR |
-            LANDLOCK_ACCESS_FS_REFER,
-    .write = LANDLOCK_ACCESS_FS_WRITE_FILE,
-    .exec = LANDLOCK_ACCESS_FS_EXECUTE,
-    .create = LANDLOCK_ACCESS_FS_REMOVE_DIR | LANDLOCK_ACCESS_FS_MAKE_CHAR |
-              LANDLOCK_ACCESS_FS_MAKE_DIR | LANDLOCK_ACCESS_FS_MAKE_REG |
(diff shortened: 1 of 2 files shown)


The-King-of-Toasters:
You mentioned locking the state before adding a rule. Marking it `_Thread_local` take care of this issue?

libc/mem/unveil.c
```diff
@@ -135,7 +132,49 @@ static int sys_unveil_linux(const char *path, const char *permissions) {
...
+ *
+ * Using unveil is irreversible. On OpenBSD, the first call immediately enforces
+ * the filesystem visibilty, and existing paths can only be updated with equal
+ * or lesser permissions. Filesystem operations that try to access invisible
+ * paths will raise ENOENT, and operations without the correct permissions raise
+ * EACCES. Unveiling can be disabled by either passing two NULL arguments or by
+ * calling pledge() without the "unveil" promise.
+ *
+ * Landlock is more permissive than OpenBSD's unveil. Filesystem visibility is
+ * only enforced after disabling, and path permissions can be increased at any
```
jart, line 155: What does this mean path permissions can be increased at any time? Does that mean it's reversible?

libc/mem/unveil.c
```diff
@@ -135,7 +132,49 @@ static int sys_unveil_linux(const char *path, const char *permissions) {
...
+ * On OpenBSD, accessing paths outside of the allowed set raises ENOENT, and
+ * accessing ones with incorrect permissions raises EACCES. On Linux, both these
+ * cases raise EACCES.
+ *
+ * Using unveil is irreversible. On OpenBSD, the first call immediately enforces
+ * the filesystem visibilty, and existing paths can only be updated with equal
+ * or lesser permissions. Filesystem operations that try to access invisible
+ * paths will raise ENOENT, and operations without the correct permissions raise
+ * EACCES. Unveiling can be disabled by either passing two NULL arguments or by
+ * calling pledge() without the "unveil" promise.
```
jart, line 152: It might be worth mentioning that unveil() on Linux by itself does not restrict `open(O_PATH)`, `fstat()`, `stat()`, and in access to file system metadata it seems.

jart:
I'm loving the documentation. Going to press merge. Please feel free to keep sending changes. This is a really wonderful contribution.

libc/mem/unveil.c
The-King-of-Toasters: I'll reword this later, but it's what your tests show policy union: The most permissive rule takes for a path is used. This means that this idiom doesn't work like on OpenBSD:

```c
unveil("/path", "rwc");
unveil("/path", "r");
// ...
open("/path", O_WRONLY|O_CREAT); // wut
```

libc/mem/unveil.c
The-King-of-Toasters: Yes, unfortunately. See [this warning](https://docs.kernel.org/userspace-api/landlock.html#filesystem-flags) in the kernel docs:

> It is currently not possible to restrict some file-related actions accessible through these syscall families: chdir(2), truncate(2), stat(2), flock(2), chmod(2), chown(2), setxattr(2), utime(2), ioctl(2), fcntl(2), access(2). Future Landlock evolutions will enable to restrict them.

jart:
Leaving a comment here to acknowledge your last round of responses. I think Landlock is pretty good even though it has caveats. It's been made clear that this is still `[wip]` so send me the doc updates at your leisure.

As for the limit of 16 that you were talking about earlier, I solicited some help from the guy who made Landlock on Twitter and he gave very similar advice. I may try writing some additional tests just to fiddle around with it and make sure I grok what's going on.

You may also noticed that I added a `-v [PERM:]PATH` flag to our pledge.com tool. I also addressed some other feedback we got on Hacker News, when it comes to things like `execpromises`.

To the best of your knowledge, is this the first time someone has used Landlock to polyfill unveil()? Because if we're the first, then we must announce it on the blog to let everyone know. So as soon as we feel it's ready, I intend to do a quick write-up and update justine.lol.

The-King-of-Toasters:
> Leaving a comment here to acknowledge your last round of responses. I think Landlock is pretty good even though it has caveats. It's been made clear that this is still [wip] so send me the doc updates at your leisure.

Sounds good, I should have some improvements ready by tomorrow. I think the docs should list the specific syscalls that are enabled, since they're not a 1:1 match with the pledge promises (e.g., you can unveil a path and still make a fifo/character device).

If you recall, I mentioned that one can re-pledge without the "unveil" promise to disable it. Right now the seccomp filter just disables the syscalls, but openbsd also does the equivalent of `unveil(NULL, NULL)` on the process ([source](https://github.com/openbsd/src/blob/73d6dc34cbeaee5f26156a444d405c98ba5510d1/sys/kern/kern_pledge.c#L502,L526)). We should do the same.

> To the best of your knowledge, is this the first time someone has used Landlock to polyfill unveil()? Because if we're the first, then we must announce it on the blog to let everyone know. So as soon as we feel it's ready, I intend to do a quick write-up and update justine.lol.

Sounds good. Though we weren't the first to make a wrapper (a quick search pulled [this](https://github.com/marty1885/landlock-unveil) up), I think ours is the most robust. I was actually inspired by the landlock usage in [gmid](https://github.com/omar-polo/gmid/blob/39ad4b8d7fc18e6c478dd41a4e5e6939842c66e8/sandbox.c#L435)

jart:
> Though we weren't the first to make a wrapper (a quick search pulled [this](https://github.com/marty1885/landlock-unveil) up)

That doesn't count. They use a joke license (WTFPL) and the author says "DO NOT" use this. Since that's the only project on GitHub when I search for "unveil landlock" I strongly suspect we are the first :-)

> Sounds good, I should have some improvements ready by tomorrow. I think the docs should list the specific syscalls that are enabled, since they're not a 1:1 match with the pledge promises

I thought Landlock didn't filter system calls. What I would probably say is, "While unveil() restricts which file system paths can be read, written, executed, and created, it does not restrict which file system operations are possible. For example, mknod() can still be called to create special files like sockets. It's recommended that unveil() be used in conjunction with pledge() to set a more comprehensive policy."

> If you recall, I mentioned that one can re-pledge without the "unveil" promise to disable it. Right now the seccomp filter just disables the syscalls, but openbsd also does the equivalent of unveil(NULL, NULL) on the process ([source](https://github.com/openbsd/src/blob/73d6dc34cbeaee5f26156a444d405c98ba5510d1/sys/kern/kern_pledge.c#L502,L526)). We should do the same.

Good to know. I'm going to do an experiment to see if we can install a blacklist filter independently of pledge().

PR #495, opened 2022-07-19
Add redbean OnLogLatency hook.

pkulchenko:
@jart, this works as discussed, but I'm not totally happy about the `OnLogLatency` hook name. If we're going to add more log-related events, then it may be fine, but if not, then something like `OnHttpRequestDone` may be better, but it's too long, and the request is not really done if it's being streamed to the client. I don't have any better ideas, but am open to suggestions if you prefer to rename it.

### DIFF
--- a/tool/net/help.txt
+++ b/tool/net/help.txt
@@ -534,6 +534,12 @@ HOOKS
           each time redbean accepts a new client connection. If it returns
           `true`, redbean will close the connection without calling fork.
 
+  OnLogLatency(reqtimeus:int,contimeus:int)
+          If this function is defined it'll be called from the main process
+          each time redbean completes handling of a request, but before the
+          response is sent. The handler received the time (in µs) since the
+          request handling and connection handling started.
+
   OnProcessCreate(pid:int,ip:int,port:int,serverip:int,serverport:int)
           If this function is defined it'll be called from the main process
           each time redbean forks a connection handler worker process. The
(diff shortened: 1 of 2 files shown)


tool/net/help.txt
```diff
@@ -534,6 +534,12 @@ HOOKS
           each time redbean accepts a new client connection. If it returns
           `true`, redbean will close the connection without calling fork.
 
+  OnLogLatency(reqtimems:int,contimems:int)
```
jart, line 537: Milliseconds?! Redbean is at least an order of magnitude faster than that. You need at least microsecond precision because static asset serving goes 10x faster for me.

```
(stat) "/tool/net/demo/index.html" latency r: 108µs c: 242µs
```

One thing you could do is hand over these numbers as floating point in seconds. Since it's a time interval (rather than an absolute time) there will be plenty of precision in a double.

jart, line 537: If you're concerned about extending this API, then see third_party/lua/lunix.c. You can use the source code for the unix.Stat object where you have an extendable struct store the fields. Then you'd just call the method to get the latency number you want, rather than having to worry about adding more arguments. This technique is super fast.

jart (changes requested):
Looks pretty good so far.

tool/net/help.txt
pkulchenko: Right, thank you for catching this typo; it was supposed to be us, so I updated the documentation.

pkulchenko: No, I'm not concerned about the API, just thinking about the name. If you're good with that, then I'm good too.

jart:
Build failure in MODE=rel unrelated. Proceeding with merge.

PR #499, opened 2022-07-21
Update IsAcceptablePath to allow .well-known/ URLs

pkulchenko:
This will still reject `/.` fragments after `/.well-known/`, but I think this is correct, as none should be expected according to [registered URIs](https://en.wikipedia.org/wiki/Well-known_URI).

### DIFF
--- a/net/http/isacceptablepath.c
+++ b/net/http/isacceptablepath.c
@@ -26,6 +26,7 @@
  *
  * 1. The substring "//" is disallowed.
  * 2. We won't serve hidden files (segment starts with '.').
+ *    The only exception is `/.well-known/`.
  * 3. We won't serve paths with segments equal to "." or "..".
  *
  * It is assumed that the URI parser already took care of percent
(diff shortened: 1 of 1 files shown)


net/http/isacceptablepath.c
```diff
@@ -66,7 +67,8 @@ bool IsAcceptablePath(const char *data, size_t size) {
       x = '/';
     }
     if (y == '/') {
-      if (x == '.') return false;
+      if (x == '.' &&  // allow /.well-known/ in the first position
+         (p - data > 2 || !startswith(data, "/.well-known/"))) return false;
```
jart, line 71: You can't assume the string is nul-terminated here. Could you please check the full length?

jart (changes requested):

net/http/isacceptablepath.c
pkulchenko: Correct; will do.

jart (approved):
Looks good. I'd also encourage you to send a change adding tests for this to `test/net/http/isacceptablepath.c`

PR #500, opened 2022-07-22
Correct typos: Round 2

shmup:
Here's another swoop through the typos.


pkulchenko:
I actually wouldn't worry about fixing typos in third_party folders, as those need to be reported upstream and will complicated any future updates/upgrades for those projects.

jart (approved):

jart:
Build failure doesn't seem related. I don't know why this is only happening to your PRs. Since master branch is fine. If this is actually a build break then I'll fix it locally.

PR #502, opened 2022-07-22
Add redbean GetResponseBody

pkulchenko:
@jart, this also includes a small fix for `GetStatus` that previously only worked for statuses sent from Lua code, but not for redbean-set statuses.


jart (approved):
Thank you!

PR #506, opened 2022-07-23
Redbean shutdown upd

pkulchenko:
@jart, here are the shutdown and daemonize changes we discussed.

I decided  not to remove the pid file, as it may be reused by the new copy of the executable (updated with a new pid), so the removal will be breaking the handling. It's not ideal, but I didn't want to add checks for timestamp or content before removing it.


jart (approved):
Looks good!

PR #520, opened 2022-08-01
Add max worker processing to redbean

pkulchenko:
@jart, this adds ability to set a limit for the number of workers. It's difficult to come up with a scenario where this beats the default configuration (performance-wise), but the main goal is to allow the number of workers to be limited without going into the meltdown mode when the system runs out of memory. I allowed it to be changed at any time, so it can be called from any server hook. It may be possible to combine this setting with `ProgramUniprocess`, but I thought it may be cleaner to keep them separate (although maybe `ProgramMaxWorkers(1)` should be the same as `ProgramUniprocess(true)`?). I also didn't add any command line options, as I don't think it may be worth setting from the command line (and it is possible to do with a Lua call anyway).

### DIFF
--- a/tool/net/redbean.c
+++ b/tool/net/redbean.c
@@ -411,6 +411,7 @@ static int sandboxed;
 static int changeuid;
 static int changegid;
 static int isyielding;
+static int maxworkers;
 static int statuscode;
 static int shutdownsig;
 static int sslpskindex;
@@ -4747,6 +4748,15 @@ static int LuaProgramHeartbeatInterval(lua_State *L) {
   return 1;
 }
 
+static int LuaProgramMaxWorkers(lua_State *L) {
+  if (!lua_isinteger(L, 1) && !lua_isnoneornil(L, 1)) {
+    return luaL_argerror(L, 1, "invalid number of workers; integer expected");
+  }
+  lua_pushinteger(L, maxworkers);
+  if (lua_isinteger(L, 1)) maxworkers = lua_tointeger(L, 1);
+  return 1;
+}
+
 static int LuaProgramAddr(lua_State *L) {
   uint32_t ip;
   OnlyCallFromInitLua(L, "ProgramAddr");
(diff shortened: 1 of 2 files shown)


tool/net/redbean.c
```diff
@@ -6912,6 +6924,7 @@ static int HandlePoll(int ms) {
         if (!polls[pollid].revents) continue;
         if (polls[pollid].fd < 0) continue;
         if (polls[pollid].fd) {
+          if (maxworkers && shared->workers >= maxworkers) continue;
```
jart, line 6927: Shouldn't you accept the connection and then send like a 503 Service Unavailable?

jart (changes requested):

tool/net/redbean.c
pkulchenko: I don't think it's needed, as a worker may become available. It's not much different from how uniprocess works, just for `maxworkers` instead of 1.

jart: Well if the event isn't handled, then won't poll spin and consume all the cpu?

pkulchenko: It shouldn't, as the effect is going to be the same as if the event wasn't there in the first place.

Think about it this way: if 5 requests come in and we are running redbean with 1 worker (uniprocess=true), the requests will be queued and handled one by one. If we're running redbean with 4 workers (maxworkers=4), only one requests will be queued and handled when one of the workers becomes available or is finished (to allow a new worker to be launched).

Since `poll` is not going to block (as there is an event pending), `HandlePoll` will exit in this case and the rest of the processing will be handled normally (inside `EventLoop`), including `ReapZombies` call, which includes handling of exited workers (`HandleWorkerExit`). Then the loop will repeat and if the number of workers goes down, a new worker will be launched.

Or am I missing something?

jart: I'm going to merge it and see if I can confirm what I suspect will happen actually happens.

jart:
Could you resolve conflicts and I'll merge?

jart (approved):

pkulchenko:
> I'm going to merge it and see if I can confirm what I suspect will happen actually happens.

Sounds good; I'm curious about your suspicions. Thank you for checking on this!

> Could you resolve conflicts and I'll merge?

Rebased against the current master. Should be all clean now.

PR #522, opened 2022-08-03
Add OnServerHeartbeat to Redbean

pkulchenko:
This adds `ProgramHeartbeatInterval` and `OnServerHeartbeat`. I've added disabling `/.heartbeat.lua` if the hook is set, but I now wonder if we should replace `.heartbeat.lua` with `OnServerHeartbeat`, as the former doesn't seem to be even documented. `OnServerHeartbeat` provides a bit more flexibility, as it can be added/removed dynamically.

### DIFF
--- a/tool/net/help.txt
+++ b/tool/net/help.txt
@@ -1368,6 +1368,10 @@ FUNCTIONS
           as Content-Range and Date, which are abstracted by the transport
           layer.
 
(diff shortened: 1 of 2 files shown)


jart (approved):

jart:
We can deprecate the heartbeat file if you want.

PR #525, opened 2022-08-05
Change types of extern const declarations to match other platforms

Jules-Bertholet:
There are a few consts I could not find the proper types for:

- `libc/sysv/consts/nd.h`
- `libc/sysv/consts/read.h`
- some of `libc/sysv/consts/termios.h`
- `libc/sysv/consts/th.h`


jart (approved):
Wow, thanks for doing this!

PR #535, opened 2022-08-14
Improve detection of boot media on bare metal

tkchia:
`hello.com` now starts up correctly on QEMU when run as a hard disk image.


jart (approved):
Wow nice! Thank you!

PR #539, opened 2022-08-16
Stop sys_lseek from truncating return value to 32 bits

### LINKED ISSUE #538: sys_lseek truncates return values to 32 bits
`libc/calls/lseek.c` takes an `int64_t offset` returns an `int64_t` and on non sysv platforms calls `sys_lseek_nt` which also returns an int64_t.

All cosmo platforms should have a 64 bit implementation. The current behavior on linux is to silently truncate the return value from the syscall.

G4Vi:
Fixes #538 . Only tested on Linux.


jart (approved):
Oh wow, great find!

PR #540, opened 2022-08-17
Create definitions.lua

stellartux:
This PR adds a definition file which allows the use of the Lua language server with redbean projects.

### DIFF
--- a/tool/net/definitions.lua
+++ b/tool/net/definitions.lua
@@ -1,0 +1,6586 @@
+---@meta
+error("Tried to evaluate definition file.")
+--Documents redbean 2.0.16, lsqlite3 0.9.4
+
+--[[
+
+SYNOPSIS
+
+  redbean.com [-?BVabdfghjkmsuvz] [-p PORT] [-D DIR] [-- SCRIPTARGS...]
+
+DESCRIPTION
+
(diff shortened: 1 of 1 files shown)


shmup:
Nice. Can you give us an example of this usage? I'm using a lua language server and probably not using it to the best of its ability. Any info you have would be great!

jart (approved):
Looks good. Could you help us write a tool that turns it back into .txt so we can reduce duplication?

shmup:
@stellartux this worked wonderfully, thanks for the explanation. In case my comment shows the way for others, I'm going to elaborate.

In vim I'm using [coc.nvim](https://github.com/neoclide/coc.nvim) configured with [josa42/coc-lua](https://github.com/josa42/coc-lua) which technically uses [sumneko/lua-language-server](https://github.com/sumneko/lua-language-server).

In my card project I've simply [added this file](https://github.com/shmup/redbean-cardgames/commit/ed791cd0f9a5a984bf82af186030d6ff9f725a9f) to my project root in `/definitions` and _it all just works_. Fantastic!

shmup:
> Looks good. Could you help us write a tool that turns it back into .txt so we can reduce duplication?

Where would a tool like this exist in the project?

Issue #541, opened 2022-08-17
putenv/setenv is unusable after clearenv / NULL pointer dereference when PutEnvImpl is called after clearenv


G4Vi:
`clearenv` sets `environ` to NULL, Neither `putenv` nor `setenv` verify `environ` is not NULL, leading to a NULL pointer dereference when either is called to set the environment.

G4Vi:
Fixed by #542

PR #542, opened 2022-08-17
Add reallocating environ if it's set NULL / Fix NULL pointer dereference when PutEnvImpl is called after clearenv

G4Vi:
Potential fix for #541

Not sure if something different should be done instead of letting the `__cxa_atexit(FreeEnviron` callbacks stack up.

### DIFF
--- a/libc/mem/putenv.c
+++ b/libc/mem/putenv.c
@@ -16,10 +16,10 @@
 │ TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR             │
 │ PERFORMANCE OF THIS SOFTWARE.                                                │
 ╚─────────────────────────────────────────────────────────────────────────────*/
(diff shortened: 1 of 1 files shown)


jart (approved):
Thank you for sending this!

I'd encourage you to write tests too when encountering issues like this. I'm a big fan of test driven development so the testing tools are pretty high quality to reflect that.

For example, with something like this you'd want to put the following code in `test/libc/mem/putenv_test.c`

```c
#include "libc/mem/mem.h"
#include "libc/runtime/runtime.h"
#include "libc/testlib/testlib.h"

TEST(putenv, test) {
  EXPECT_EQ(0, clearenv());
  EXPECT_EQ(0, putenv("hi=there"));
  EXPECT_STREQ("there", getenv("hi"));
}
```

Then you'd run `make MODE=dbg -j16 o/dbg/test/libc/mem`

PR #544, opened 2022-08-17
Fix stdio fmt of "%.0e" and "%.0g"

### LINKED ISSUE #543: stdio fmt issues with "%0.e" and "%.0g"
```c
#include <stdio.h>
#include <string.h>

typedef struct {
    const char *fmt;
    double num;
    const char *expected;
} Test;

static const Test Tests[] = {
    {"%.0e", 1.875, "2e+00"},
    {"%.0e", 0.875, "9e-01"},
    {"%.0g", 12345.6789, "1e+04"},
    {"%#.0g", 12345.6789, "1.e+04"}
};

int main(void) {
    for(unsigned i = 0; i < (sizeof(Tests) / sizeof(Tests[0])); i++)
    {
        char buf[24];
        snprintf(buf, sizeof(buf), Tests[i].fmt, Tests[i].num);
        fwrite((strcmp(buf, Tests[i].expected) == 0)  ? "PASS" : "FAIL", 4, 1, stdout);
        fwrite(" got ", 5, 1, stdout);
        fwrite(buf, strlen(buf), 1, stdout);
        fwrite(" expected ", 11, 1, stdout);
        fwrite(Tests[i].expected, strlen(Tests[i].expected), 1, stdout);
        fwrite("\n", 1, 1, stdout);
    }
}
```

glibc prints:
```
PASS got 2e+00 expected 2e+00
PASS got 9e-01 expected 9e-01
PASS got 1e+04 expected 1e+04
PASS got 1.e+04 expected 1.e+04
```

cosmo prints
```
FAIL got 1e+00 expected 2e+00
FAIL got 8e-01 expected 9e-01
FAIL got 12345.6789 expected 1e+04
FAIL got 12345.6789 expected 1.e+04
```

perl versions of these tests were found in perl's `t/op/sprintf.t`

G4Vi:
Fixes #543

`"%.0e"` was fixed by always using dtoa mode 2: max(1,ndigits) significant digits so the number will still be rounded even if no decimal places will be printed, the change should only change behavior when precision is 0.

The fix to `"%.0g"` is similar, according to `man 3 printf`  "if the precision is zero, it is treated as 1", so perc is now rounded up to 1 instead of 0 if necessary. This also leads to always using dtoa mode 2.

EDIT: Changing this to WIP as "%.0g", 10 and "%.0g", -10. appears to be failing.
EDIT2: Not sure if the behavior required by tests is desired, see next comment.


G4Vi:
Is there a specific spec of printf cosmo is trying to follow? The failing tests below  from fmt_test.c in `{fmt, input, expected}` form (do ensure that "%.0g" prints the shortest representation of 10 and -10, however it differs from glibc and my interpretation of the linux `man 3 printf`
```
{"%.0g", 10., "10"},
{"%.0g", -10., "-10"},
```
glibc prints
```
1e+01
-1e+01
```
for those tests.

My interpretation of the `g, G` conversion specifier in psuedo code is as follows:
```
if(precision == undef)
{
    precision = 6;
}
else if(precision == 0)
{
    precision = 1;
}
if((exponent_from_conversion < -4) || (exponent_from_conversion >= precision))
{
    usestylee();
}
else
{
    usestylef();
}
```

The "%g" patch currently included causes 10 and -10 to match the glibc behavior of encoding them as `1e+01` and `-1e+01`.

jart:
> Is there a specific spec of printf cosmo is trying to follow?

Glibc is the gold standard of C libraries. Anything that helps us be closer to Glibc in terms of behavior would be very welcome.

If we have tests that specify a behavior that's inconsistent with Glibc, then our tests are wrong. Many of them were borrowed from Paland's printf() implementation which was intended to be a simple solution for embedded systems. We've modified our implementation a lot since then to bring ourselves into closer conformance. For instance, a lot of the double formatting code in our printf() implementation was borrowed from the dtoa authors.

Since you're already a contributor, please ping me when this is ready to merge. Then I'll approve it.

G4Vi:
@jart Thanks for the quick reviews. I fixed the %g tests to match glibc and added more. LGTM.

jart (approved):
Thank you!

PR #545, opened 2022-08-18
add basic putenv/setenv test

G4Vi:


jart (approved):
Thanks for sending a test!

PR #546, opened 2022-08-18
setlocale: Add fake support for locale="")

G4Vi:
`setlocale( int category, "");` is supposed to set the category or all categories in the case of the category `LC_ALL` to the values specified in the environment.

Actually querying the environment would be unlikely to be helpful as we only support the C or POSIX locales, so why not fake it. 

Some programs such as perl call `setlocale( LC_ALL, "");`, at startup expecting it to be able to set the locale from the environment and prints an error if it fails.

We could instead of this patch, actually setup the environment with `LC_*="C"` and change `setlocale` to query it, but so far this seems to work and is simpler.


jart (approved):
Thank you!

PR #549, opened 2022-08-18
add libc/str/nltypes.h to langinfo.h to match glibc

G4Vi:
glibc `langinfo.h` includes `nl_types.h`.

In the wild perl counts on this as it doesn't include `nl_types.h` anywhere.


jart (approved):
Thanks!

Issue #455, opened 2022-06-28
Fix signal handling under Rosetta


dkulchenko:
I'll preemptively mark this one as community-supported, but wanted to seek input:

Ctrl-C or Ctrl-D to quit Redbean doesn't work correctly under Rosetta. On non-assimilated binaries (running under ape), issuing Ctrl-C or Ctrl-D hangs the process and creates a rapid memory leak (used 64GB of RAM in about 10 seconds) which doesn't stop until I `killall -9 .ape`. On assimilated binaries, there is no memory leak and we see this instead:

```
I2022-06-27T17:40:54.089084:tool/net/redbean.c:6930:redbean:89097] (srvr) listen http://127.0.0.1:8080
I2022-06-27T17:40:54+000928:tool/net/redbean.c:6930:redbean:89097] (srvr) listen http://192.168.1.187:8080
I2022-06-27T17:40:54+000026:tool/net/redbean.c:6930:redbean:89097] (srvr) listen http://10.147.19.71:8080
>:
rosetta error: unsupported privilege level: 0
[1]    89097 trace trap  ./redbean.com
❯
```

(that's for a Ctrl-C). And this for a Ctrl-D:

```
I2022-06-27T17:41:16.856741:tool/net/redbean.c:6930:redbean:89110] (srvr) listen http://127.0.0.1:8080
I2022-06-27T17:41:16+000815:tool/net/redbean.c:6930:redbean:89110] (srvr) listen http://192.168.1.187:8080
I2022-06-27T17:41:16+000020:tool/net/redbean.c:6930:redbean:89110] (srvr) listen http://10.147.19.71:8080
>:
I2022-06-27T17:41:17.399988:tool/net/redbean.c:6952:redbean:89110] (srvr) received SIGHUP
I2022-06-27T17:41:17+000020:tool/net/redbean.c:6955:redbean:89110] (srvr) killing process group
rosetta error: unsupported privilege level: 0
[1]    89110 trace trap  ./redbean.com
❯
```

I've tracked the issue down to signal handling - if I comment out `SigInit();` in redbean.c, there are no errors or leaks on both assimilated and non-assimilated binaries. Here's a Ctrl-D:

```
I2022-06-27T17:42:53.387199:tool/net/redbean.c:6930:redbean:89253] (srvr) listen http://127.0.0.1:8080
I2022-06-27T17:42:53+000892:tool/net/redbean.c:6930:redbean:89253] (srvr) listen http://192.168.1.187:8080
[...]

PR #552, opened 2022-08-19
Work around Rosetta signal handling issue (redbean variant)

dkulchenko:
Rosetta does something strange to the signal handling registers but setting SA_SIGINFO prevents the issue from happening. Detect Rosetta and set the flag when necessary.

Tested on both M1 and non-M1 and things are fully working on both.

See https://github.com/jart/cosmopolitan/issues/455 for discussion.

@jart This patch is for Redbean only, pull request #553 applies the change to sigaction.c instead to apply globally for  cosmopolitan in general.

### DIFF
--- a/libc/calls/calls.h
+++ b/libc/calls/calls.h
@@ -182,6 +182,7 @@ int symlinkat(const char *, int, const char *);
 int sync_file_range(int, int64_t, int64_t, unsigned);
 int sys_ptrace(int, ...);
 int sysctl(const int *, unsigned, void *, size_t *, void *, size_t);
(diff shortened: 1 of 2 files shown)


jart (changes requested):
Wow. Nice Rosetta detecting function. We'll likely need it in the future. Here's what I propose. Could we modify sigaction() to *always* use `SA_SIGINFO` when `IsXnu()`? That'll be fine with x86-64 architecture.

dkulchenko:
Sure thing! Would you prefer I do that in redbean (this pull request) or in sigaction itself (pull #553)?

jart:
The other one please. It'll help avoid a system call and some code size.

PR #553, opened 2022-08-19
Work around Rosetta signal handling issue (sigaction variant)

dkulchenko:
This is a twin pull request to #552, but with the changes applied to sigaction globally instead of just to redbean. @jart Feel free to pick your favorite and close the other :)

Rosetta does something strange to the signal handling registers but setting SA_SIGINFO prevents the issue from happening. Detect Rosetta and set the flag when necessary.

Tested on both M1 and non-M1 and things are fully working on both.

See https://github.com/jart/cosmopolitan/issues/455 for discussion.

### DIFF
--- a/libc/calls/sigaction.c
+++ b/libc/calls/sigaction.c
@@ -183,6 +183,10 @@ static int __sigaction(int sig, const struct sigaction *act,
       if (IsXnu()) {
         ap->sa_restorer = (void *)&__sigenter_xnu;
         ap->sa_handler = (void *)&__sigenter_xnu;
+
+        // mitigate Rosetta signal handling strangeness
+        // https://github.com/jart/cosmopolitan/issues/455
+        ap->sa_flags |= SA_SIGINFO;
       } else if (IsLinux()) {
         if (!(ap->sa_flags & SA_RESTORER)) {
           ap->sa_flags |= SA_RESTORER;


libc/calls/sigaction.c
```diff
@@ -449,6 +449,14 @@ static int __sigaction(int sig, const struct sigaction *act,
  * @vforksafe
  */
 int sigaction(int sig, const struct sigaction *act, struct sigaction *oldact) {
+  if(act && IsXnu()) {
+    // mitigate Rosetta signal handling strangeness
+    // https://github.com/jart/cosmopolitan/issues/455
+    struct sigaction *actrosetta = act;
+    actrosetta->sa_flags = SA_SIGINFO;
```
jart, line 456: `|=`

jart (changes requested):

libc/calls/sigaction.c
jart: Also please create a copy of the data structure, since `act` is const.

dkulchenko: I don't quite know C well enough to know how to do that :/ I tried:

```c
struct sigaction actrosetta = *act;
actrosetta.sa_flags |= SA_SIGINFO;
act = &actrosetta;
```

and 

```c
struct sigaction actrosetta;
memcpy(&act, &actrosetta, sizeof(act));
actrosetta.sa_flags |= SA_SIGINFO;
act = &actrosetta;
```

but both error out with `error:tool/net/redbean.c:7179: check failed: -1 != -1 EFAULT/14/Bad address`.

jart: Try this:

```diff
diff --git a/libc/calls/sigaction.c b/libc/calls/sigaction.c
index e40179d45..de6c83f50 100644
--- a/libc/calls/sigaction.c
+++ b/libc/calls/sigaction.c
@@ -181,6 +181,7 @@ static int __sigaction(int sig, const struct sigaction *act,
       memcpy(&copy, act, sizeof(copy));
       ap = &copy;
       if (IsXnu()) {
+        ap->sa_flags |= SA_SIGINFO;
         ap->sa_restorer = (void *)&__sigenter_xnu;
         ap->sa_handler = (void *)&__sigenter_xnu;
       } else if (IsLinux()) {
```

jart: As for your examples, the first one you wrote looks like it should work. Unless you nested in curlies and it went out of scope. The second example, the parameter order is incorrect for memcpy(dest, src, len).

dkulchenko: Perfect - that'll do it!

dkulchenko: Force-pushed and tested on M1 and non-M1.

jart (approved):

jart:
Thank you!

PR #561, opened 2022-08-22
Update redbean GetResponseBody to take response compression into account

pkulchenko:
@jart, this fixes an issue with GetResponseBody for those responses that have been compressed in the buffer. This returns their uncompressed versions (as the compressed content is not very useful, especially without knowing the original size). I used `cpm.gzipped` value to track the original size or mark a special case when the Deflate generator is used. I tested with normal, gzipped, and streamed content.

Also, as I was working through the code, I noticed that `cpm.gzipped` is not reset in `SetStatus`, but some other flags from `cpm` are. I wonder if we should check further on that.

### DIFF
--- a/tool/net/help.txt
+++ b/tool/net/help.txt
@@ -1067,9 +1067,13 @@ FUNCTIONS
           IsPrivateIp or IsLoopbackIp return true. When multiple addresses
           are present in the header, the last/right-most address is used.
 
-  GetResponseBody() → str
-          Returns the response message body if present or an empty string.
-          Also returns an empty string during streaming.
+  GetResponseBody()
+      ├─→ body:str
+      └─→ nil, error:str
+          Returns the (uncompressed) response message body if present or an
+          empty string. May also return a partial or empty string during
+          streaming, as the full content may not be known at the call time.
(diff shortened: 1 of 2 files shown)


jart:
Could you fix this build error?

```
tool/net/redbean.c:4001:3: error: 's' may be used uninitialized in this function [-Werror=maybe-uninitialized]
 4001 |   lua_pushlstring(L, cpm.gzipped > 0 ? s  // return decompressed
      |   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 4002 |                    : cpm.gzipped == 0 ? cpm.content : "", size);
      |                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

pkulchenko:
I sure can, but what's the best way to fix it (short of setting the default value to an empty string)? Logically it's going to be initialized. Interesting that it actually compiled cleanly for me.

pkulchenko:
Updated; I don't get any build errors or warnings.

jart (approved):

PR #566, opened 2022-08-23
Add MapContentType to redbean to update extension-to-contenttype map

pkulchenko:
@jart, I ended up implementing `MapContentType` (that does both set and get), as I thought that `Get/SetContentType` may be too confusing.

Couple of things to point out:
1. The default mapping is checked first (for performance reasons), so it's not possible to overwrite default values, but I don't think it's really needed.
2. The set mapping can't be "unset", but I don't think it's needed either.
3. The dot '.' processing is needed because FindContentType expects a path (with a dot), but manually set extensions don't (and shouldn't) have dots.
4. The code ended up being a bit longer than I expected, but it's mostly because I'm storing the assigned values in a Lua table, but need to access it from both Lua and C sides. I've tested all the paths and everything seems to be working correctly, but would love your review, as I'm sure it can be further improved.

### DIFF
--- a/test/tool/net/lua_test.lua
+++ b/test/tool/net/lua_test.lua
@@ -21,3 +21,12 @@ assert(0200 == 128)
 assert("\e" == "\x1b")
 assert("hi" * 3 == "hihihi")
 assert("hello %d" % {123} == "hello 123")
+
+-- test MapContentType
+assert(MapContentType("txt") == "text/plain")
+assert(MapContentType("htm") == "text/html")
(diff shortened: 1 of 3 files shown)


jart (changes requested):
Could you write Lua unit tests for this in `test/tool/net/`? That would help better explain how it works.

pkulchenko:
Added tests and fix calling it from REPL/CLI.

jart (approved):

PR #567, opened 2022-08-23
prevent pydoc's help from showing extra junk

ahgamut:
when calling `help` pydoc decides what to show based upon a bunch of hasattr calls

- hasattr(obj, "attribute") is True even if the attribute is None,
-> modules loaded via CosmoImporter have hasattr(module, "__path__") as True
-> pydoc thinks of them as containin subpackages
-> (even though "__path__" is NONE)
-> so it prints a bunch of submodule junk

now that doesn't happen, so pydoc does the expected thing


jart (approved):
Nice! Thank you.

ahgamut:
before PR:

image

after PR (conforms to usual Py36 `help` output):

image

ahgamut:
@jart the headings in the `help` output are found within `pydoc.py`, so further customization can be don there

Issue #565, opened 2022-08-22
warning: allocated section `.tdata' not in segment


andy5995:
* Manjaro 21.3.6
* gcc (GCC) 12.1.1

I get these warning when  using the command line shown on the README:

```
/usr/bin/ld.bfd: hello.com.dbg: warning: allocated section `.tdata' not in segment
/usr/bin/ld.bfd: warning: .note.gnu.build-id section discarded, --build-id ignored
```
Just thought I'd report it in case it's something you'd like to address. No problem for me. ;)

```
gcc -g -Os -static -nostdlib -nostdinc -fno-pie -no-pie -mno-red-zone \
  -fno-omit-frame-pointer -pg -mnop-mcount -mno-tls-direct-seg-refs \
  -o hello.com.dbg hello.c -fuse-ld=bfd -Wl,-T,ape.lds -Wl,--gc-sections \
  -include cosmopolitan.h crt.o ape-no-modify-self.o cosmopolitan.a
```

/cc @jammyjamjamman

jart:
I wish I could address it. It's been happening ever since we introduced `_Thread_local` support. I've tried to make it go away multiple times but haven't been able to figure out what's causing it. It appears to be harmless.

If anyone who reads this is a linker script expert who can help us solve this, please see `ape/ape.lds`.

tkchia:
Hello @jart, hello @andy5995,

I had lately been messing around with TLS support [for a separate project of mine](https://gitlab.com/tkchia/solidate/-/blob/99ad0449d64a155f88e70fcdf3029fcb0f199fe7/solidate.ld).  I found that I had to do a few things:

  * The `.tdata` and `.tbss` sections are _not_ meant to be the actual TLS area for the first thread, but rather as _initializers_ to be used to prime the TLS contents of each thread, including the first thread.  So the `.tdata` contents need to be copied out to some other place at run time.  (Think of `.tdata` as an area of ROM, which you copy out to RAM, in order to initialize the RAM.)
  * The `.tdata` section should be assigned to _two_ program segments: a `PT_LOAD` segment, and a `PT_TLS` segment.  (Since `.tdata` 's contents are read-only, it can go into a read-only `PT_LOAD` segment.)
  * The `.tbss` section only needs to be assigned to a `PT_TLS` segment.

[...]

PR #571, opened 2022-08-25
Fix "allocated section `.tdata' not in segment" warning (#565) and some related issues

tkchia:
* Need to place `.tdata` inside _two_ PHDRS segments
* Remove (IMO) ugly alignment padding between `.tdata` and `.tbss`
* Treat `.tdata` as part of the program's data segments: place it between `_etext` and `_edata`
* ~~Tweak calculation of `v_ape_realsectors`, for bare metal (padding the image size to page size multiple caused disk read errors in QEMU)~~ (**edit:** I hope to fix this bit later)
* Zap now-unused `v_ape_realpages` and `v_ape_highsectors` symbols

### DIFF
--- a/ape/ape.lds
+++ b/ape/ape.lds
@@ -376,10 +376,6 @@ SECTIONS {
     KEEP(*(.piro.pad.data))
     KEEP(*(.dataepilogue))
 /*END: NT FORK COPYING */
-    HIDDEN(_edata = .);
-    PROVIDE_HIDDEN(edata = .);
-    KEEP(*(SORT_BY_NAME(.zip.*)))
-    HIDDEN(_ezip = .);
   } :Ram
 
   .tdata . : {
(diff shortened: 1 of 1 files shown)


tkchia:
Hello @jart,

Regarding `_edata`, may I know your intended treatment of the content between `_edata` and `_ezip`?  E.g. is it meant to be part of the initial program image to be loaded (or mapped) into virtual memory?

I see that you included the area between `_edata` and `_ezip` in the calculation of `ape_ram_filesz`.  But it was not included in the calculation of `v_ape_realsectors` or `v_ape_highsectors`.

Thank you!

tkchia:
> Tweak calculation of `v_ape_realsectors`, for bare metal (padding the image size to page size multiple caused disk read errors in QEMU)

This seems to be mostly its own separate problem &mdash; I think I will try to tackle it with a separate pull request.

Thank you!

PR #574, opened 2022-08-26
Stop APE bare metal loader from reading beyond program image end

tkchia:
This allows e.g. `qemu-system-x86_64 -s o/examples/hello.com -serial stdio` to work without having to add extra padding to the end of the `hello.com` "disk image".

(The sector count computation is divided among two instructions in the assembly code.  This is done on purpose, to prevent an ASCII `0x27` (single quote) byte from appearing in the bare metal loader code, which will break the shell script loader.  There is probably a better way.)


jart (approved):
Looks good. Tested it on the fleet locally too just to confirm no breakages on other platforms.

PR #577, opened 2022-08-27
Fix "allocated section `.tdata' not in segment" warning (#565)

tkchia:


jart:
Thank you for sending this. Something about this change causes binaries to not run on Windows. I'm looking into what exactly is wrong. I'll report back when I figure it out.

jart:
Aha! It appears that, for some reason, the end of the binary wasn't being extended to the page alignment. Thank you so much for helping.  I'm thrilled we finally figured this out. I'm going to push a commit that makes a slight addition to this change in the linker script.

PR #579, opened 2022-08-29
Improve zipos directory support

G4Vi:
This patch allows the zipos to behave more like a normal filesystem by adding virtual DIR/dirent capability to zipos `stat`, `opendir`, and `readdir`.

`/zip/` is now treated as a directory, `struct DIR *dir = opendir('/zip/'); struct dirent *ent = readdir(dir);` Can now be used to enumerate the toplevel of the zip file.

In addition, directories that are not explicitly encoded in the zip, but are referenced (parent directories in a path to a directory or file) can be found with `opendir` and `readdir`. However, it is not foolproof; to inexpensively to avoid duplicates in `readdir`, the previous `struct dirent d_name` is compared against to determine if a directory has already been found.

Edit: Looking into that failing test.

### DIFF
--- a/libc/stdio/dirstream.c
+++ b/libc/stdio/dirstream.c
@@ -326,6 +326,7 @@ static struct dirent *readdir_impl(DIR *dir) {
   uint8_t *s, *p;
   struct Zipos *zip;
   struct dirent *ent;
+  struct dirent *lastent;
   struct dirent_bsd *bsd;
   struct dirent_netbsd *nbsd;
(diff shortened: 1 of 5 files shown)


G4Vi:
Fixed the failing test at least locally.

Using the offset of the zip central directory as the magic return value from  ` __zipos_find` wasn't safe. 0 should be okay, but we could change the api to more explicitly deal with the virtual items.

jart (approved):
Nice! Thank you!

PR #582, opened 2022-08-31
inet_ntop: Change AF_INET6 required buffer length from 72 to 46.

G4Vi:
I'm not sure why a minimum buffer length of 72 was required. The longest IPv6 address appears to be 46 with terminating zero: `FFFF:FFFF:FFFF:FFFF:FFFF:FFFF:255.255.255.255` and no tests appear to require anything bigger.
		
On Windows `INET6_ADDRSTRLEN` is 65 instead of 46 everywhere else to accommodate `[ffff:ffff:ffff:ffff:ffff:ffff:255.255.255.255%2147483648]:65535`, however Windows' `inet_ntop` only requires a buffer of 46 like unix. https://docs.microsoft.com/en-us/windows/win32/api/ws2tcpip/nf-ws2tcpip-inet_ntop

Some programs such as (perl `cpan/Socket`) expect `inet_ntop` to work with an `INET6_ADDRSTRLEN` sized buffer, so at a minimum it could be adjusted to use `INET6_ADDRSTRLEN` instead of 72. However, to be permissive and explicit of the size of string we produce (like with the `AF_INET` implementation), this PR sets it to 46.
		
I think the current `inet_ntop` implementation in cosmo only produces strings of length 39 (requiring buffers of length 40), but 46 is safer in-case we ever want to output the IPv4-mapped IPv6 format. The Linux man page says `inet_ntop` requires `INET6_ADDRSTRLEN` minimum sized buffers(46).


jart:
You appear correct on the length. We might want to use 48 to play it safe. I guess Linux defines it that way to accommodate the square brackets maybe?

```
linux   #define INET6_ADDRSTRLEN        (48)                    /home/jart/vendor/linux/include/linux/inet.h
xnu     #define INET6_ADDRSTRLEN        46                      /home/jart/vendor/xnu/bsd/netinet6/in6.h
freebsd #define INET6_ADDRSTRLEN        46                      /home/jart/vendor/freebsd/sys/netinet6/in6.h
openbsd #define INET6_ADDRSTRLEN        46                      /home/jart/vendor/openbsd/sys/netinet6/in6.h
netbsd  #define INET6_ADDRSTRLEN        46                      /home/jart/vendor/netbsd/sys/netinet6/in6.h
```

jart (approved):

jart:
I'm going to just merge since that concern shouldn't apply here.

PR #592, opened 2022-09-05
sigaction: Discard ignored signals on New Technology

G4Vi:
When a signal handler is set to `SIG_IGN` or `SIG_DFL` for signal with it's default behavior to ignore, the existing signal instances should be discarded even if they're blocked.

The new test added already passed on `sysv`, but failed on New Technology. This was discovered attempting to implement sigpending (WIP).

Thanks for all the recent reviews!


jart (approved):
Thank you!

PR #597, opened 2022-09-07
Implement sigpending for sysv and nt.

G4Vi:
`sigpending_test.com` ran successfully on all 3 BSDs, XNU, Linux, and NT.

The FreeBSD sigpending syscall number was updated to use the new version instead of compat.


jart (approved):
You did a really outstanding job implementing this polyfill. It shows how well you've studied the codebase. Thank you!

PR #599, opened 2022-09-07
add FP_ILOGB constants

G4Vi:


libc/math.h
```diff
@@ -64,6 +64,8 @@
 #define FP_ZERO      2
 #define FP_SUBNORMAL 3
 #define FP_NORMAL    4
+#define FP_ILOGB0    -2147483648
```
jart, line 67: `-2147483647-1` unless you want the type to be `long`

jart (changes requested):

libc/math.h
G4Vi: Thanks, I now understand why some implementations used that idiom. Now fixed.

jart (approved):

PR #600, opened 2022-09-07
Fix typos in CP437 → UTF-8/-16 mapping tables

tkchia:


tkchia:
On second thought, it might be better to have CP437 `0xfa` &rightarrow; Unicode `0x00b7` as a "secondary" multi-mapping instead.

PR #602, opened 2022-09-08
Refine `V=0` command logging

notwa:
This fixes off-by-one bugs, as well as missing carriage returns caused by command truncation. When the terminal's width is 0 or unknown, line feeds are used instead. Otherwise, the command is padded with spaces to clear the line when the terminal is dumb.

---

from a brief test, it looks like `GetTerminalWidth` returns 0 for GitHub workflows, so its logs *should* be the same, but with `\n` EOLs instead of `\r`. let me know if you have any concerns.

screenshots of the patched bugs:
image
(exaggerated)
image

### DIFF
--- a/tool/build/compile.c
+++ b/tool/build/compile.c
@@ -333,7 +333,6 @@ int GetTerminalWidth(void) {
   } else {
     ws.ws_col = 0;
     ioctl(2, TIOCGWINSZ, &ws);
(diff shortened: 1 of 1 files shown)


jart (approved):
Looks good. Thank you for sending this change. Also thank you for sending the copyright assignment.

PR #605, opened 2022-09-09
Add sigpending to redbean unix.* module

pkulchenko:
@G4Vi, @jart, this adds `sigpending` to unix.* module (from #597). Everything appears to work, but I'm not sure how to test for pending signals.

### DIFF
--- a/third_party/lua/lunix.c
+++ b/third_party/lua/lunix.c
@@ -1797,6 +1797,20 @@ static int LuaUnixSigsuspend(lua_State *L) {
   return LuaUnixSysretErrno(L, "sigsuspend", olderr);
 }
 
+// unix.sigpending()
+//     ├─→ mask:unix.Sigset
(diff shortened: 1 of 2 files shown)


G4Vi:
@pkulchenko Nice work. Seeing that `lunix.c` appears to have most of the signal calls, you could try porting the [libc sigpending test](https://github.com/jart/cosmopolitan/blob/master/test/libc/calls/sigpending_test.c) to lua. The calls to `signal()` can be replaced with the right incantation of `sigaction()`, (`signal()` is implemented via `sigaction` in the libc).

If you `raise()` after blocking a signal with `sigprocmask()` it will become pending and should be returned by `sigpending`. If you unblock a signal the handler should automatically be called then and thus the signal no longer pending. Ignoring a blocked signal (set handler to `SIG_IGN`), should also drop it from pending.

jart (approved):
Thank you!

PR #606, opened 2022-09-09
set stackaddr if mmap is called in pthread_create

ahgamut:
when calling pthread_create with a pthread_attr_t *attr,

if attr != NULL AND attr->stackaddr != 0,
then stack is not managed by cosmo
pt->spawn.stk = attr->stackaddr,
and so pt->attr.stackaddr = pt->spawn.stk is a no op

if attr == NULL OR attr->stackaddr == 0,
the stack is managed by cosmo,
pt->spawn.stk is from a successful mmap,
(and this is the oneline change)
so pt->attr.stackaddr SHOULD BE SET AS pt->spawn.stk


jart (approved):
Nice! Thank you! Why didn't I think of that?

PR #609, opened 2022-09-11
Get bare metal code (mostly) working outside of an emulator

tkchia:
These patches fix a few issues, which were causing the bare metal code to crash (and reboot) when run on a real machine, outside of an emulator.  These have to do with
  * the retrieval of the system memory map (`int 0x15`, `eax` = `0xe820`), and
  * the initial creation of the page tables for virtual memory operation.

I have debugged and tested the patched code, by booting it from a USB drive, on a x86-64 PC with a "legacy BIOS".

Thank you!

### DIFF
--- a/ape/ape.S
+++ b/ape/ape.S
@@ -1392,9 +1392,16 @@ a20:	cli
 
 //	Initializes long mode paging.
 pinit:	push	%ds
+	push	%es
 #define	SEG 0x79000
 	mov	$SEG>>4,%ax
 	mov	%ax,%ds
+	mov	%ax,%es
+	xor	%di,%di
+	xor	%ax,%ax
+	mov	$(0x7f000-SEG)/2,%cx
+	cld
+	rep stosw
 	movl	$0x7d000+PAGE_V+PAGE_RW,0x7e000-SEG	# PDPT←PML4T (+)
 	movl	$0x7c000+PAGE_V+PAGE_RW,0x7e800-SEG	# PDPT←PML4T (-)
 	movl	$0x7b000+PAGE_V+PAGE_RW,0x7d000-SEG	# PDT←PDPT   (+)
@@ -1403,13 +1410,14 @@ pinit:	push	%ds
 	movl	$0x79000+PAGE_V+PAGE_RW,0x7a000-SEG	# PD←PDT     (-)
 	mov	$512,%cx				# PD±2MB
 	mov	$PAGE_V+PAGE_RW,%eax
-	xor	%si,%si
-0:	mov	%eax,(%si)
+	xor	%di,%di
+0:	stosl
 	add	$0x1000,%eax
-	add	$8,%si
+	scasl						# di += 4
 	loop	0b
 	mov	$0x7e000,%eax				# PML4T←CR3
 	mov	%eax,%cr3
+	pop	%es
 	pop	%ds
 	ret
 	.endfn	pinit
--- a/ape/ape.S
+++ b/ape/ape.S
@@ -329,7 +329,7 @@ dsknfo:	push	%bx
 	movpp	%es,%ds
 	xor	%si,%si
 	mov	%si,%es
-	mov	$0x1510,%si		# mman::pc_drive_base_table
+	mov	$0x1d10,%si		# mman::pc_drive_base_table
 	xchg	%si,%di
 	movsw	#→ headunloadtime, headloadtime
 	movsw	#→ shutofftime, bytespersector
@@ -382,11 +382,11 @@ pcread:	push	%ax
 	add	$512>>4,%si
 	mov	%si,%es
 	inc	%al			# ++sector
-	cmp	0x151c,%al		# mman::pc_drive_last_sector
+	cmp	0x1d1c,%al		# mman::pc_drive_last_sector
 	jbe	2f
 	mov	$1,%al
 	inc	%dh			# ++head
-	cmp	0x1520,%dh		# mman::pc_drive_last_head
+	cmp	0x1d20,%dh		# mman::pc_drive_last_head
 	jbe	2f
 	xor	%dh,%dh
 	inc	%cx			# ++cylinder
@@ -1291,27 +1291,38 @@ lcheck:	pushf				# check for i8086 / i8088 / i80186
 //	Gets memory map from BIOS.
 e820:	mov	$0x0510>>4,%di		# mman::e820
 	mov	%di,%es
-	xor	%edi,%edi		# es:di is destination buffer
+	xor	%di,%di			# es:di is destination buffer
 	xor	%ebx,%ebx		# ebx is an api state tracker
 1:	mov	$0xE820,%eax		# magic
 	mov	$8+8+4+4,%ecx		# sizeof(struct SmapEntry)
 	mov	$0x534d4150,%edx	# magic number
+	movl	$1,8+8+4/*SmapEntry::acpi3*/(%di) # prefill ACPI attributes;
+					# apparently some buggy BIOSes say
+					# that they return this field, yet
+					# do not fill it correctly
 	int	$0x15			# ax,bx,cx,dx,di → ax,bx,cx
 	jc	9f			# cf = unsupported or abuse
 	cmp	%edx,%eax		# more magic means success
 	jne	9f
 	test	%cx,%cx			# discard empty results
 	jz	5f
+	mov	8/*LODWORD(SmapEntry::size)*/(%di),%eax
+	or	8+4/*HIDWORD(SmapEntry::size)*/(%di),%eax
+	jz	5f
 	cmp	$8+8+4+1,%cx		# discard if ignore flag
 	jb	4f
-	testb	$1/*ignore*/,8+8+4/*SmapEntry::__acpi3*/(%di)
-	jnz	5f
+	testb	$1/*don't ignore*/,8+8+4/*SmapEntry::acpi3*/(%di)
+	jz	5f
 4:	add	$8+8+4+4,%di		# keep entry
 5:	test	%ebx,%ebx		# last entry?
 	jz	8f
-	cmp	$0x1000,%di
+	cmp	$(256-1)*(8+8+4+4),%di
 	jb	1b
-8:	ret
+8:	xor	%ax,%ax			# add a blank sentinel entry
+	mov	$(8+8)/2,%cx
+	cld
+	rep stosw
+	ret
 9:	mov	$REAL(str.e820),%di
 	call	rldie
 	.endfn	e820
@@ -1407,7 +1418,7 @@ pinit:	push	%ds
 //
 //	@see	Intel Manual V3A §4.1.2
 golong:	cli
-	lidt	0x1522					# mman::bad_idt
+	lidt	0x1d22					# mman::bad_idt
 	mov	%cr4,%eax
 	or	$CR4_PAE|CR4_PGE|CR4_OSFXSR,%eax
 	mov	%eax,%cr4
--- a/ape/ape.S
+++ b/ape/ape.S
@@ -485,19 +485,19 @@ ape.mbrpad:
 ape_disk:
 	.stub	ape.diskid,quad
 	.org	0x1be,0x00
-	.macro	.partn	x
-	.stub	ape.part\x\().status,byte	# 0=absent / 0x80=present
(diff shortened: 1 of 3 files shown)


tkchia:
(Actually, there is one needed additional change which I have not pushed, since it is mainly an expedient hack.

The bare metal bootloader is meant to be run as an MBR from track 0, head 0, sector 1 of the first hard disk (or "hard disk").  But my PC's firmware seems to prefer to bypass the MBR, and parse the partition table itself &mdash; and it does not like an empty partition table.

Eventually what I did was to patch the MBR, to add a degenerate "partition entry" at offsets `0x1be`&mdash;`0x1cd` which encompasses the MBR itself and more:
```
000001A0   31 C9 B1 03  01 CA AC 5E  0C 80 EE 5A  AC EE 42 49  1......^...Z..BI
000001B0   79 FA C3 00  00 00 00 00  00 00 00 00  00 00 00 00  y...............
000001C0   00 00 00 00  00 00 00 00  00 00 00 00  00 00 00 00  ................
```
&downarrow;
```
000001A0   31 C9 B1 03  01 CA AC 5E  0C 80 EE 5A  AC EE 42 49  1......^...Z..BI
000001B0   79 FA C3 00  00 00 00 00  00 00 00 00  00 00 80 00  y...............
000001C0   01 00 EE FF  FF FF 00 00  00 00 FF FF  FF FF 00 00  ................
```

The problem is, I am not sure we _should_ be doing this &mdash; creating a partition that includes the MBR itself.)

Thank you!

ghaerr:
Hello @tkchia,

> The problem is, I am not sure we should be doing this — creating a partition that includes the MBR itself.)

Are you saying that what is happening is the that Cosmopolitan boot block, which now needs to double as an MBR, is not being loaded and executed by your BIOS, since the BIOS is parsing the first partition table itself, and trying to load a different volume boot sector?

I haven't decoded your above "degenerate" MBR entry, I assume it points to the Cosmopolitan MBR as the volume boot record, so the Cosmo MBR is loaded and finally executed (at 7c00:0000)?

Thank you!

tkchia:
Hello @ghaerr,

> Are you saying that what is happening is the that Cosmopolitan boot block, which now needs to double as an MBR, is not being loaded and executed by your BIOS

Yes, for some reason this seems to be what is happening. :neutral_face:  If I just leave the empty MBR as-is, then the BIOS just ignores the thumb drive, and boot from the computer's internal hard disk.

> I assume it points to the Cosmopolitan MBR as the volume boot record, so the Cosmo MBR is loaded and finally executed (at [0:7c00])?

Yes, it is simply this:
  * Boot indicator: `0x80` (i.e. active partition)
  * Partition start head: `0x00`
  * Partition start sector (bits 0&mdash;5), high bits of start track (bits 6, 7): `0x00`
  * Partition start track (low bits): `0x00`
  * Operating system indicator: `0xee` (OK, this is totally bogus; I copied it from a partition table entry elsewhere)
  * Partition end head: `0xff`
  * Partition end sector, high bits of end track: `0xff`
  * Partition end track (low bits): `0xff`
  * Starting sector LBA: `0x00000000`
  * Ending sector LBA: `0xffffffff`.

(For what it is worth... personally I think, in the longer term, it would be good if the bare metal code can be installed as a partition, rather than taking up the MBR.)

Thank you!

jart:
> The problem is, I am not sure we should be doing this — creating a partition that includes the MBR itself.)

You're making my dreams come true! Now we can claim APE binaries are disk images!

jart:
Do you need to patch the MBR when you could just update the linker or assembly magic here?

https://github.com/jart/cosmopolitan/blob/55ddf73484d06aa7572e81c94f64b1b8729c9300/ape/ape.S#L477-L506

tkchia:
Hello @jart,

> Do you need to patch the MBR when you could just update the linker or assembly magic here?

Well, it will still be a rather ugly hack, I suppose &mdash; but let me try that. :slightly_smiling_face: 

Thank you!

ghaerr:
Hello @tkchia,

> in the longer term, it would be good if the bare metal code can be installed as a partition, rather than taking up the MBR.

Given that the MBR and (non-partition) VBR share the same signature and executable code required for Cosmopolitan, couldn't one just use your new MBR as the VBR for any partition? That is, any Cosmopolitan .com file should be able to be `dd`'d directly into the partition start and it should boot, provided the MBR knows about it, correct? (And any existing Cosmo .com file should be able to be booted as a VBR directly as a floppy image).

With regards to the MBR with a degenerate partition entry having to be used in the first place - this is only a problem for "pseudo" hard drives like thumb flash drives on your system, correct? I would only be guessing why, but it seems your PC BIOS won't allow execution of unknown MBR code for portable drives (and possibly with good reason, considering the commercially-supplied viral MBR you diagnosed on ELKS a few weeks back). Thus it seems your system would not allow a bare VBR (non-partitioned) boot unless on floppy media.

Thank you!

tkchia:
Hello @ghaerr,

> Given that the MBR and (non-partition) VBR share the same signature and executable code required for Cosmopolitan, couldn't one just use your new MBR as the VBR for any partition?

The volume boot record should start up fine &mdash; that is true.  But it still needs to locate the rest of the Cosmopolitan program after the first 512 bytes.  If the program does not start at track 0, head 0, sector 1, then the rest of the program will not be starting at track 0, head 0, sector 2, and the VBR will need to somehow locate it.  This part is not implemented yet.

Thank you!

ghaerr:
Hello @tkchia,

> If the program does not start at track 0, head 0, sector 1, then the rest of the program will not be starting at track 0, head 0, sector 2, and the VBR will need to somehow locate it. 

I see. So all non-degenerate VBRs (meaning any VBR which doesn't start at track 0, head 0, sector 2) will either have to have the matching MBR partition start_sector long offset encoded into them, or use the sometimes-but-not-always-implemented `DS:SI` (or sometimes `DS:BP`) pointer to the booted MBR partition to get the start_sector offset. I seem to remember at ELKS we couldn't rely on the `DS:SI` MBR partition pointer and had to go with the start_sector written into the VBR by an external tool.

Thank you!

jart (approved):
I know it's a work in progress, but it's still progress! Looking forward to seeing more of these. Please also send an examples/vga.c program to help advertise the presence of this feature and help folks understand how easily they can get started.

jart:
I'm sorry I just realized `examples/hello4.c` is that example. I'm going to rename it to `examples/vga.c` after this change.

tkchia:
Hello @ghaerr,

> [EDIT: There is a third option, which if Cosmo VBRs would be made MSDOS/FAT compatible by including a BPB (BIOS Parameter Block) at offsets 0x0B-0x35 in the VBR, then the start_sector long at offset 0x1C ("hidden sectors") could be read or written by some existing tools. This would also be a convenient standard location for storing CHS should that be required in the future.]

Not very sure how that will work out though.  A boot sector that has a MS-DOS-style BPB will most likely need to start with a short jump or a near jump (`0xeb` or `0xe9`).  This may be a showstopper, if the same bootloader must also look like a shell script and a Windows PE program at the same time (i.e. starting with `0x4d 0x5a`).  Then there is the problem of somehow accommodating a FAT and a root directory...

I guess a (very silly) fourth option might be for the VBR code to (re)load the MBR, then look up an appropriate partition table entry from there.  This can perhaps be combined with one of the other methods you mentioned.

Thank you!

tkchia:
Hello @jart,

> I'm going to rename it to `examples/vga.c` after this change.

Thanks!  Actually I merely modified a pre-existing `examples/hello4.c`, to make it VGA-capable.

> I'm happy you discovered the brilliance of how we're able to use the PML4T to create memory multi-mappings. This is one of the reasons why I want so badly to be able to run my programs on bare metal. Because it's such a powerful capability, to be able to do things like copy memory without copying it.

OK. :slightly_smiling_face:  I am thinking that it should also be good for the `BANE`-based mapping to also cover areas of physical memory that are reserved (`mm->e820[i].type` ≠ `kMemoryUsable`), so we can easily do memory-mapped I/O for some devices.

Thank you!

ghaerr:
Hello @tkchia, hello @jart,

> It turns out we can address physical addresses [0, 0x200000 - 1]
via virtual addresses [BANE, BANE + 0x200000 - 1], even within
__get_virtual( ), rather than having to go through VAs [0, 0x200000 - 1].


I must admit to doing a lot of reading, trying to catch up to both of you with regards to long mode's memory management and the new PML4T register. The current (and future) capabilities of Cosmoplitan's total internal handling of 64-bit long mode and managing all of memory itself are quite impressive. May I ask a few questions, to try to fully understand what is happening in this area?

- At startup, `pinit` creates a mapping of the first 2MB of physical memory by setting up the PML4T (%CR3) and the other page tables at just below 512K (0x79000-0x7FFFF) in low memory. The 2MB memory is mapped to the bottom of the top half of the long mode canonical address space (called BANE). Is this correct?
- Virtual addresses [0..2MB-1] are also mapped to physical addresses of the same.
- Now that the page tables themselves can be accessed directly through BANE in long mode, long mode is entered and `__map_phdrs` is called using 0x500 for the location of the `mman` structure.
- After the BIOS INT 15h function e820h to get the OS/BIOS memory layout, the rest of the physical memory is mapped to BANE offsets through the `__invert_memory` function, while the `__get_virtual` function can be used to allocate new pages for bare metal program use when needed.

Is that how things work?

> I'm happy you discovered the brilliance of how we're able to use the PML4T to create memory multi-mappings. 

Do you mean here that the lower 2MB of physical memory is mapped to VA [0..2MB-1] as well as to BANE+[0..2MB-1]? So that multiple VAs point to the same PAs?

BTW, what does BANE stand for? 

Pretty cool stuff!!

Thank you!

jart:
> At startup, pinit creates a mapping of the first 2MB of physical memory by setting up the PML4T (%CR3) and the other page tables ending at 4K below 512K (0x79000-0x7EFFF) in low memory. The 2MB memory is mapped to the bottom of the top half of the long mode canonical address space (called BANE). Is this correct?

pinit() in ape.S initializes a trie data structure that multimaps the first 2mb of physical memory to BANE and NULL. NULL is 0 and BANE is -140737488355328 or 0xffff800000000000 which is the lowest possible memory address on x86 with the PML4T addressing scheme. So if you want to  have a pointer to memory address 16, then it'd be valid to use either of these virtual addresses: 0x10 or 0xffff800000000010. They point to the same thing. We use BANE to linearly model physical memory. The positive addresses might change, for example, if you decide to unmap NULL to create a guard page. But we still need a way to access those physical memory addresses. So what we do is just bitwise OR (or addition) BANE into the physical address, and then we get a virtual address that means the same thing. Since once we access long mode, there is no way to access physical addresses. You have to indirect through a virtual address that you've defined to map to a physical address.

> Now that the page tables themselves can be accessed directly through BANE in long mode, long mode is entered and __map_phdrs is called using 0x500 for the location of the mman structure.

Correct. The assembly code hands over the BIOS memory map to the C code using that address. Speaking of which, all fixed memory addresses should be documented in the "memory plan" text file. We should add this. Possibly as a comment, since it isn't FRAMESIZE aligned. https://github.com/jart/cosmopolitan/blob/master/libc/runtime/memtrack64.txt

> __get_virtual function can be used to allocate new pages for bare metal program use when needed.

That's a low-level function which allocates pages and returns a pointer to the page table entry. Portable bare metal programs would use `mmap()` and `malloc()` to allocate memory. It should go much faster than than it would on Linux too, since we don't have to pay the 1000ns context switching costs. See https://github.com/jart/cosmopolitan/blob/master/libc/intrin/directmap-metal.c

> Do you mean here that the lower 2MB of physical memory is mapped to VA [0..2MB-1] as well as to BANE+[0..2MB-1]? So that multiple VAs point to the same PAs?

Yup. We can multimap. We can copy memory without moving it. It's a privilege that's reserved only for operating systems.

> BTW, what does BANE stand for?

I call it BANE because it resembles `LONG_MIN` but I probably shouldn't, because one of the advantages of having addresses truncated to 48 bits is we don't have to worry so much about the problematic aspects of `LONG_MIN`. We should change it.

> A related quick question about __normalize_e820

Yeah that looks wrong. Please send a PR fixing it.

paulwratt:
:)  (aye carumba)

good catch @ghaerr , isnt it great being "schooled"

that is a nice design decision @jart and its great to see @tkchia pull it all together

(who said VGA and MBR were not linked :)

PR #611, opened 2022-09-11
Bare metal: do not use `push`/`pop` in long mode until stack set up

tkchia:
This patch fixes a random problem with text segment corruption that would crop up occasionally.


jart (commented):
Thanks!

PR #613, opened 2022-09-12
Bare metal VGA: implement "status report" escape codes

tkchia:


libc/vga/readv-vga.c
```diff
@@ -0,0 +1,49 @@
...
+│ OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,        │
+│ ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR        │
+│ OTHER DEALINGS IN THE SOFTWARE.                                              │
+╚─────────────────────────────────────────────────────────────────────────────*/
+#include "libc/calls/struct/fd.internal.h"
+#include "libc/calls/struct/iovec.h"
+#include "libc/calls/struct/iovec.internal.h"
+#include "libc/vga/vga.internal.h"
+
+ssize_t sys_readv_vga(struct Fd *fd, const struct iovec *iov, int iovlen) {
```
jart, line 32: Nice! This is really coming together. It's going to be a fun project implementing all these system call wrappers. I'm hoping that by running on metal, we avoid the context switching latency, and can shave hundreds of nanoseconds off every system call. Then, we can get something like an HTTP server running on metal and benchmark it.

libc/vga/readv-vga.c
```diff
@@ -0,0 +1,49 @@
...
+  /* NOTE: this routine is always non-blocking. */
```
jart, line 33: It might be worth mentioning how this falls through to to the serial port read if 0 is returned.

jart (approved):
Thank you!

jart:
This is soooo cool.

image

ghaerr:
Hello @tkchia,

Now that I have Cosmopolitan building again on macOS, I added a static yoink of the VGA console to the [D-Flat memopad.com demo](https://github.com/ghaerr/dflat) (built using `make-cosmo`) and tested it outputting to the VGA console using QEMU emulating bare metal and - it runs! 

Pretty cool. Impressive indeed to have Cosmo programs running in long mode on bare metal, with ANSI console support. (Keyboard input isn't working, even from serial, I haven't seen why that might be yet). However, it appears the calculated VGA attribute colors are not quite correct: here's a screenshot:
Cosmo VGA memopad

The colors should be blue background with cyan title bar, as shown on the screenshot on the D-Flat page. I suspect red/blue may be swapped during RGB calculation.

(Note: since QEMU is currently displaying VGA output to serial as well, the ANSI sequences used to turn off the cursor, and turn on mouse reporting are being sent to the terminal running QEMU. As such, on ^C exit, the terminal will be in a bad state. To remedy this, the sequences "\e[?1000;1002;1003;1015;1006l" and "\e[1;0;0m\e[?25h\n" can be echoed to the terminal to reset.)

Thank you!

tkchia:
Hello @ghaerr,

> The colors should be blue background with cyan title bar, as shown on the screenshot on the D-Flat page. I suspect red/blue may be swapped during RGB calculation.

Whoops!  I will try to get this fixed soon.  I think I forgot to swap the colour bits around from the VT100 order to the VGA order in some places.

Thank you!

tkchia:
Hello @ghaerr,

> (Keyboard input isn't working, even from serial, I haven't seen why that might be yet).

This sounds strange indeed.  Perhaps the `qemu-system-x86_64 -d` ... `-D` ... debug logs can shed some light on what is going on?

Thank you!

Issue #35, opened 2021-01-29
cross platform GUI / 2D software rendering


nikhedonia:
I've noticed that cosmopolitan supports the creation of gui applications on windows.
I'm curious on your thoughts on how you'd go about extending this to the remaining platforms.
I'd be happy to make contributions in this direction but I'm not too familliar on how things work on Mac & BSD.

jart:
Cosmopolitan currently has a razor sharp focus on tui interfaces because it's something we're able to do better than anyone else. I think cross-platform GUIs are great, but I need [your support](https://github.com/sponsors/jart) to raise the resources it requires.

The way I'd likely approach the problem is by vendoring something like SDL into the codebase to give you a high-performance blank canvas with audio that works consistently across platforms. I think starting off this way makes the most sense, since games need high-performance graphics and therefore can't be written as web guis like most modern software these days.

Please note that even GUI libraries as simple as SDL usually have a treasure trove of hidden dependencies. Cosmopolitan would need to vendor those too in order to uphold its mission of deterministic hermetically-sealed never-breaking behavior. We accomplish that by not making the assumption that the system did the legwork for us. Therefore it would some research to figure out the best way to vendor SDL dependencies and integrate with window management systems. Here's an example of the research that went into figuring out a way to portably integrate with kernels: [libc/sysv/syscalls.sh](https://github.com/jart/cosmopolitan/blob/master/libc/sysv/syscalls.sh), [libc/sysv/consts.sh](https://github.com/jart/cosmopolitan/blob/master/libc/sysv/consts.sh), and [libc/nt/master.sh](https://github.com/jart/cosmopolitan/blob/master/libc/nt/master.sh).

I can make it happen. I believe it should happen. I don't think 200mb Electron hello world binaries are sustainable. Cosmopolitan can do it better than anyone else, if we raise the resources. So please check out https://github.com/sponsors/jart and tell your friends who might be interested in becoming sponsors.

jart:
If you want to see some examples of the impact Cosmopolitan is already having in the terminal interface realm, then check out:

- [printvideo](https://justine.lol/printvideo.html) renders .mpg videos in the terminal
- [nesemu1](https://justine.lol/nesemu1.html) play nintendo games in the terminal
- [apelife](https://justine.lol/apelife/index.html) conway's game of life tui + gui
- [memzoom](https://justine.lol/memzoom/index.html) process/file memory monitor
[...]

PR #616, opened 2022-09-13
Bare metal VGA: clarify code comments

tkchia:


jart (approved):
Approved. What would we have to do to get an RGB array canvas on which to draw the characters ourselves?

tkchia:
Hello @jart,

> Approved. What would we have to do to get an RGB array canvas on which to draw the characters ourselves?

Thanks!  The 80 × 25 × 16 text mode which I currently use, via
```
	mov	$0x0003,%ax
	...
	int	$0x10
```
simply does not provide an RGB array canvas.  (Though it does allow one to switch to a different set of font glyphs, or change the font height.)

To get a pixel-by-pixel canvas, we will need to actually switch to a different mode, a graphical screen mode with a memory-mapped video frame buffer.  This can be done using [VESA BIOS Extensions](https://www.delorie.com/djgpp/doc/rbinter/ix/10/4F.html).

If you ask me, I might implement the whole thing this way:
  * In real mode:
    * use VBE to enumerate video modes supported by the BIOS;
    * try to find a video mode which uses a pixel format (RGBX8888, BGRX8888, RGB565, etc.) we support;
    * switch to the video mode, and also note down the frame buffer's physical address;
  * Then in protected mode:
    * map the frame buffer into virtual memory;
    * plot pixels in the frame buffer.

Thank you!

ghaerr:
Hello @tkchia, hello @jart,

> To get a pixel-by-pixel canvas, we will need to actually switch to a different mode, a graphical screen mode with a memory-mapped video frame buffer.

Wow! Should a memory-mapped framebuffer be made available to bare metal programs via VESA BIOS, that would immediately allow a large number of graphical programs to run standardized on bare metal, including the Microwindows/Nano-X implementation already ported to Cosmopolitan (discussed in https://github.com/jart/cosmopolitan/issues/35#issuecomment-1085354969 and implemented in https://github.com/jart/cosmopolitan/issues/35#issuecomment-1097504620). (Although mouse and keyboard support would still need implementation, hopefully through standard BIOS calls).

In addition to quickly providing the ability to write and run programs using GUI frameworks such as Nuklear, MicroUI, Win32, Nano-X and others, a small-stripped down [Single header Truetype/Freetype library](https://github.com/nothings/stb/blob/master/stb_truetype.h) could be used to draw any Unicode glyph using one of the fonts @jart previously suggested, compiled into the APE executable.

> switch to the video mode, and also note down the frame buffer's physical address;

We also need the frame buffer pixel format to be saved (in the `mman` struct along with the other BIOS data?). I have blit/conversion routines for any number for 32-bit (and smaller) formats, including 32-bit ARGB, RGBA, etc for fast drawing, including alpha-blending, if needed.

> If you ask me, I might implement the whole thing this way:

Sounds great! Since (software) scrolling graphical frameuffers can be quite slow, we might want a mechanism to keep the 80x25x16 text mode for bare metal programs that need a VGA console but no graphics or non-CP437 glyphs.

> plot pixels in the frame buffer.

Given a framebuffer address and pixel format, I could produce a number of small demos that would allow Win32 API and (X11-like) Nano-X API in fairly short order. Even if STB-Truetype is used for VGA console driver text output, an upper level could load other fonts or possibly change the console driver font scaling factor.

> Though it does allow one to switch to a different set of font glyphs, or change the font height.

When changing to a different set of glyphs, does that change the entire set of 256 code point glyphs, or can individual glyphs be set per code point?

Thank you!

jart:
Are you suggesting we implement the WIN32 API? I want a framebuffer so we can have a terminal that's larger than 80 x 25 characters. But we could expose a frame buffer API to applications so they could draw whatever they want. How could it be possible that a VGA display is slow? Assuming we buffer frames in normal memory, how many times could we copy the whole frame to the monitor?

tkchia:
Hello @jart,

> How could it be possible that a VGA display is slow? Assuming we buffer frames in normal memory, how many times per second could we copy the whole frame to the monitor?

Scrolling a large graphics frame buffer using software &mdash; i.e. doing a `memmove( )` from one area of video memory to another &mdash; is slow.  I can attest to that.  If I remember correctly, scrolling up a 800&times;600 (or thereabouts) screen full of RGBX pixels in this way would take about &frac14; second or so.

However, it should be much faster to copy a frame from main memory to video memory.

> I want a framebuffer so we can have a terminal that's larger than 80 x 25 characters.

There are several ways to do that with VESA VBE (each with its trade-offs).  Switching to a "true colour" video mode probably offers the most flexibility.  But we can also try to e.g. use an "indexed colour" graphics mode, or even switch to a different text mode.

In any case, let me see if I can prototype out some sort of graphical frame buffer support, to help advance the discussion.

Thank you!

ghaerr:
Hello @jart, hello @tkchia,

> Assuming we buffer frames in normal memory, how many times per second could we copy the whole frame to the monitor?
> However, it should be much faster to copy a frame from main memory to video memory.

Video versus normal memory isn't usually the problem, instead just copying large data blocks takes time. While the time it takes to display a frame of a double-buffered graphics application and scrolling one line of text is about the same, scrolling usually adds considerably to user latency because the screen needs to scroll on every LF output, and of course, text users love to scroll screens up and down a lot, so scrolling 10 lines takes ten times as long. There are ways around this, using 'jump scrolling', described below.

To quantify text versus graphics scrolling a bit more, lets say that scrolling an 80x25 text screen takes time 1. This would include moving 80 x 25 x 2 (attribute+char code) = 4000 byte memcpy. Consider a graphics screen using the same font size (8x16), and running 32-bit RGBA (truecolor) pixel format. Of course, the actual displayed glyph size depends on monitor DPI as well as resolution, but that means moving 80 x 25 x 8 x 16 x 4 = (640x400 x 4) = 1,024,000. So we're talking 256 times slower. No one user 640x400 anymore; even running a 1920x1200 display (which easily allows more text rows/columns) would take 1920x1200x4 = 9,216,000 byte memcpy. Scrolling this screen once would be 2,304 times slower than the super-fast text scroll. I'll leave it to you to calculate what the means in real time using the very fast Cosmopolitan `memcpy`.

All modern terminal emulators (and I would suggest us too, should we implement a graphics console) use a 'jump scroll' technique where the queued input buffer is scanned for LFs, which are counted and then the screen is scrolled once for the total linefeed count, then the rest of the data drawn. This speeds things up by an order of magnitude, usually.

All this said, I'm all for a graphics console, provided that we can also have a text console, as text modes are always faster. Other ways to greatly speed up text output are:
- Use 24-bit RGB which is 25% faster since less data moved (excluding alignment complications)
- Use 16-bit (RGB565) which is 50% faster, but at the cost of less color resolution
- Scroll less screen data by using windows (this is why "terminal emulators" seem so fast - they're scrolling much less of the screen)
- Instruct Freetype/truetype to pre-create and cache bitmaps instead of using floating point based glyph drawing each time
- Always clear background bits first to background color in single fillrect, then draw foreground bits only for each character

Thank you!

ghaerr:
> Are you suggesting we implement the WIN32 API?

No, although I have a single-file implementation of [Win32 GDI](https://github.com/ghaerr/microwindows/blob/master/src/mwin/wingdi.c) and [Win32 USER](https://github.com/ghaerr/microwindows/blob/master/src/mwin/winuser.c) should we want to. 

I'm thinking that since Cosmopolitan's scope is actually quite wide, with lots of different user focus, it might be best to keep Cosmo's implementation of graphics capabilities low-level, with an eye towards allowing other developers to construct a (standard or non-standard) graphics API on top of it, rather than attempting to pick some standard. This is because few agree on what's the best graphics or windowing API, and GUI libraries can take radically different approaches (think message- and callback-driven Win32 versus more modern immediate-mode libraries like Nuklear or MicroUI).

Thus, a layered graphics implementation similar to a network stack makes a lot of sense. Let me describe that a little more:

For Cosmopolitan, the lowest level could be a set of routines to search for and return a framebuffer (possibly polyfilled as usual for bare metal versus Linux versus XNU etc) would allow a console to attach and draw to it, as well as be the anchor for porting standard graphics libraries (as third_party) to Cosmo. Ideally, a framebuffer size and requested pixel format would be passed, along with possibly an enumeration routine called beforehand.

On top of that, an "engine library layer" could include routines that could be used by other Cosmo components (think VGA graphics console, Unicode glyph selection, etc) that might include the following:
- Fast 64-bit BLIT routines for rapid data transfer, possibly with pixel format conversion
- Fill rectangle routine for rapid space-filling using standardized RGBA color input and conversion to actual framebuffer format
- Truetype font draw routines using STB-Truetype along with Unicode and font loading/compilation for standardizing Cosmo graphical text output to always be correct
- Color <-> pixel conversion routines

You'd be surprised at what kinds of astounding graphical output can be produced with just fill rect, blit, and text output. (See the tiny immediate-mode [MicroUI](https://github.com/ghaerr/microui) for an example).

The above would allow for a possible VGA graphics console on bare metal (or also Linux/XNU), while also allowing development of popular graphics APIs on top of Cosmo. The layered architecture makes it possible. For instance, the VGA console should not be hard-wired into VESA BIOS, but rather make a Cosmo call to get a framebuffer, which is redirected to VESA BIOS for bare metal. It then uses library routines (fill rect, truetype glyph output, etc) to portably draw into the framebuffer, with only the library concerned with hardware pixel format. Likewise, a Cosmo program running on Linux could call Cosmo to get a framebuffer, which ends up opening up a graphical SDL window that allows pixels to be drawn by `memcpy`ing to the returned framebuffer pointer. Later, the non-text mode "VGA console" itself is abstracted to become a "graphical console" and could run on any framebuffer, not just bare metal.

Thank you!

jart:
Our memmove() goes 134 GB/s.

https://github.com/jart/cosmopolitan/blob/134ffee51900a4d70e8e68b598374856ceec8355/libc/intrin/memmove.c#L59

According to my calculation, copying memory backwards on a 640x480 display would take 22µs.

```c
char mem[400][640][4];

void Move(void) {
  memmove(mem, &mem[10][0][0], &mem[400][0][0] - &mem[10][0][0]);
}

BENCH(move, bench) {
  EZBENCH2("move", donothing, Move());
}
```

If I do the same thing with a 1080p framebuffer, then it takes 200µs. Does anyone want a 5000 FPS Cosmo HD display?

```
 *     move 640x480        l:    68,204c    22,030ns   m:    68,963c    22,275ns
 *     move 1920x1080      l:   646,694c   208,879ns   m:   642,830c   207,631ns
```

I know not all memory behaves the same way and that video mapped memory can go slower than normal memory. But that shouldn't matter since we can just do our work in normal memory and then take a snapshot every so often which gets copied to the video memory.

ghaerr:
Thanks for the calculations @jart, very interesting!

> video mapped memory can go slower than normal memory.

Yes, in particular not sure whether PC cards allow for 64-bit moves on the bus or whether it's 32-bit memory. Of course, most all graphics now is done by the GPU, but that's an entirely different story.

Sounds like things could be quite fast on your system, and scrolling might not be an issue. However, the original glyph drawing code must be taken into account, and that can take lots of cycles too when displaying truetype, which, for the best visual case will use an 4- or 8- bit alpha value to blend the glyph edges into the background... (which then requires an alpha-blend blit, which is slower and can't use memcpy, etc).


> we can just do our work in normal memory and then take a snapshot every so often which gets copied to the video memory. 

Exactly. Not processing immediately also allows for the 'jump scrolling' LFs to accumulate (providing we have enough buffer space) and we get that benefit as well. Having a system-tunable "snapshot" timer allows for individual systems to be tuned for the cost/performance tradeoff, both for graphics VGA console as well as arbitrary double-buffered framebuffer output.

jart:
You don't seem to want to accept that you're dealing with an expert. Cosmopolitan can do things like to play 4K (2864x2160) MPEG videos in the terminal using just CPU (no GPU), with half as much resources as it takes AVLIB, because we wrote the core graphical algorithms by hand. https://twitter.com/JustineTunney/status/1412408110040354816/photo/1 I think 80x40 BIOS text display is a great stepping stone, but I wouldn't want to impose that limitation upon myself due to untested assumptions about performance. Whatever we build, we'll find a way to make it go fast. The important thing is that we build.

ghaerr:
Cool @jart, I have to run just to keep up with understanding things you've written years ago. Then again, that's one of the things I really enjoy about this project! :)

I agree completely about untested performance assumptions. Looking ahead, a very exciting thing about a graphical VGA console is the ability to use and display all Unicode on bare metal, with no OS, thus keeping total compatibility across each of the Cosmopolitan platforms. So - let's keep the pedal to the (bare) metal - and keep building! (And by the way, if we can produce a super-fast truetype cached glyph renderer (with antialiasing and sub-pixel positioning using integer arithmetic), that'll be worth its weight in gold - see [text rasterization problem description](https://github.com/ghaerr/agg-2.6/blob/master/agg-doc/pdf/Anti-Grain%20Geometry%20-%20Texts%20Rasterization%20Exposures.pdf) and [cl-aa algorithm](http://projects.tuxee.net/cl-vectors/section-the-cl-aa-algorithm)).

> Cosmopolitan can do things like to play 4K (2864x2160) MPEG videos in the terminal using just CPU (no GPU), with half as much resources as it takes AVLIB, because we wrote the core graphical algorithms by hand.

I didn't know this, and that's very good news, as software-only graphics needs all the help it can get to compete with modern GPU graphical display systems. I look forward to helping integrate fast algorithms into Cosmopolitan production code.

tkchia:
Hello @jart, hello @ghaerr,

OK, I just did up a quick (?) proof of concept bootloader, modified from `tool/build/emubin/spiral.c`, of how one might use VESA to switch to a graphics mode with a frame buffer, and draw into the buffer: https://github.com/tkchia/cosmopolitan/commit/dcc6133df3e30ef32fa8d6fe195086ca1f6fbb96.

(Admittedly, this was also partly to convince myself that this whole jig might possibly work.  I had not really dabbled in VBE frame buffers or unreal mode very much before. :eyes: )

The code switches to unreal mode &mdash; i.e. real mode but with 4 GiB segment limits &mdash; at the start.  This is needed because the frame buffer lies outside of the &approx; 1 MiB space that one can address via "normal" real mode.

Thank you!

ghaerr:
Hello @tkchia,

> I just did up a quick (?) proof of concept bootloader, modified from tool/build/emubin/spiral.c

Looks interesting. Can I ask how you are running this (via QEMU)? (I ask because I tried building the Cosmopolitan CGA demo given in https://github.com/jart/cosmopolitan/pull/588#issuecomment-1243162814), but for some reason (I'm on macOS) I can't get that spiral.bin to boot in QEMU nor run in blinkenlights (blinkenlights gets a segmentation violation and exits).

Also, does emubin/vesaspiral.bin include or execute the A20 gate code from ape/ape.S at startup? Unreal mode requires A20 gate on, which isn't always the case at PC boot, and I haven't been able to investigate the spiral demo (yet).

Thank you!

tkchia:
Hello @ghaerr,

> Looks interesting. Can I ask how you are running this (via QEMU)?

The same as always: `qemu-system-x86_64 -hda o/tiny/tool/build/emubin/vesaspiral.bin`.  (`-serial stdio` is optional here.)

> (I ask because I tried building the Cosmopolitan CGA demo given in [#588 (comment)](https://github.com/jart/cosmopolitan/pull/588#issuecomment-1243162814)), but for some reason (I'm on macOS) I can't get that spiral.bin to boot in QEMU nor run in blinkenlights (blinkenlights gets a segmentation violation and exits).

Are you able to build and run `vesaspiral.bin`?  You should be able to use a similar command to build it:
```
make -j16 MODE=tiny o/tiny/tool/build/blinkenlights.com o/tiny/tool/build/emubin/vesaspiral.bin
```
[Here](https://github.com/jart/cosmopolitan/files/9584993/20220916.zip) are the binaries for `spiral.bin` and `vesaspiral.bin` that I built.

> Also, does emubin/vesaspiral.bin include or execute the A20 gate code from ape/ape.S at startup? Unreal mode requires A20 gate on, which isn't always the case at PC boot, and I haven't been able to investigate the spiral demo (yet).

Darn, I forgot to do that in the `unrealstart.inc` code.  I should probably correct this &mdash; and also make a few other corrections &mdash; if I am to submit this code upstream.

Anyway, this whole exercise was mainly to show (and see) how frame buffer graphics with the help of VESA VBE might work. :slightly_smiling_face: 

Thank you!

ghaerr:
Hello @tkchia,

Thank you for your response to my (seemingly dumb) questions. Yes, I have been having trouble building the emubin .bin binaries, and was having trouble seeing how these booted since I didn't see where the boot code+signature was being added (which is necessary for the -fda and -hda QEMU arguments, correct?). I'll look further into both `spiral.bin` as well as downloading and running `vesaspiral.bin` and report back.

> I forgot to do that in the unrealstart.inc code. 

With regards to the unreal mode ASM code being included in a .c file using the `asm` keyword: it's potentially pretty neat to be able to include larger chunks of ASM in C files, but I find it hard to read in bulk, with all the \t's and \n's; would it be possible to use something similar to the following, where tabs are expanded to spaces?:
```
asm(".section .start,\"ax\",@progbits\n"
     "    .globl    _start\n"
     "_start:\n"
     "    ljmp    $0,$1f\n"
     "1:  cli\n"
     "    lgdt    %cs:.gdtr\n"
     "    mov     %cr0,%EAX\n"
     "    or      $1,%al\n"
     "    mov     %EAX,%cr0\n"
...
```
More specifically, does `asm` require a TAB in front of each opcode, or could spaces also work?

With regards to how unreal mode is being used with the vesaspiral.c and its compilation: are these files being compiled to produce 16-bit C code, or 32? That is, unreal mode requires an `addr32` prefix on data segment DS and ES accesses to access offsets greater than 64K; is that prefix being generated somewhere, or does the data segment GDT have the "large/32-bit" bit set (I forget the name of the bit) being set to automatically execute 32-bit data fetches on DS/ES accesses? (If the latter, then I assume the entire program is compiled and run as a 16-bit executable).

Sorry for my confusion - it is compounded by the fact that when blinkenlights (on my system) emulates `spiral.bin`, it segfaults on accessing address 0x7E00, which is strange, except that's the address the spiral is being generated into, thus my confusion regarding the overall intended versus actual execution environment. (And yes, it is strange that blinkenlights is segfaulting running `spiral.bin`, given that @jart just posted a screenshot of it working). I will dig into debugging the emulation issue once I more fully understand the intended execution environment.

Thank you!

ghaerr:
Hello @tkchia,

Update on QEMU and `spiral.bin`: both your `spiral.bin` and `vesaspiral.bin` run with QEMU, I see the 0xAA55 signature on your versions, so that answers by boot signature question above. (And `vesaspiral.bin` looks nice :) The version of `spiral.bin` built by my (macOS) system starts with a special @jart-defined NOP `boframw` (defined by a macro and treated specially by blinkenlights). This version is only 480 bytes long and is missing the 0xAA55 signature. Thus, this is a macOS build problem which I will look into and report a fix elsewhere. I am guessing that it appears that the .bin files are being built using some other Cosmo startup code not intended for the 16-bit bootable .bin binaries being discussed here.

(Here's a screenshot of my incorrectly built `spiral.com`, just for reference. Perhaps it is in fact just garbage):

Screen Shot 2022-09-16 at 9 22 53 AM

Thank you!

tkchia:
Hello @ghaerr,

The numerous `dec %ax` "instructions" are most probably supposed to be 64-bit `rex.w` prefixes.  It looks like `spiral.c` was for some reason built as 64-bit code, when Blinkenlights (and QEMU) was expecting code meant for 16-bit mode.

Incidentally, I use a somewhat different build process for `vesaspiral.c`.

> are these files being compiled to produce 16-bit C code, or 32? That is, unreal mode requires an addr32 prefix on data segment DS and ES accesses to access offsets greater than 64K; is that prefix being generated somewhere, or does the data segment GDT have the "large/32-bit" bit set (I forget the name of the bit) being set to automatically execute 32-bit data fetches on DS/ES accesses?

For building `vesaspiral.c`, I use GCC's `-m16` switch, which produces code for _16-bit_ mode, and emits ELF32 (rather than ELF64) object files.  The compiler inserts the necessary `0x66` and `0x67` instruction prefixes in this case.

The quick switch to and from protected mode is to expand the segment _limits_ in the `ds`, `es`, and `ss` registers' [descriptor caches](https://wiki.osdev.org/Descriptor_Cache), to allow addresses above `0xffff`.  If the CPU tries to read from `[eax + ecx]`, and `eax + ecx` = `0xe0000000`, but `ds`'s limit is only `0xffff`, we will get a CPU exception.

Thank you!

ghaerr:
Hello @tkchia,

Thank you very much for your thoughts and answering my questions, your knowledge helps a lot! That points me in the direction of finding out why spiral.bin is getting erroneously built as 64-bit on macOS.

> I use GCC's -m16 switch, which produces code for 16-bit mode, and emits ELF32 (rather than ELF64) object files. The compiler inserts the necessary 0x66 and 0x67 instruction prefixes in this case.

Wow, very nice! So the compiler will emit code using the 0x66 prefix for 32-bit register loads, and always emits the 0x67 address-size override when addressing memory. That could come in handy in some other projects! Is that supported in your `ia16-elf-gcc` compiler for ELKS?

With regards to your vesaspiral.c code switching to VESA mode 0x10D on line 110, I found [this article](https://wiki.osdev.org/VESA_Video_Modes) states that in VESA 2.0, the (predefined) VESA Defined Mode Numbers were all deprecated, and that the program should search for a desired mode instead, (and use that number). I'm wondering out of curiosity, whether you've tried any mode enumerations, and whether (hopefully) there might be some modes with much higher resolution available?

The article also contains some sample code for enumerating VESA modes.

If vesaspiral.c were modified to take an argument that enumerated modes and displayed them, it could be interesting to run on a variety of systems to see what kinds of (automatic VESA) resolutions might be available. I mention this because I think I could pull some demo code together fairly quickly to demonstrate Truetype font display/draw code at different point/pixel sizes which might then be easily added to a framebuffer VGA console, but the whole project is only really worth it if much higher resolutions are available.

(Intersestingly, the same article states that reading from video memory is very slow, so slow that double-buffering is recommended instead. Thought you might find that interesting).

Thank you!

jart:
spiral.c is only able to be built in `MODE=tiny` which is probably where you got hung up. It *is* built in 16-bit mode because the build requests `spiral.real.o` which causes the Makefile to insert a `-wrapper build/realify.sh` flag to GCC which uses this sed script https://github.com/jart/cosmopolitan/blob/master/build/realify.sed to turn x86-64 assembly into i8086 assembly. It works really well if your code is written in a word-size agnostic way (i.e. you're not doing serialization). It produces significantly tinier code than `-m16` which is essentially a 32-bit mode that always uses an OSZ prefix. The one weakness realify.sed has is it's not able to cope with situations where GCC generates code with REX.B prefixes, so it's best to avoid operating on bytes.

jart:
> would it be possible to use something similar to the following, where tabs are expanded to spaces?:

I would prefer that inline asm code use `\t` rather than actual tabs. The compromise I would propose is that, if you have multiple lines of assembly that would be ugly to put in C string literals, then please consider placing that code in a `.S` file instead. You can use tabs there.

ghaerr:
Hello @jart, hello @tkchia,

Thanks to both your help, I've identified the macOS build problem with `spiral.bin`. The cause is that the `realify.sed` script is not being run to process the 64-bit -> 16-bit conversion because of the way the gcc -wrapper `realify.sh` determines the assembler filename (it uses `if [ "${1##*/}" = as ]; ...`).

For Linux, the toolchain is prefixed by `o/third_party/gcc/bin`, while (currently on my macOS uncommitted version) the assembler is not prefixed (it uses `AS = x86_64-linux-musl-as`). Thus, no `/` is seen and the realify.sh doesn't set 16-bit code nor invoke `realify.sed`.

I will submit a patch for this, if that is OK. (I think a good path forward would be to allow a TOOLSPREFIX=/usr/local/bin set in `~/.cosmo.mk` which would be then be substituted for `o/third_party/gcc/bin` in `build/definitions.mk`, if you're OK with that). That would allow macOS users like me to (experimentally, I know) build and contribute to Cosmopolitan by using Felker's homebrew-installed x86_64-linux-musl toolchain.

Thank you!

jart:
Please submit a change for both. Be sure to send the copyright assignment intent email described in CONTRIBUTING.md.

paulwratt:
> the (predefined) VESA Defined Mode Numbers were all deprecated

@ghaerr although that is true, there are some mode numbers that are still usable, including with VESA 3.0, but I dont think those are the ones you were looking for (ie higher resolution) - I dont remember exactly which modes they were, but I think there was 3 of them (I wrote my own VESA mode switcher in 2000-ish).

As for searching for better/higher modes (text or gfx), remember that although the card/gpu has the mode, the screen might not support it, so be careful (thats essentually what the Windows control Panel does, block out the modes that the screen (supposedly) cant support).

tkchia:
Hello @jart, hello @ghaerr,

I am currently trying to figure out a good way to add support for VESA frame buffer graphics modes to the current VGA code.

Currently the code is kind of meant to be optimized &mdash; with the compiler's help &mdash; to work specifically with 80 × 25 × 16 text mode.  And the (fluffy) goal is to get the code to work with _some_ frame buffer graphics mode whose geometry may be unknown until run time.  _And_ hopefully the code will still support a console using the good old text mode as a sort of fallback (?), and still keep it all reasonably small and fast (?).

So I guess the question is, whether there is a way for me (or us) to get from point _A_ to point _B_, so to speak, without going completely crazy. :face_with_spiral_eyes::dizzy:

Hello @ghaerr,

> > I use GCC's -m16 switch, which produces code for 16-bit mode, and emits ELF32 (rather than ELF64) object files. The compiler inserts the necessary 0x66 and 0x67 instruction prefixes in this case.

> Wow, very nice! So the compiler will emit code using the 0x66 prefix for 32-bit register loads, and always emits the 0x67 address-size override when addressing memory. That could come in handy in some other projects! Is that supported in your `ia16-elf-gcc` compiler for ELKS?

No &mdash; `ia16-elf-gcc` does not know anything about 32-bit registers or addresses.  But you might be able to link `ia16-elf-gcc` compiled code with code produced by (say) a mainline `i686-linux-gnu-gcc` compiler.

> With regards to your vesaspiral.c code switching to VESA mode 0x10D on line 110, I found [this article](https://wiki.osdev.org/VESA_Video_Modes) states that in VESA 2.0, the (predefined) VESA Defined Mode Numbers were all deprecated, and that the program should search for a desired mode instead, (and use that number). I'm wondering out of curiosity, whether you've tried any mode enumerations, and whether (hopefully) there might be some modes with much higher resolution available?

Well, after writing `vesaspiral.c`, I did try doing a VESA mode enumeration (`int 0x10`, `ax` = `0x4f00`) on my "legacy BIOS" machine &mdash; by manually running a `debug.com` session in FreeDOS.

It turns out that, on my machine at least:
  * Mode `0x10d` is actually unsupported (argh!).
  * However, there is a few other modes that are usable, including mode `0x111` (640 × 480 × 64K colors).  I later did get `vesaspiral.c` working with this screen mode.
  * There are a few other modes that are returned in the video mode list, but some of these turn out to be unusable.  (Ralf Brown's Interrupt List did warn about this...)
  * For some reason, the OEM mode numbers, with mode numbers &le; `0xff`, are not returned in the mode list.

> (Interestingly, the same article states that reading from video memory is very slow, so slow that double-buffering is recommended instead. Thought you might find that interesting).

Well, I did come across that warning a while back, and shrugged it off at first.  That was a bad decision. :neutral_face: 

Thank you!

ghaerr:
Hello @jart,

> The cause is that the realify.sed script is not being run to process the 64-bit -> 16-bit conversion because of the way the gcc -wrapper realify.sh determines the assembler filename

Looking further into this, I have finally completely understood and fixed problem; long story short is the issue isn't the way `realify.sh` determines the -wrapper executable as I previously stated (which works), but actually that the macOS `sed` program is buggy! (Yes, unbelievably it doesn't work with the 64-bit -> 16-bit `realify.sed` script passed to it, but only for some input!!)

For instance, here's the `sed`script input and output diff, using the broken macOS `sed` (note `sed` works fine on Linux, only fails on macOS):
```
154,155c154,155
< 	.cfi_offset 6, -16
< 	movq	%rsp, %rbp
---
> 	.cfi_offset 6,-16
> 	movq	%sp,%bp
157,158c157,158
< 	andq	$-16, %rsp
< 	subq	$16, %rsp
---
> 	andq	$-16,%sp
> 	subq	$16,%sp
```
In the last lines above, `sed` failed to convert `andq` to `and`. The build outputs the following error messages on the above:
```
/Users/greg/net/cosmopolitan/o/tmp/o_tiny_tool_build_emubin_spiral_real_o.l3jfn0/cckJcsn0.s: Assembler messages:
/Users/greg/net/cosmopolitan/o/tmp/o_tiny_tool_build_emubin_spiral_real_o.l3jfn0/cckJcsn0.s:155: Error: unsupported instruction `movq'
/Users/greg/net/cosmopolitan/o/tmp/o_tiny_tool_build_emubin_spiral_real_o.l3jfn0/cckJcsn0.s:157: Error: invalid instruction suffix for `and'
```
The problem is fixed by replacing `sed` by `sed.com` in `realify.sh`. Kind of amazing, that an APE binary works better than the macOS standard version, huh?! I am seeing the wisdom now in having tools included in a standalone repo for a project of this complexity.

I plan on submitting a PR with this fix for `spiral.bin`, along with the TOOLSPREFIX= enhancement. I mention this in the case where `sed.com` might be needed before it is built (unlikely), but sed.com might want to be present in build/bootstrap.

Thank you!

ghaerr:
Hello @tkchia,

> So I guess the question is, whether there is a way for me (or us) to get from point A to point B, so to speak

Thank you for your work and testing on VESA BIOS graphics modes, as well as the enumeration test results, I found it quite interesting. I am disappointed to learn that no high resolutions seem to be reported as available (although the 640x480x16 mode could be usable for development, but not likely meeting user expectations for a desktop graphics console).

At this point, it doesn't seem to me the major issue is how get a graphical VGA console working that also works in 80x25 text mode with compiler-optimization-eliminated code for unused features. Instead, it seems just getting a decent 32-bit RGBA linear frame buffer (from the BIOS) remains a big problem.

As such, I'm thinking that, instead of looking at VESA BIOS, perhaps looking at other systems built on bare-metal, but still somewhat small, might be worthwhile. For instance, SerenityOS or Haiku OS might have some low-level display driver code that might be usable as a frame buffer driver for Cosmopolitan bare metal. Yes, this brings the big problem of specific video card-level support, which of course is usually firmly in the operating system driver purview. I'll take a browse around those and other projects and see what things look like. Those operating systems basically have the same problems as Cosmopolitan bare metal would have, with regards to hardware compatibility.

> However, there is a few other modes that are usable, including mode 0x111 (640 × 480 × 64K colors). I later did get vesaspiral.c working with this screen mode.

Nice! I would think that having a vesaspiral.c demo contributed that works in 640x480x16 could be useful for further experimentation in Cosmopolitan, even just using unreal mode (provided @jart agrees). The next step might be building vesaspiral.c to run in long mode, for further experimentation using the regular parts of the Cosmo library. I don't know whether the VESA enumeration routines and video mode switch can run in long mode, or would all have to be done in real mode prior.

> For some reason, the OEM mode numbers, with mode numbers ≤ 0xff, are not returned in the mode list.

Would these OEM modes be the ones that might implement the higher resolutions or 32-bit pixel formats we're looking for? Is there video card documentation that might allow for an extended enumeration to be performed?

Thank you!

tkchia:
Hello @paulwratt,

> As for searching for better/higher modes (text or gfx), remember that although the card/gpu has the mode, the screen might not support it, so be careful (thats essentually what the Windows control Panel does, block out the modes that the screen (supposedly) cant support). Better to use EDID to get a list of monitor supported modes first, and choose from that.

I would like to know more about this.  What is the correct way to obtain this EDID information, under a "legacy BIOS" setup?  How can we use this information to filter the list of video modes returned by VESA VBE?  Thank you!

---

Hello @ghaerr,

> At this point, it doesn't seem to me the major issue is how get a graphical VGA console working that also works in 80x25 text mode with compiler-optimization-eliminated code for unused features. Instead, it seems just getting a decent 32-bit RGBA linear frame buffer (from the BIOS) remains a big problem.

Well, I am not so sure &mdash; a 640 &times; 480 &times; 64K colours mode _does_ sound like quite a decent graphics mode, in my book.

In any case, yes, the enumeration on my PC _did_ produce quite a few other mode numbers &mdash; including one mode `0x17f` which was effectively a BGRX graphics mode with a resolution exceeding 1024 &times; 768.  I just have not got around yet to actually trying out these more exotic video modes.

(On QEMU, mode `0x17f` happens to be a 1440 &times; 900 &times; 16M video mode.  I guess it is probable that the same mode number will mean different things on different machines.)

Thank you!

ghaerr:
Hello @tkchia,

> a 640 × 480 × 64K colours mode does sound like quite a decent graphics mode, in my book.

Ok. Although short on resolution and colors for non-text graphics, 640x480x16 could work quite well for a first-pass graphical VGA console supporting 80x30 text output using 8x16 bitmapped (non-aliased) character glyphs, and enough colors to differentiate all VGA text attributes distinctively. Also, starting with 640x480x16 will surely work on every monitor, temporarily removing the concerns that @paulwratt brings up about setting modes a monitor can't support (which can actually destroy the monitor in some cases).

If interested, I can contribute a set of portable 64-bit routines (perhaps in libc/vga/lib?) that will allow rectangle filling and bitmap text output (with later truetype support for unicode glyphs), along with a compiled in CP437 font, which would enable the VGA console to be enhanced to run in a preset graphics mode either selected at compile-time or dynamically determined during real mode startup. These routines would allow conversion of colors specified in RGB format to 16-bit 565 (64K) or 555 (32K) format, as well as (future) support for 32-bit RBGA, BGRA or 24-bit RGB/BGR pixel format. That would allow us to continue moving forward on a bare metal graphics console without having to get into the very messy details of rolling our own frame buffer.

> On QEMU, mode 0x17f happens to be a 1440 × 900 × 16M video mode.

Thanks, that will be useful for testing. I've been trying to track down the QEMU-supported VESA modes, might you know where they are listed?

> whether there is a way for me (or us) to get from point A to point B, so to speak, without going completely crazy. 

Perhaps use a static yoink "graphics_console" to cause a basic VESA VBE graphics mode to be selected at real mode startup, and then wrap the lowest level text output routines in the VGA console with a function that either writes a CP 437 character glyph index to a BANE+0xB800 offset, or uses the above graphics text drawing routine to draw the compiled-in bitmap glyph to the VBE frame buffer address+BANE. When the "graphics_console" yoink is not present, the graphics routines are compile-optimized out.

Thank you!

tkchia:
Hello @ghaerr,

I have started a new pull request (https://github.com/jart/cosmopolitan/pull/631) as a small step towards generalizing the VGA console support.

Experimentation-wise &mdash; I see that Microwindows already has code for DJGPP on MS-DOS which interfaces with VESA VBE?  If so, then you are already halfway there. :slightly_smiling_face:  Most probably you just need to tweak the Microwindows code a bit, to explore the video modes available through VESA VBE.

> I've been trying to track down the QEMU-supported VESA modes, might you know where they are listed?

No idea yet.  According to the QEMU source tree, QEMU's ROM BIOS images (including the VGA option ROM) actually come from other projects.

Thank you!

paulwratt:
> For some reason, the OEM mode numbers, with mode numbers ≤ 0xff, are not returned in the mode list.

those should be remaped into the list you got. I believe there are 2 different ways to get "a mode list", one shows both, the other (I think) is the currently recommended way, which maps some modes below 0xff above 0xff (I think thats what I found - I _had_ a VESA 3.2? Voodoo 5000 card(s) when I wrote the Vesa Switcher, but the other hardware I had was VESA 2.0 and the Vesa VBE Scitech (interposer) Driver)

jart:
> The problem is fixed by replacing sed by sed.com in realify.sh. Kind of amazing, that an APE binary works better than the macOS standard version, huh?! I am seeing the wisdom now in having tools included in a standalone repo for a project of this complexity.

:-)

tkchia:
Hello @ghaerr, hello @jart, hello @paulwratt,

> If vesaspiral.c were modified to take an argument that enumerated modes and displayed them, it could be interesting to run on a variety of systems to see what kinds of (automatic VESA) resolutions might be available.

I have been tinkering with the possibility of adding this video mode enumeration functionality to Cosmopolitan `libc` itself.  What I am thinking is this: if the user presses a particular magic key &mdash; e.g. one of the Shift keys &mdash; at boot time, the real mode code can see this magic key, and display a list of all the supported VESA video modes.

(Perhaps this magic key can bring up further debugging functionality too &mdash; but that is for the future.)

I have got a mostly working implementation of this, but I hope to clean up the code first.

Thank you!

ghaerr:
Hello @tkchia,

> if the user presses a particular magic key — e.g. one of the Shift keys — at boot time, the real mode code can see this magic key, and display a list of all the supported VESA video modes.

Wow, impressive! So this would be similar to the way(s) that Linux can be booted, where a given graphics mode (or also kernel image) can be selected. If we end up going down the route of actually running graphics (consoles at least) on bare metal, this would come in super handy for both selectability by differing users/hardware as well as testing. Back years ago when I was developing the Microwindows framebuffer code, this capability came in very handy for testing to support a variety of graphics resolutions (and pixel formats).

Thank you!!

PR #619, opened 2022-09-14
add getgroups and setgroups

G4Vi:
Also functions that returned a `gid` or `uid` were changed to use an `unsigned` or the respective `_t` type from `int`. (Removes the need to overriding their type in Perl `Configure`) and makes the prototypes of the functions using the `id` types consistent.


libc/calls/getgroups.h
```diff
@@ -0,0 +1,11 @@
+#ifndef COSMOPOLITAN_LIBC_CALLS_GETGROUPS_H_
+#define COSMOPOLITAN_LIBC_CALLS_GETGROUPS_H_
+#include "libc/calls/weirdtypes.h"
+#if !(__ASSEMBLER__ + __LINKER__ + 0)
+COSMOPOLITAN_C_START_
+
+int getgroups(int size, gid_t list[]);
```
jart, line 7: This could go in calls.h if you use a uint32_t

libc/intrin/describegidlist.c
```diff
@@ -0,0 +1,56 @@
...
+  if ((!IsAsan() && kisdangerous(list)) ||
+      (IsAsan() && !__asan_is_valid(list, size * sizeof(list[0])))) {
+    ksnprintf(buf, N, "%p", list);
+    return buf;
+  }
+  buf[0] = '{';
+  unsigned printleft = N - 1;
+  unsigned i;
+  for (i = 0; i < size; i++) {
+    printleft -= ksnprintf(&buf[N - printleft], printleft, "%u, ", list[i]);
```
jart, line 42: Here's a design pattern I think will work better for ksnprintf() https://github.com/jart/cosmopolitan/blob/master/libc/intrin/describentoverlapped.c

jart (changes requested):
Looks pretty good. Some suggestions. Thank you for sending this!

libc/calls/getegid.c
```diff
@@ -17,22 +17,23 @@
...
+#include "libc/intrin/strace.internal.h"
 #include "libc/runtime/runtime.h"
 
 /**
  * Returns effective group ID of calling process.
  * @return group id
  */
-int getegid(void) {
-  int rc;
+gid_t getegid(void) {
```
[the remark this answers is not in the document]
G4Vi: I switched them to `uint32_t`.

test/libc/calls/getgroups_test.c
```diff
@@ -0,0 +1,47 @@
...
+  }
+}
+
+TEST(setgroups, test) {
+  int rc;
+  uint32_t src[5];
+  EXPECT_EQ(-1, setgroups(0, NULL));
+  if (IsLinux() || IsNetbsd() || IsOpenbsd() || IsFreebsd() || IsXnu()) {
+    EXPECT_EQ(EPERM, errno);
+    EXPECT_EQ(-1, setgroups(sizeof(src) / sizeof(src[0]), src));
```
jart, line 43: It's possible to say:

```c
EXPECT_SYS(EPERM, -1, setgroups(sizeof(src) / sizeof(src[0]), src));
```

The nice thing about this macro is it's able to test that nothing clobbers `errno`.

There's also an `ARRAYLEN()` macro for this situation.

jart, line 43: It'll be interesting to see what happens with this on non-linux where I'm not sure the `setgroups(0, NULL)` drop privilege thing exists. `setgroups()` isn't part of POSIX I think since I don't see a definition for it there.

jart (approved):
Thank you!

PR #620, opened 2022-09-14
fix MODE=dbg build

G4Vi:
https://github.com/jart/cosmopolitan/commit/6f7d0cb1c3962f7cb474b04de75df7cd82a3e5d3 broke `MODE=dbg` builds due to missing declarations in `third_party/python/launch.c`


jart (approved):
Thanks!

PR #621, opened 2022-09-15
Allow optional double quotes in dd parameters on macOS

pkulchenko:
This fixes an issue with ape.S using skip/count parameters without doublequotes on macOS thus breaking macho payload parsing.


jart (approved):
Thank you!

PR #622, opened 2022-09-15
Bare metal VGA: fix mapping of ECMA-48 colors to VGA colors

tkchia:
See https://github.com/jart/cosmopolitan/pull/613#issuecomment-1247490693


jart (approved):
Thank you!

jart (approved):
Approved. Sorry for the latency. I've been fending off some DDOS attacks the past few days.

PR #624, opened 2022-09-15
Bare metal: fix corner case in `__normalize_e820( )`

tkchia:
See https://github.com/jart/cosmopolitan/pull/609#issuecomment-1247083243


jart (approved):
Thank you!

PR #625, opened 2022-09-16
Bare metal: export structure member offsets from C code, rather than compute them by hand

tkchia:


paulwratt:
thats a nice set of changes @tkchia 

I know the comments in `[libc/runtime/mman.internal.h](https://github.com/jart/cosmopolitan/pull/625/commits/540b4e441b8adb9d3e8871b0ba2263514d0c6cd2#diff-f2853235b29b4ff806f8c80d0a89acb4c56accdb9fd067360e5ec959bc3ed41a)` are just for reference, but are they still accurate now, seems like the line with the add struct duplicates a reference

ghaerr:
Hello @tkchia,

Very elegant fix for getting rid of hard-coded structure offsets in ASM, I may have to borrow that trick for some other projects :) Also nice catch on the `short` alignment offset on the unpacked `mman` struct. I can see this fix is very much needed, as the wrong addresses have been scribbled on for a while now!

Perhaps `export_offsetof` should be moved upwards to a system-wide .internal.h file, given how useful it is.

Thank you!

ape/ape.S
```diff
@@ -328,7 +329,7 @@ dsknfo:	push	%bx
 	movpp	%es,%ds
 	xor	%si,%si
 	mov	%si,%es
-	mov	$0x1d10,%si		# mman::pc_drive_base_table
+	mov	$mm+"struct mman::pc_drive_base_table",%si
```
jart, line 332: Wuuuuut

libc/intrin/mman.greg.c
```diff
@@ -128,7 +128,26 @@ static noasan textreal void __invert_memory(struct mman *mm, uint64_t *pml4t) {
   }
 }
 
+/**
+ * Exports information about the offset of a field within a structure type,
+ * so that assembly language routines can use it.  This macro can be invoked
+ * from inside a function whose code is known to be emitted.
+ */
+#define export_offsetof(type, member)                       \
```
jart, line 136: This is great; I'm loving it!

jart (approved):
LGTM

PR #630, opened 2022-09-20
accept: switch 2nd param to take struct sockaddr *

G4Vi:
Fixes perl's `Configure` last arg of `accept` detection.

`accept4` was left alone as a change was not needed.


jart (approved):
Thank you!

PR #631, opened 2022-09-20
Bare metal VGA: pass video mode information via `struct mman`

tkchia:
This should hopefully help us adapt the VGA console code, to work with video modes other than 80 × 25 × 16 text.

### DIFF
--- a/libc/intrin/mman.greg.c
+++ b/libc/intrin/mman.greg.c
@@ -148,6 +148,12 @@ noasan textreal void __setup_mman(struct mman *mm, uint64_t *pml4t) {
   export_offsetof(struct mman, e820);
   export_offsetof(struct mman, e820_end);
   export_offsetof(struct mman, bad_idt);
+  export_offsetof(struct mman, pc_video_type);
+  export_offsetof(struct mman, pc_video_scan_width);
+  export_offsetof(struct mman, pc_video_width);
+  export_offsetof(struct mman, pc_video_height);
+  export_offsetof(struct mman, pc_video_framebuffer);
+  export_offsetof(struct mman, pc_video_framebuffer_size);
   __normalize_e820(mm);
   __invert_memory(mm, pml4t);
 }
--- a/libc/runtime/mman.internal.h
+++ b/libc/runtime/mman.internal.h
@@ -18,8 +18,25 @@ struct mman {
   unsigned char pc_drive_last_head;      /* 0x1d21 */
   unsigned char pc_drive;                /* 0x1d22 */
   char bad_idt[6];                       /* 0x1d23 */
+  unsigned char pc_video_type;           /* 0x1d29 */
+  unsigned short pc_video_scan_width;    /* 0x1d2a — line width including any
+                                                     invisible "pixels" */
+  unsigned short pc_video_width;         /* 0x1d2c — width in chars. (text)
+                                                     or pixels (graphics) */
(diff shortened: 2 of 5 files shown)


ghaerr:
Hello @tkchia,

Great idea getting a structure defined for passing up values determined in real mode up to long mode. My comments are that we should probably start now with a slightly different naming scheme between fields required for glyphs (which define the overall text screen size e.g. 80x25 in both text and graphics modes), and fields used for text-only or graphics-only use. This will make things easier in the future, since we won't get confused between the differing units used in each.

Possible suggestions for naming:
```
text_width, text_height: number of columns and rows of text displayed on console
char_width, char_height: width and height of each displayed character in pixels (graphics only)
gr_width, gr_height: width and height of graphics (frame) buffer in pixels
fb_stride: width in bytes of each frame buffer row, including pad bytes to get to next row
fb_width: width in bytes of each row (corresponding to pixels times frame buffer pixel format size)
fb_addr: frame buffer address
fb_size: frame buffer size in bytes (used for possible mmap, so may be larger than fb_stride times gr_height)
fb_format: pixel format
```
Keeping the frame buffer width and stride in bytes helps to remove subsequent calculations/multiplies, and separating console rows/cols  (width/height) from framebuffer width/height in pixels, since the graphics console will still have to track character rows and columns separately. It isn't clear that the `fb_` fields really need to be filled in for text mode operation, although doing so could result in allowing various additional text modes to run with little modification to the VGA console.

I would also suggest changing BGRX5551 to BGR555, since there are no frame buffers that don't use bit 0 (but there are some which use bit 15, e.g BGR1555).

Thank you!

paulwratt:
that should also make it easier to work with HDMI monitors using a frambuffer too

tkchia:
Hello @ghaerr,

The `tty.c` code &mdash; following `pty.c` &mdash; currently uses field names `xn` and `yn` for the (text) width and height.  Possibly we can continue using this scheme of using short names (?).  But of course, any new field names should really be documented properly.

Thank you!

ghaerr:
Hello @tkchia,

> Possibly we can continue using this scheme of using short names

Sounds good to me, I also prefer shorter rather than longer variable names, as long as there is no confusion. My previous suggestions were just to try to sketch out a possible scenario for how the graphics routines might interface with a framebuffer and eventually a graphics console. Ultimately, a graphics console will mostly track unicode code points on an x,y plane and likely pass an opaque `struct framebuffer *` to any of the required graphics drawing routines (fillrect, draw text, and scroll).

Thank you!

tkchia:
Hello @ghaerr, hello @paulwratt,

On second thought, I think I would like to do some additional groundwork first before getting this patch applied.  I will close this PR for now and possibly re-submit it later.  Thank you!

tkchia:
Hello all,

I have re-submitted this patch, as part of PR https://github.com/jart/cosmopolitan/pull/636 .

Hello @paulwratt,

> that should also make it easier to work with HDMI monitors using a frambuffer too

Well, there is the the Graphics Output Protocol when one boots up in UEFI mode, I suppose. :slightly_smiling_face:  It will be interesting to get a frame buffer console working there.

Thank you!

paulwratt:
it is a shame I dont have x86_64 SBC to test this - the HDMI monitor standard has a default "preferred mode" which is different per monitor, and the same brand with the came circuit, but a larger (and higher res) LCD/etc can have different properties (ie 0 or -20 borders - overscan) .. but I digress ..

PR #634, opened 2022-09-22
fix vista via revert

G4Vi:
It appears Windows 7 support actually broke earlier in 3265324e00634e8e9d659ba705da5b0ce88e78ef according to my new bisect testing `o/examples/hello.com` on Windows 7.

This is just a squashed commit of reverting to before it.


JoshuaWierenga:
In case it is relevant I tried to find the exact issue in that commit and it appears to be the use of memory location `0x6fe000040000` in [InitializeFileDescriptors](https://github.com/jart/cosmopolitan/blob/3265324e00634e8e9d659ba705da5b0ce88e78ef/libc/intrin/g_fds.c#L46) causing an access violation, prior to this commit, an array(`fds->__init_p`) was used that appears to be statically allocated near the end of the binary. I have manually applied almost every other change in that commit including every else in that file and a sample application still works on vista.

G4Vi:
@JoshuaWierenga Good find. Having the vista branch contain as many tested changes as possible would be best. I've marked this as a draft as that sounds like a better fix. Maybe make a new PR with those changes, then I could test Actually Portable Perl on Windows 7 with it, and if that goes well I'd give the thumbs up for merging that.

JoshuaWierenga:
@G4Vi I am working on that, I made the mistake of fixing 3265324e00634e8e9d659ba705da5b0ce88e78ef directly so currently unable to push, currently recreating changes on top of the vista branch. Hopefully, nothing else changed that causes issues. Will know soon enough.

JoshuaWierenga:
@G4Vi Done! Had no issues moving changes forward to the newest commit on the Vista branch. Just opened the issue #635.

PR #635, opened 2022-09-22
Get cosmopolitation working on Vista again

JoshuaWierenga:
Ideally the fixed size array would be removed on Vista/7 as well but I am not entirely sure how. If it's truly related to older versions of Windows having smaller virtual address spaces then perhaps using a smaller address would be enough.

### DIFF
--- a/libc/intrin/g_fds.c
+++ b/libc/intrin/g_fds.c
@@ -22,6 +22,7 @@
 #include "libc/intrin/pushpop.h"
 #include "libc/intrin/weaken.h"
 #include "libc/nt/runtime.h"
+#include "libc/runtime/memtrack.internal.h"
 #include "libc/sysv/consts/o.h"
 #include "libc/thread/thread.h"
 
@@ -43,10 +44,11 @@ textstartup void InitializeFileDescriptors(void) {
   struct Fds *fds;
   __fds_lock_obj._type = PTHREAD_MUTEX_RECURSIVE;
   fds = VEIL("r", &g_fds);
-  fds->p = fds->e = (void *)0x6fe000040000;
+  fds->p = fds->e = (void *)kMemtrackFdsStart;
   fds->n = 4;
   fds->f = 3;
-  fds->e = _extend(fds->p, fds->n * sizeof(*fds->p), fds->e, 0x6ff000000000);
+  fds->e = _extend(fds->p, fds->n * sizeof(*fds->p), fds->e,
+                   kMemtrackFdsStart + kMemtrackFdsSize);;
   if (IsMetal()) {
     extern const char vga_console[];
     pushmov(&fds->f, 3ull);
--- a/libc/calls/reservefd.c
+++ b/libc/calls/reservefd.c
@@ -50,8 +50,8 @@ int __ensurefds_unlocked(int fd) {
   bool relocate;
   if (fd < g_fds.n) return fd;
   g_fds.n = fd + 1;
-  g_fds.e =
-      _extend(g_fds.p, g_fds.n * sizeof(*g_fds.p), g_fds.e, 0x6ff000000000);
+  g_fds.e = _extend(g_fds.p, g_fds.n * sizeof(*g_fds.p), g_fds.e,
+                    kMemtrackFdsStart + kMemtrackFdsSize);
   return fd;
 }
 
--- a/libc/nt/version.h
+++ b/libc/nt/version.h
@@ -9,11 +9,21 @@ bool32 GetVersionEx(struct NtOsVersionInfo *lpVersionInformation);
 
 #if defined(__GCC_ASM_FLAG_OUTPUTS__) && !defined(__STRICT_ANSI__)
 #define IsAtLeastWindows10() (GetNtMajorVersion() >= 10)
+#define IsAtleastWindows8p1() \
+  (GetNtMajorVersion() > 6 || (GetNtMajorVersion() == 6 && GetNtMinorVersion() == 3))
 #define GetNtMajorVersion()    \
   ({                           \
     uintptr_t __x;             \
     asm("mov\t%%gs:96,%q0\r\n" \
         "mov\t280(%q0),%b0"    \
+        : "=q"(__x));          \
+    (unsigned char)__x;        \
+  })
+#define GetNtMinorVersion()    \
+  ({                           \
+    uintptr_t __x;             \
+    asm("mov\t%%gs:96,%q0\r\n" \
+        "mov\t284(%q0),%b0"    \
         : "=q"(__x));          \
     (unsigned char)__x;        \
   })
--- a/libc/runtime/memtrack.internal.h
+++ b/libc/runtime/memtrack.internal.h
@@ -7,7 +7,6 @@
 #include "libc/macros.internal.h"
 #include "libc/thread/tls.h"
 #include "libc/nt/version.h"
-#include "libc/nt/enum/version.h"
 #include "libc/runtime/stack.h"
 #include "libc/sysv/consts/ss.h"
 #if !(__ASSEMBLER__ + __LINKER__ + 0)
@@ -39,7 +38,7 @@ COSMOPOLITAN_C_START_
 #define _kMem(NORMAL, WIN7) \
   (!IsWindows() || IsAtLeastWindows10() ? NORMAL : WIN7)
 #define _kMemVista(NORMAL, WINVISTA) \
-  (!IsWindows() || NtGetVersion() >= kNtVersionWindows81 ? NORMAL : WINVISTA)
+  (!IsWindows() || IsAtleastWindows8p1() ? NORMAL : WINVISTA)
 
 struct MemoryInterval {
   int x;
--- a/libc/runtime/memtrack.internal.h
+++ b/libc/runtime/memtrack.internal.h
@@ -32,6 +32,9 @@ COSMOPOLITAN_C_START_
 #define kMemtrackZiposStart _kMemVista(0x6fd000040000, 0x5d000040000)
 #define kMemtrackZiposSize \
   (_kMemVista(0x6fdffffc0000, 0x5dffffc0000) - kMemtrackZiposStart)
+#define kMemtrackNsyncStart _kMemVista(0x6fc000040000, 0x5c000040000)
+#define kMemtrackNsyncSize \
+  (_kMemVista(0x6fcffffc0000, 0x5cffffc0000 - kMemtrackNsyncStart)
 #define _kMmi(VSPACE)                                                   \
   ROUNDUP(VSPACE / FRAMESIZE * (intptr_t)sizeof(struct MemoryInterval), \
           FRAMESIZE)
--- a/libc/runtime/memtrack.internal.h
+++ b/libc/runtime/memtrack.internal.h
@@ -7,6 +7,7 @@
 #include "libc/macros.internal.h"
 #include "libc/thread/tls.h"
 #include "libc/nt/version.h"
+#include "libc/nt/enum/version.h"
 #include "libc/runtime/stack.h"
 #include "libc/sysv/consts/ss.h"
 #if !(__ASSEMBLER__ + __LINKER__ + 0)
@@ -26,17 +27,19 @@ COSMOPOLITAN_C_START_
 #define kFixedmapStart _kMem(0x300000000000, 0x000040000000)
 #define kFixedmapSize \
   _kMem(0x400000000000 - 0x300000000000, 0x000070000000 - 0x000040000000)
-#define kMemtrackFdsStart \
-  (ROUNDDOWN(_kMem(0x6fe000000000, 0x68000000), FRAMESIZE * 8) - 0x8000 * 8)
-#define kMemtrackFdsSize _kMem(0x001000000000, 0x04000000)
-#define kMemtrackZiposStart \
-  (ROUNDDOWN(_kMem(0x6fd000000000, 0x6c000000), FRAMESIZE * 8) - 0x8000 * 8)
-#define kMemtrackZiposSize _kMem(0x001000000000, 0x10000000)
+#define kMemtrackFdsStart _kMemVista(0x6fe000040000, 0x5e000040000)
+#define kMemtrackFdsSize  \
+  (_kMemVista(0x6feffffc0000, 0x5effffc0000) - kMemtrackFdsStart)
+#define kMemtrackZiposStart _kMemVista(0x6fd000040000, 0x5d000040000)
+#define kMemtrackZiposSize \
+  (_kMemVista(0x6fdffffc0000, 0x5dffffc0000) - kMemtrackZiposStart)
 #define _kMmi(VSPACE)                                                   \
   ROUNDUP(VSPACE / FRAMESIZE * (intptr_t)sizeof(struct MemoryInterval), \
           FRAMESIZE)
 #define _kMem(NORMAL, WIN7) \
   (!IsWindows() || IsAtLeastWindows10() ? NORMAL : WIN7)
+#define _kMemVista(NORMAL, WINVISTA) \
+  (!IsWindows() || NtGetVersion() >= kNtVersionWindows81 ? NORMAL : WINVISTA)
 
 struct MemoryInterval {
   int x;
--- a/libc/zipos/open.c
+++ b/libc/zipos/open.c
@@ -52,9 +52,10 @@ static void *__zipos_mmap(size_t mapsize) {
   assert(mapsize);
   offset = maptotal;
   maptotal += mapsize;
-  start = (char *)0x6fd000040000;
+  start = (char *)kMemtrackZiposStart;
   if (!mapend) mapend = start;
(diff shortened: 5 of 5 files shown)


G4Vi:
I'll test this tonight and report back.

G4Vi:
:+1:  works for me on Windows 7.

JoshuaWierenga:
I am struggling to even get perl to build, issues appear related to the groups commit since the vista branch doesn't have it and I have so far been unable to cherry-pick it in.

paulwratt:
keep try .. (do it by hand if you have too :)

JoshuaWierenga:
The best option I can think of would be some kind of automated testing on vista/7, then I can just start at the commit this branch tries to fix and then work forward until the groups commit is included, fixing any additional commits as required. The issue then is that there are some commits on the vista branch that don't actually appear on the main branch

G4Vi:
> I am struggling to even get perl to build, issues appear related to the groups commit since the vista branch doesn't have it and I have so far been unable to cherry-pick it in.

My apologies for incomplete instructions, For vista builds right now uncomment everything after the comment
 `# uncomment as needed for older cosmopolitan` in `perl5/hints/cosmo.sh`  Once we resolve the vista builds in general issue I'm hoping to merge the groups stuff and such so that manual patch is no longer needed.

JoshuaWierenga:
I moved from wsl2 to a ubuntu VM and it appears to have built without modification to either it or my cosmo branch, though I did have some issues with not having a full suite of build tools available and so it couldn't find message queue functions.

JoshuaWierenga:
I get `The application failed to initialize properly (0xc0000005)`, would that be related to what you got or because I didn't make the vista changes?

G4Vi:
> I get `The application failed to initialize properly (0xc0000005)`, would that be related to what you got or because I didn't make the vista changes?

It definitely means `perl.com` is corrupt, but I don't believe that error code and dialog came up on me for the issue I described. could you please clarify what version of cosmopolitan you used and what command you ran in cmd.exe to produce it.

JoshuaWierenga:
I used the cosmo version from this pull and ran perl.com without parameters which I assume should have just done nothing?

G4Vi:
> I used the cosmo version from this pull and ran perl.com without parameters which I assume should have just done nothing?

`perl.com` with no parameters should just attempt to read a program from stdin, Ctrl-C should close it. Typing `print "hello\n";`  before the Ctrl-C should print hello to the terminal when perl is closed.

Interesting it seems like there's intermittent behavior, for me when I  built `perl.com` from this pull it worked for printing hello, so I didn't discover the failure until attempting to run perldoc in the command above.

JoshuaWierenga:
image
So this should have worked?

JoshuaWierenga:
That is with the hint changes you suggested. No change on vista with those changes.

G4Vi:
> image So this should have worked?

Ctrl-D on most shells. Ctrl-C in cmd.exe

JoshuaWierenga:
Right, so it works fine on linux then.

JoshuaWierenga:
Applying this patch directly to the offending commit and ignoring all the newer ones does allow basic use of perl to work on vista. 
image

G4Vi:
Does `perl.com /zip/usr/local/bin/perldoc5.36.0` print the usage message on vista?

G4Vi:
> I can now try stuff like `perldoc` but I have no idea where it is as I don't have either `/zip` or `./zip`. Where can I get a copy?

`/zip` is a magic path that refers the executable/zip file itself. `perldoc5.36.0` should already be embedded inside of `perl.com`

JoshuaWierenga:
I just figured that out, cool. Nope, it quits without output on vista but runs fine on linux. Is progress though as I have reproduced your error.

JoshuaWierenga:
I ran it under windbg and get Illegal instruction:
image

JoshuaWierenga:
I suspect its the other __init_p change in that commit causing issues,
perhaps need to special case that for windows as well. I will try that and
removing the unrelated changes and see what happens.

JoshuaWierenga:
Hmm, it's running on vista right now but I have no clue which version it is, there is a small chance it is working with some changes I made a few hours ago. 
image

JoshuaWierenga:
I think I found the second bug in that commit, zipos is also modified to use a much higher memory address. Thats the fancy `/zip` internal archive right? Would explain why my patch fixes simple applications but not perl.

G4Vi:
Nice, yes, zipos implements the internal zip archive. If perldoc is now working, it might be worth testing cosmo's redbean web server `make -j8 o//tool/net && o//tool/net/redbean.com` for another program that uses zipos and a lot of cosmo in general.

JoshuaWierenga:
So perhaps I was right about the zipos changes, _extend was new in this commit. 
image

JoshuaWierenga:
@G4Vi I am not that happy with the patch as it stands currently as I have just restored the old functionality on windows but got a wip patch that actually works as far as I know. https://github.com/JoshuaWierenga/cosmopolitan/tree/vistafix

G4Vi:
Thanks for the update. Anything that's not working should probably be tested on c3208eb9d5d39f270be3d0c9dce39badde593212 as with the lack of testing on vista/windows 7 there's a chance more things broke or never worked before the known problematic commits.

JoshuaWierenga:
Definitely want to work on windows testing, any chance that part of cosmo works as is on windows and it's just the library itself that doesn't work, even with the provided gcc?

G4Vi:
@jart Before Windows 7 bugged out, how were you doing the testing?

Also @JoshuaWierenga I'm not sure if you're aware of it, but there's a cosmopolitan discord server. I've generated a temporary invite only good for an hour: https://discord.gg/f2sqaZN

JoshuaWierenga:
I am not sure what fix I was even looking for. The problem was already mentioned by jart on the original issue for the vista break which is that windows 8.0 and older(yes it's not just vista and 7!) on x64 cap the user virtual address space to 8tb instead of 128tb. Just detecting older versions of windows, using smaller memory values for zipos and fd, and hoping there is no conflict appears the best option. Though I have no option other than to conflict with asan provided [memtrack](https://github.com/jart/cosmopolitan/blob/master/libc/runtime/memtrack64.txt) is correct since it covers the entire valid range. Looks like there is some other stuff there I may need to move as well.

JoshuaWierenga:
Branch with working patch for 3265324e00634e8e9d659ba705da5b0ce88e78ef is at https://github.com/JoshuaWierenga/cosmopolitan/tree/vistafix2.

JoshuaWierenga:
Not at all relevant since there was never any real claim for support but I did try xp x64 edition because why not. Sadly it didn't work, even for simple applications. It's not to do with anything changed here because release 2.0.1 doesn't work either.

libc/intrin/g_fds.c
```diff
@@ -43,29 +43,37 @@ textstartup void InitializeFileDescriptors(void) {
...
+    // Use old fixed size descriptor array
+    // TODO find way to use dynamic array to prevent potental race condition like on other OSes
+    fds->p = fds->__win_init_p;
+
     SetupWinStd(fds, 0, kNtStdInputHandle);
     SetupWinStd(fds, 1, kNtStdOutputHandle);
     SetupWinStd(fds, 2, kNtStdErrorHandle);
+  } else {
+    fds->p = fds->e = (void *)0x6fe000040000;
+    fds->e = _extend(fds->p, fds->n * sizeof(*fds->p), fds->e, 0x6ff000000000);
```
jart, line 59: I'm surprised _extend ended up in the vista branch because I didn't intend to have it in there. We might want to recut the branch or simply revert the change that added `_extend`

JoshuaWierenga: I will update this pull when I have a chance as I have a working version with _extend at [vistafix2](https://github.com/JoshuaWierenga/cosmopolitan/tree/vistafix2). The only issue is it uses memory assigned to asan on windows 8 and older, at least according to [memtrack64](https://github.com/jart/cosmopolitan/blob/master/libc/runtime/memtrack64.txt). Issues to consider with that fix include whether or not the address is actually stored somewhere to avoid dynamically allocating something else to it and if the platform check substantially impacts the performance and binary size.

It would also be useful to have some testing on vista/7/8, even if it's just something I set up locally, how is cross-platform testing handled typically? That would help ensure my fix works and there are no issues in newer commits since I have just been working on top of the _extend commit directly.

JoshuaWierenga: If the patch doesn't totally work and isn't reasonably fixable, or the performance/size is adversely impacted then I can go back to the previous commit. However, I see no reason to drop functionality that works fine on other platforms if not absolutely necessary.

jart: If you want to keep _extend() then all you need to do is just pick a 32-bit address to use when it's Windows 7 or lower. There's a couple text files in libc/runtime/ where you can document your decisions. You should see in memtrack.internal.h the ternary statements used in the `#define` that special case the hard-coded addresses. We just need that to happen for the new code that's using the _extend() API.

Let me know as soon as you think it's ready for merge, and I'll merge it. I can also do testing legwork for you too, where I run your change on my test fleet to make sure there aren't any regressions with other operating systems.

Thanks so much!

jart:
> @jart Before Windows 7 bugged out, how were you doing the testing?

I describe my testing process in this hacker news comment https://news.ycombinator.com/item?id=32921983 It's very hard for contributors to replicate my test fleet locally. Only one contributor has done it so far. I'm actively seeking grants to make it part of our GitHub Actions process. Until that happens, just ping me on Discord and ask me to clone your PR and test it for you. I'll do that for you any time, any day. I just can't do it for Windows 7 and Vista right now because my Win7 VM croaked and I don't know how to get another one up and running in a way that respects Microsoft's IP. 

So to test on Win7 and Vista, all you have to do is build the .com and .com.dbg files, `scp` them both to your Windows machine, and run them. You'll get a sub-par backtrace as you saw, which gives you hard coded addresses rather than file:line numbers. The way you turn the fixed code addresses into a line of code, is you go back to your Linux build machine, and run `objdump -Sd foo.com.dbg` and search for the address. That'll give you the code location.

jart:
> windows 8.0 and older(yes it's not just vista and 7!) on x64 cap the user virtual address space to 8tb instead of 128tb

That's great to know! What about Windows 8.1? Does that support the full 48-bit address space?

The fast path approach to testing the Windows version right now is in `libc/nt/version.h` which currently only does the major version, grabbed directly from the PEB. 

https://github.com/jart/cosmopolitan/blob/ecb2ef7c39e6296d6b1e1c5525b5d3d48e34d66b/libc/nt/version.h#L10-L20

JoshuaWierenga:
@jart I found https://learn.microsoft.com/en-us/windows/win32/memory/memory-limits-for-windows-releases#memory-and-address-space-limits which says that its 8tb on x64 for windows 8.0/server 2012 and older and 128tb for windows 8.1/server 2012 R2 and newer.

I will hopefully have time later today to update this pull, just need to clean up the existing changes from [vistafix2](https://github.com/JoshuaWierenga/cosmopolitan/tree/vistafix2) and include all commits from after the _extend one, including the one with the _extend constant changes in memtrack.internal.h.

JoshuaWierenga:
Just had a quick look at commits in both the vista and master branches and made some notes, mostly for when I try to merge some changes. Some of this might end up being a different pull since the changes are not required to make vista/7/8 work again.

Starting at c3208eb9d5d39f270be3d0c9dce39badde593212 which runs on vista and up.
1. 3265324e00634e8e9d659ba705da5b0ce88e78ef is already on master and vista and needs already made patch.
2. e9272f03fbdadee8ced8670514ec5f987c47bbd2 through 0305194d9854556e2bbef1fdcfda6e76910b1baf are already on master and vista.
3. b69f3d2488dbaf9dcc541e699f5b7c09fbf046e0 through 9cf2bdc9a806409055bd4d624dc2fa5b03090729 on master are missing on vista.
4. Need to keep c03359c637293171b7d62ba1c7efee56e82b7538 from vista branch.
5. e0fabd1d497fb3ffcbdb674667812aaf26d84261 through 446a1f7c5aac1b110439c429d24c69261af49f95 are already on master and vista.
6. 06f9a5b627176675552a75274a1a2748c0cc6804 on master is missing on vista.
7. 9fa672506738cd210cb819e3fd05d2dd7d3f2997 is on master and vista but vista version(827f25f054b577e6f1a90dcddfbc9ffd3f6c4f53) reverts part of 446a1f7c5aac1b110439c429d24c69261af49f95.
8. 8f5678882d3a2ea15e3011847f24bc29574c28cd and up are missing because of vista split, except for 3f4988984183dae5bf8a39b9fcbb8cdf78ce10c7 which partially exists, see below.
9. 050062bcbb3d21899299e07ccf8acef92755bf3f, a5fa90a21f911e33c17f5b624f4e6ae6aa6aabb4 and c64e7eaf7c73172948277a3cc002428ebb6f75c0 from vista add parts of 3f4988984183dae5bf8a39b9fcbb8cdf78ce10c7 from master with some differences: 
libc/assert.h has an additional assert macro change on master but not on vista.
libc/calls/fcntl-nt.c has some include changes on master but not on vista.
libc/calls/preadv.c and pwritev.c have some include changes on master but not on vista.
libc/nt/kernel32/SuspendThread.s was modified on vista but not on master.
libc/nt/master.sh has an additional due to SuspendThread.S on vista but not on master.
test/libc/calls/test.mk has an additional change(LIBC_NT_KERNEL32) on master but not on vista.
10. Need to keep 30140812f05162b40e9a2c6d21af2ab250b30542 from vista branch.

JoshuaWierenga:
Was not expecting it to close but guess it makes sense since I did remove my commit, will reopen in a moment.

JoshuaWierenga:
@jart @G4Vi Finally got a working version. Tested hello.com, redbean.com, perl interactive printing, and perldoc and all work just as well as vistafix2 did.

JoshuaWierenga:
So provided that works fully, that leaves documenting it in the memory text files in libc/runtime and perhaps finding a faster method of getting the nt point release.

JoshuaWierenga:
@jart Does it matter that the only place to put fd and zipos within the first 8tb overlaps with asan shadow memory in memtrack64.txt as it starts at near the bottom and goes way past 8tb. Is it also worth moving nsync as it's in the same block just before 112tib?

jart:
Yes nsync would also need to have its own planned address for vista.

If you want to support ASAN (I never bothered to do it with Vista in the past) then with an 8tb vaspace the shadow memory region would be on the interval [0x7fff0000,0x10080000000) and nothing else should exist at those addresses.

```
>>: hex(8*1024*1024*1024*1024)
'0x80000000000'
>>: (0x80000000000>>3)+0x7fff8000
1101659078656
>>: hex((0x80000000000>>3)+0x7fff8000)
'0x1007fff8000'
>>: hex((0>>3)+0x7fff8000)
'0x7fff8000'
```

JoshuaWierenga:
I am currently testing getting the nt minor version from peb on nt 6.x and it appears to work. 
I think for now I will just plan out a different memory map for old windows, won't actually move anything other than fd, zipos and perhaps nsync. Are the constants for nsync currently on the vista branch? I did see the commits adding nsync itself on there but not sure if I need to get a newer commit to actually set its address. Regardless, where is the shadow region set, memtrack as well?

JoshuaWierenga:
Is it worth making an 8tb version of memtrack64.txt? If so should I base it on the one in the vista branch or the current one? Not sure if it was updated just to reflect changes already made or for changes made only on master since the split.

JoshuaWierenga:
Might just include b69f3d2488dbaf9dcc541e699f5b7c09fbf046e0 since that is the only other commit with significant memory changes, half the stuff here is just from that anyway.

G4Vi:
I can test this branch tonight. I think it's a great idea to split up the changes to have the community supported vista branch working again sooner then later so it could be used as a base for groups backports and such.

G4Vi:
LGTM!

@JoshuaWierenga in the future it would be great if you could name your PR branch something different from the upstream branches, so that the first  `git checkout` isn't ambiguous of which one you're getting without specifying remote.

JoshuaWierenga:
Sorry about that. I found the same issue a new days ago and so tried to
rename the branch, sadly the pull doesn't update and hence the branch
deletion message half way up.

jart (approved):
Awesome. Thank you!

jart:
By the way, have you sent the copyright assignment intent email to me? See CONTRIBUTING.md. I'm searching my inbox and I'm not sure. If you haven't, could you please send it?

JoshuaWierenga:
@jart Just sent the email.

jart (approved):
Thanks!

JoshuaWierenga:
@paulwratt Thanks but it wasn't the commit itself that confused me but the reasoning. The changes near 112tb were to reflect changes in the commit before but there were way more changes than that and so I was wondering if they reflect other changes in that commit or if the file was just out of date for a while. I am sidestepping most of that with #646 as it just adds both of the memory commits so it shouldn't matter. I can't of course just modify that file, I need to find everything it describes and make changes there, jart sticking 4s everywhere does help since now the addresses are not just round numbers but still lots of stuff to figure out.

PR #636, opened 2022-09-22
Bare metal: tweaks pertaining to physical memory handling

tkchia:
- Factor `__invert_memory()` into two routines, to make it easier for other `libc` routines to create their own negative address mappings
- Tweak `__invert_memory()` to map only the usable memory as given in the `0xe820` memory map, and no more
- Arrange for VGA console initialization to map video memory separately into virtual memory (in case needed)
- Rename `libc/vga/vga-init.S` to `libc/vga/rlinit-vga.S`, to reduce confusion as to the module's purpose


tkchia:
And:
- Pass video mode information via `struct mman`

ghaerr:
Hello @tkchia,

Looks good, thanks for the changes to `pc_video_stride`, etc. This is shaping up nicely for combined text/graphics console enhancement.

> Arrange for VGA console initialization to map video memory separately into virtual memory (in case needed)

Nice. Since `pc_video_framebuffer` is going to be the physical address, we might want to document that in `mman.internal.h`, to avoid potential confusion, since `ape.S` may have already mapped that to the same VA in the case of a text console, but not graphics console. (Both text and graphics console will use `BANE+pc_video_framebuffer` for simplicity of addressing fixed physical memory, I presume).

> 	`movl	$8*(80*25*2),mm+"struct mman::pc_video_framebuffer_size`

Isn't the framebuffer size just 80\*25\*2? Why the further multiply by 8?

> Tweak __invert_memory() to map only the usable memory as given in the 0xe820 memory map, and no more

Your new code `if (ps < 0x200000 && ps + size <= 0x200000) ...` actually just prohibits a (repetitive, but not harmful) (re)map of the lower 2M of physical RAM, correct? That is, we're actually going to continue to ignore any BIOS-reported memory mappings below 2M and treat all that as linear physical (even if not present), instead of "map only the usable memory given in the 0xe820 map, and no more", or am I missing something?

Thank you!

tkchia:
Hello @ghaerr,

> Isn't the framebuffer size just 80*25*2? Why the further multiply by 8?

There are actually 8 "video pages" worth of 80 &times; 25 screens at `0xb8000`.  We really only use video page 0 for now, but I am thinking there is not much harm in also mapping the other 7 video pages into virtual memory.

(Actually my earlier calculation was not quite right.  It should not be 8 &times; (80 &times; 25 &times; 2) = 32,000 bytes, but 32,768 bytes.  Each page starts on a 2,048-byte boundary.  I have pushed a correction.)

Thank you!

tkchia:
Hello @ghaerr,

> That is, we're actually going to continue to ignore any BIOS-reported memory mappings below 2M and treat all that as linear physical (even if not present), instead of "map only the usable memory given in the 0xe820 map, and no more", or am I missing something?

The old code was not ignoring the existing `[BANE, BANE + 0x200000 - 1]` mapping.  Rather, for each `[start, end - 1]` physical memory area &mdash; including the one starting at physical address `0` &mdash; it was actually mapping an area `[start, end + 0x200000 - 1]` into virtual memory.  With this patch, we will only try to map `[start, end - 1]`.

Thank you!

tkchia:
Hello @ghaerr,

> Since `pc_video_framebuffer` is going to be the physical address, we might want to document that in `mman.internal.h`, to avoid potential confusion,

We probably should. :slightly_smiling_face: 

> since `ape.S` may have already mapped that to the same VA in the case of a text console, but not graphics console.

My understanding is that the `[0, 0x200000 - 1]` &rightarrow; `[0, 0x200000 - 1]` identity mapping only really exists at the very early startup stages.  The APE startup code actually invalidates this mapping later, once it has built up the final virtual memory mapping.
```
kernel:	movabs	$ape_stack_vaddr,%rsp
	add	$ape_stack_memsz,%rsp
	movl	$0,0x7b000			# unmap null 2mb
```

Thank you!

ghaerr:
Hello @tkchia,

> With this patch, we will only try to map [start, end - 1].
> The APE startup code actually invalidates this mapping later, once it has built up the final virtual memory mapping.

Thank you for the explanation and showing that code, I wasn't aware the original [0..2MB-1] VA mapping was unmapped, and got confused that the same BANE mapping occurred during `ape.S` startup (which it doesn't).

However, the BANE mappings only occur if the BIOS 0xe820 memory is reported as `kMemoryUsable`, right? Since prior to this PR, the text video memory for the VGA console was't specifically mapped, does that mean that the BIOS has always reported the video memory as `kMemoryUsable`? I was under the impression that classification was only for non-device memory that could be used by an OS for its own text/data, etc.

I have been thinking of writing a program that displayed the memory map, to easily see the entire memory map used on bare metal. It appears that @jart already wrote a nice `PrintMemoryIntervals` function in libc/runtime/printmaps.c, as well as a test program with a similar function in test/libc/runtime/memtrack_test.c. I will look into seeing how easily that might be converted to a .com file to be run on bare metal, as a utility, rather than test program.

Thank you!

tkchia:
Hello @ghaerr,

> Since prior to this PR, the text video memory for the VGA console was't specifically mapped, does that mean that the BIOS has always reported the video memory as `kMemoryUsable`?

Actually no &mdash; the video memory will probably be reported as "reserved", or it might simply be left out of the `0xe820` map altogether (per the ACPI specification &mdash; argh).

Meanwhile, it seems Cosmopolitan `libc` just decides to treat everything below 2 MiB as "normal" (and cacheable) memory.  But perhaps it should not.

> I will look into seeing how easily that might be converted to a .com file to be run on bare metal, as a utility, rather than test program.

:slightly_smiling_face: 

Thank you!

paulwratt:
Is there a configuration where that lower 2Mb _does not_ exist (or the ability to use it) - if so maybe a note of it somewhere (even if that is a different/docs thing/PR)

tkchia:
Hello @paulwratt, hello @ghaerr,

OK, I guess I spoke a bit too loosely &mdash; Cosmopolitan `libc` _does_ know that some parts of the first 2 MiB of memory are not "conventional" RAM, and will allocate RAM from this region correctly.

However, when mapping physical addresses to virtual addresses, the PAs `[0, 0x200000 - 1]` &mdash; including the ROM BIOS and whatnot &mdash; are mapped to VAs as if they are good old cacheable memory.  (This is mainly a problem for certain memory-mapped I/O devices, where the order in which one reads or writes to the "memory" matters a great lot.)

Thank you!

tkchia:
Subsuming the commits in this PR under PR https://github.com/jart/cosmopolitan/pull/637 .  Thanks!

PR #639, opened 2022-09-23
Add imaxdiv, wcscoll, getdtablesize

izissise:
Some added function implementation that are missing when compiling bash

### DIFF
--- a/test/libc/fmt/imaxdiv_test.c
+++ b/test/libc/fmt/imaxdiv_test.c
@@ -1,0 +1,65 @@
+/*-*- mode:c;indent-tabs-mode:nil;c-basic-offset:2;tab-width:8;coding:utf-8 -*-│
+│vi: set net ft=c ts=2 sts=2 sw=2 fenc=utf-8                                :vi│
+╞══════════════════════════════════════════════════════════════════════════════╡
+│ Copyright 2022 Hugues Morisset                                               │
+│                                                                              │
+│ Permission to use, copy, modify, and/or distribute this software for         │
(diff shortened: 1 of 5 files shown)


izissise:
Fixed for contributing guidelines

test/libc/fmt/imaxdiv_test.c
```diff
@@ -0,0 +1,61 @@
...
+
+  res = imaxdiv(0, -11);
+  EXPECT_EQ(0, res.quot);
+  EXPECT_EQ(0, res.rem);
+
+  res = imaxdiv(0, 11);
+  EXPECT_EQ(0, res.quot);
+  EXPECT_EQ(0, res.rem);
+
+  res = imaxdiv(INT64_MIN, -1);
```
jart, line 54: 1. `INTMAX_MAX`
2. Why isn't this raising an arithmetic exception? It should. 

```c
  volatile intmax_t i, j;
  i = INTMAX_MIN;
  j = -1;
  i = i / j;
  return 0;
}
```

jart (changes requested):
Thanks for sending this. Some review comments.

test/libc/fmt/imaxdiv_test.c
jart: Scratch that. According to GitHub Actions, it does raise SIGFPE. Good.

izissise: Oops I meant not to overflow, I fixed the test

jart (approved):
Looks good. Thank you so much!

PR #640, opened 2022-09-24
Bare metal: set up basic CPU exception handling, with IDT & TSS

tkchia:
When a CPU exception (e.g. protection fault, divide by zero) happens in bare metal mode, the program should now be able to catch the exception and dump some information on the exception to the serial console, if one ropes in (`STATIC_YOINK`) the symbol `_idt`.

I hope to get the exception handlers to dump information to the VGA screen as well.  It seems I will need to modify `kprintf` to do this.  Once this is implemented, it should be easier to diagnose any program crashes that occur on bare metal.

Thank you!

### DIFF
--- a/libc/intrin/interrupts.S
+++ b/libc/intrin/interrupts.S
@@ -1,0 +1,173 @@
+/*-*- mode:unix-assembly; indent-tabs-mode:t; tab-width:8; coding:utf-8     -*-│
+│vi: set et ft=asm ts=8 tw=8 fenc=utf-8                                     :vi│
+╞══════════════════════════════════════════════════════════════════════════════╡
+│ This is free and unencumbered software released into the public domain.      │
+│                                                                              │
+│ Anyone is free to copy, modify, publish, use, compile, sell, or              │
+│ distribute this software, either in source code form or as a compiled        │
+│ binary, for any purpose, commercial or non-commercial, and by any            │
+│ means.                                                                       │
+│                                                                              │
+│ In jurisdictions that recognize copyright laws, the author or authors        │
+│ of this software dedicate any and all copyright interest in the              │
+│ software to the public domain. We make this dedication for the benefit       │
+│ of the public at large and to the detriment of our heirs and                 │
+│ successors. We intend this dedication to be an overt act of                  │
+│ relinquishment in perpetuity of all present and future rights to this        │
+│ software under copyright law.                                                │
+│                                                                              │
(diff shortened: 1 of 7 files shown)


ghaerr:
Amazing work, @tkchia!

libc/intrin/interrupts.S
```diff
@@ -0,0 +1,173 @@
...
+	mov	%es,%eax
+	push	%rax
+	mov	$GDT_LONG_DATA,%eax	# ...& load ds, es, ss correctly
+	mov	%eax,%ss
+	mov	%eax,%ds
+	mov	%eax,%es
+	ezlea	.excep_msg,di		# stack should be 16-byte aligned now
+	xor	%eax,%eax		# kprintf is variadic; remember to
+					# pass no. of vector regs. used (= 0)
+	call	kprintf			# print error message
```
jart, line 94: Is this going to force a dependency on kprintf() into o/tiny/examples/life.com? If so, we might want to do a one-off error report.

libc/intrin/interrupts.S
```diff
@@ -0,0 +1,173 @@
...
+	lidt	_idtr			# load IDTR
+	ezlea	_idt,di
+	pushpop	32,%rcx			# fill IDT entries for CPU exceptions
+	ezlea	__excep0_isr,dx
+1:	mov	%edx,%eax
+	stosw
+	mov	%cs,%eax
+	stosw
+	mov	%rdx,%rax
+//		   ┌P:present
```
jart, line 126: I love how you did the CP437 diagrams.

jart (approved):
I can't say how impressed I am with this change. Doing this is something that's mystified me for quite some time.

libc/intrin/interrupts.S
tkchia: Hello @jart,

> Is this going to force a dependency on kprintf() into o/tiny/examples/life.com?

Not yet.  I am trying to arrange to link in the IDT logic only when the program is already linking in some sort of crash debugging (I am still not familiar with how Cosmopolitan's crash debugging capability works).

Thank you!

tkchia:
Hello @jart,

> Doing this is something that's mystified me for quite some time.

Well, it did take me quite a while to realize that I had to reload the GDTR to make it point to virtual memory &mdash; so that, when (re)loading `%cs` to service an interrupt, the CPU could properly load `%cs` 's descriptor from main memory, by way of the GDTR.  I am not sure if this was the problem you encountered.

Thank you!

PR #643, opened 2022-09-28
Make `ape.S` shell script magic less prone to accidental breakage

tkchia:
The metal boot sector code was wrapped in a single-quoted string, like so:
```
MZqFpD='
...MZ HEADER...
...E_LFANEW...
...BIOS BOOT SECTOR...'
```
This might break the shell code loader if the boot sector code contains a 0x27 (single quote) byte.  This patch wraps the boot sector code in a here-document instead:
```
MZqFpD='
...MZ HEADER...' <<'@'
...E_LFANEW...
...BIOS BOOT SECTOR...
@
```
This is harder to break — when interpreted as a shell script, the code can only accidentally terminate the here-document if there is a `"\n@\n"` sequence inside the BIOS boot portion.

### DIFF
--- a/ape/ape.S
+++ b/ape/ape.S
@@ -251,7 +251,6 @@ pc:	cld
 	xor	%cx,%cx				# current cylinder
 	xor	%dh,%dh				# current head
 	mov	$v_ape_realsectors,%di		# total sectors
-	sub	$v_ape_realslacksectors,%di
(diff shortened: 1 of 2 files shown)


jart (approved):
I've run into that problem a bunch of times. Clever solution to the problem! Thanks!

PR #645, opened 2022-10-02
Fix re.NOSUB memory error

stellartux:
This change fixes the following error.

```bash
[dan@mainpc cosmo]$ o/tool/net/redbean.com -i
>: re.search(" ", " ", re.NOSUB)
redbean: memory allocation error: block too big
stack traceback:
[C]: in function 're.search'
stdin:1: in main chunk
```


jart (approved):
Thank you!

PR #646, opened 2022-10-04
Fully support old windows memory restrictions

JoshuaWierenga:
As mentioned in #635, I want to fully support old windows's 8tib user virtual address space limit to reduce the chance of further breaks by ensuring anything using virtual memory has at least the option to only use addresses below 8tb. 

TODO:
- [x] Support asan? Not sure how much work this is or even where to start. Will likely postpone to a later pull, its not required to support 8tib virtual memory. I would three state this but looks like github md doesn't support that.
- [x] Further support old windows specific addresses
  - [x] ape/ape.lds has an assert mentioning 0x700000040000, is this important? It's labelled as stack in memtrack64.
  - [x] libc/runtime/brk.c has a comment mentioning APE doing allocations at 0x10008004000, is this important? Just automap?
  - [x] libc/runtime/memtrack.internal.h has addresses for automap, memtrack and fixedmap set above 8tib, these likely need to be moved.
  - [x] Figure out what _mmi is and where stack and winargs are setup.
  - [x] Figure out how the extra kernel memory at the upper end of memtrack64 works. Appears to be a hypothetical Unix thing and so irrelevant to Windows.
 - [x] Document any memory map changes in either libc/runtime/memtrack64.txt or a 8tib version of it. memtrack64-8tib.txt has everything in it and has a function to define addresses in terms of their 128tib counterparts which have been moved into a matching memtrack64-128tib.txt.
 - [ ] Add some form of automated testing.
 
Might be more things to consider, more substantial testing might pick them up but that might take a while. I can just do a simple search for 0x to find more examples in the short term.


JoshuaWierenga:
Looks like the builds failed during nsync, given the next commit adds more of it I will probably need that as well.

JoshuaWierenga:
Well I guess its time for 3f4988984183dae5bf8a39b9fcbb8cdf78ce10c7.

JoshuaWierenga:
Something(probably 149e4969ff946fe367f407cf94efe13ebe82892a) has broken old windows compat again, this might take a while.

JoshuaWierenga:
Had a quick go at an 8tib memory map, not finished yet. I think that's the right values for automap given new asan size and relationship between asan and automap position and size from the full map.

```
0x000000000000 to 0x00007ffbffff image stuff
0x00007ffe0000 to 0x00007ffeffff free
0x00007fff0000 to 0x01008003ffff asan
0x010080040000 to 0x01e7fffbffff automap
0x01e7fffc0000 to 0x01fffffbffff _mmi
0x01fffffc0000 to ?              free(kMemtrack block from memtrack.internal.h?)
?              to ?              fixedmap
?              to 0x05c00003ffff free
0x05c000040000 to 0x05cffffbffff nsync
0x05d000040000 to 0x05dffffbffff zipos
0x05e000040000 to 0x05effffbffff g_fds
0x05effffc0000 to ?              free
?              to ?              stack
?              to ?              winargs
?              to ?              free
?              to 0x07ffffffffff kernel?
```

JoshuaWierenga:
@jart Are there any specific requirements for memory addresses other 64k alignment and the exact sizing for asan? I have tried to maintain the size of things as well as possible with the exception of automap since like asan its size was larger than 8tib on its own.

JoshuaWierenga:
Still trying to diagnose old windows crashes again, definite an issue with 149e4969ff946fe367f407cf94efe13ebe82892a but appears related to the other addresses in memtrack.internal.h. They are also problematic but restoring the working values does not fix all applications crashing on startup.

jart:
> memory requirements

It depends on whether or not you're going for ASAN support. If you are, then the way we currently select base addresses for fixed blocks, so that they have optimal behavior and easiness, is the following invariant:

    !(p & 0xffff) && !(((p >> 3) + 0x7fff8000) & 0xffff)

Must be the case. That are very few `p` you can choose where the above expression will evaluate as true. But if you use those, then you can not only use the `_extend()` API but with other stuff it avoids allocating extra pages.

JoshuaWierenga:
I might be missing something but I tried looping through all byte addresses below 8tib and checking which were allowed by that expression. I had to stop it because it found way too many, so either I did it wrong or "very few `p`" doesn't really fit. Regardless, manually checking the values I decided above all pass the test.

JoshuaWierenga:
Old windows regression is because of memtrack.internal.h and stack.h changes in 82efe400f6ea17ad55a02a88e22486b4679c7921, fixing the memory map will fix the issues.

JoshuaWierenga:
That's as far I as I can get without more understanding of how asan is configured and what the kernel memory at the end of the regular user memory map is for.

JoshuaWierenga:
I tried running printarg.com to see what PrintMemoryIntervals gave, shows that I have successfully mapped at least g_fds and stack/winargs though due to not yet patching IsShadowFrame, it reports them as shadow memory:
vmware_qxuzqqYBzs

Compared to windows 10 which of course uses the existing full memory map and can display those blocks correctly:
image

jart:
You've gotten pretty far. Thanks for fixing memtrack's debugging tools too!

I recently wrote another tool that might be helpful for you: https://github.com/jart/cosmopolitan/blob/master/tool/viz/virtualquery.c

JoshuaWierenga:
I am not totally sure how to read this as it conflicts quite badly with both the 8tb and 128tb memory maps. Everything through 0x1fffff should be null but many allocated blocks exist. The first block that actually exists is image at 0x400000 with some related blocks shortly after. Quite a few blocks in arena. I assume that free blocks can be used for things as otherwise there are two allocations in the first free. The 6014gib region is interesting as it covers asan, automap, memtrack, fixedmap, two big free blocks, nsync and zipos as it ends at 0x05dfffff0000 which is just before g_fds, how are the rest handled if not as separate allocations within windows, another arena? or is it meant to be 6014gb which gives 0x057... I think and is before nsync at 0x05c00004000 but by quite a lot. Then g_fds and stack/winargs are allocated as already shown with printargs.com. Not sure beyond that as I did search for info regarding kernel data being mapped in user space as given in the original memory map but didn't find anything obvious.
image

JoshuaWierenga:
Well, that was messy, noticed a mistake in the memory map info file and accidentally amended the wrong commit so that's three force pushes for a one line fix.

JoshuaWierenga:
This might not have anything to do with the changes I made but out of nowhere the setrlimit_test.c is failing, 
image

jart:
Is that a flake? It's a time based test so I could foresee it flaking. But the margin is so large.

JoshuaWierenga:
Ruled out anything I did by compiling the master branch, could be wsl weirdness so I guess don't read anything into it until I can reproduce it on an actual linux machine.

JoshuaWierenga:
Apparently not that far off topic, since it now fails with a memory error. I found a repo with a bunch of winapi polyfills for xp and was able to get quite a few simple programs running. Tried redbean however and it crashes with a segmentation fault in the zip code. The same binary runs fine on vista though, so still might not be worth fixing. 
firefox_irVERL1ba2

JoshuaWierenga:
This has been sitting here for quite a while. Ignoring asan almost everything is mostly done, just need to figure out testing and how some high parts of the memory map are managed.
After this is merged I want to see how many of the changes on the main branch after the vista split can be brought over. I kept track of the overlap in https://github.com/jart/cosmopolitan/pull/635#issuecomment-1264610494. It should be possible to get to at least cosmo 2.1 since I don't think that many memory changes were made in before it.
If testing can be figured out then I would also be interested in continuing work on xp pro x64 edition.

JoshuaWierenga:
Short of full testing being back, would either vista or 8.0 be easier to setup given issues with 7? Otherwise perhaps just pick some complex applications that use lots of cosmo functionality, I was already using redbean for this but others would be nice.

jart:
This is really exciting. Help me understand what you mean by full testing. Let's talk more on Discord. I'll review the code progress on this change more in depth tomorrow.

JoshuaWierenga:
Not sure if it's worth changing but I can move nsync, zipos, g_fds, stack and winargs to match down to the two leading hex digits with their 128tib locations just like I did for automap, _mmi and fixedmap. In this case that just means adding 0x010000000000 to them all.

JoshuaWierenga:
I am a bit busy working on blink now, but I plan to get back to this sometime soon. I just converted the pull back to draft since some extra commits have been pushed to the vista branch that I need to deal with. Regardless, this is mostly done and just needs some testing on one or more older versions of windows < 8.1.

JoshuaWierenga:
I cleaned up the addresses to match their 128tib versions a bit better and rebased this branch on top of the current vista branch. I did test blink on windows 7 and it works fine but running the full test suite would still be nice.
I was however, just checking memplan.c to make sure that winargs would continue to work and noticed that the arena at the beginning of the memory map has the same values on all versions of windows but is setup separately for < and >= windows 10, I assume this isn't intended as only asan is meant to be setup differently(i.e. not at all currently though I did plan for it to work at some point). Also, the image has a different size on windows 8.1 and older, is this intended as well?

G4Vi:
I tested building APPerl with it and running on Windows 7 and it works great.

JoshuaWierenga:
Currently trying to figure out exactly how stack and winargs are positioned. GetStaticStackAddr appears to return 0x700000000000 + APPEND so is my current hardcoded 0x070000040000 for old windows wrong? Should it be 0x070000000000(ape_stack_vaddr) +APPEND?
The newer version of WinMainNew added from master in this pull should then put winargs in 0x070000000000 to 0x070000000000 + sizeof(struct WinArgs)?

JoshuaWierenga:
Just fixed the stack to work the same way on 8tib as it does on 128tib however I am still not confident I know what that means.
Nothing major has changed other than that and so @G4Vi's testing and my own should still apply. 
The only issue I can think of is that I have yet to find where the created winargs struct is used so perhaps that address is hardcoded somewhere and so may not work.

The memory map description files are definitely wrong in some places but now now agree fully outside of the necessary address changes for 8tib vs 128tib. I will fix these fully in a separate commit and provide a tool to automate that process and check the addresses used work with asan given check you gave ages ago but they are close enough for now.

G4Vi (approved):
I'm not able to verify the accuracy of the mapping, but from my testing (with various APPerl builds) it works. Based on that and the reasoning in the commits, documentation, this PR, and the discord, I'll Approve.

PR #649, opened 2022-10-04
[metal] Add basic character output for VGA graphics mode console

tkchia:
The VGA console code can now (https://github.com/tkchia/cosmopolitan/commit/2994634a20d3d2fbc23e22d2051906c81f253c5d) do some basic text output onto a VGA graphics screen, with the help of a public domain bitmap font.  Scrolling also works properly.

To do:
* [ ] implement some sort of "screen of death", to allow easier diagnosing of program crashes when they happen
* [ ] allow switching to a different font &mdash; possibly with a different glyph size &mdash; even after console has already been initialized
* [ ] support TrueType fonts, or at least some sort of graymap fonts


libc/integral/c.inc
```diff
@@ -448,6 +448,16 @@ typedef struct {
 #endif
 #endif
 
+#ifndef unrollloops
+#if !defined(__STRICT_ANSI__) &&                           \
+    ((__GNUC__ + 0) * 100 + (__GNUC_MINOR__ + 0) >= 407 || \
+     __has_attribute(__optimize__))
+#define unrollloops __attribute__((__optimize__("unroll-loops")))
```
jart, line 455: Oh cool I didn't know about this.

libc/vga/tty-graph.c
```diff
@@ -0,0 +1,287 @@
...
+      *plotter++ = __builtin_add_overflow(bits, bits, &bits) ? fg : bg;
+      *plotter++ = __builtin_add_overflow(bits, bits, &bits) ? fg : bg;
+      *plotter++ = __builtin_add_overflow(bits, bits, &bits) ? fg : bg;
+      xleft -= 8;
+    }
+    if (xleft) {
+      uint8_t bits = *bitmap++;
+      switch (xleft) {
+        default:
+          *plotter++ = __builtin_add_overflow(bits, bits, &bits) ? fg : bg;
```
jart, line 189: Cool fallthrough overflow pattern!

libc/vga/vga-font-default.c
```diff
@@ -0,0 +1,1441 @@
...
+    0x00, /* ........ */
+    0x00, /* ........ */
+    0x00, /* ........ */
+    0x00, /* ........ */
+    0x00, /* ........ */
+    0x00, /* ........ */
+  },
+  [1] = { /* 33 */
+    0x00, /* ........ */
+    0x18, /* ...##... */
```
jart, line 33: I guess this has a nice retro aesthetic that'll help us make progress until we can have truetype.

jart (approved):
Thanks for sending this! Approved.

libc/vga/tty-graph.c
tkchia: I suspect that is about as optimized as one can go without bringing in SIMD instructions. :)

ghaerr:
Hello @tkchia, very nice work implementing the graphical VGA console. I'm quite impressed with your optimizations.

I notice that a 24bpp RGB/BGR mode isn't (yet) implemented - is VESA VBE not returning any of these modes?

Thank you!

ghaerr:
Hello @tkchia,

> I recall that my (actual) PC did in fact return some 24bpp modes, including modes 0x118 (1024 × 768 × 64K colours) and 0x112 (640 × 480 × 64K colours) 

Those modes are 16bpp (=64K)... I was referring to 24bpp modes, which are 16M colors, but stored as 24 rather than 32 bits per pixel.

> Do you think it will be useful to support these modes?

I think most modern cards supporting 24bpp would also support 32bpp RGBA/BGRA pixel format, perhaps take a further look to see whether the 24bpp formats have higher resolutions available for display?

>  I imagine that it will be a bit difficult to deal with them in an optimized manner...

Yes, of course the 24bpp format doesn't allow aliasing as short or long integers, but the preparation of such blitters would still be useful for 1) getting ready for having to deal with a separate alpha channel that may require byte moves to/from the framebuffer, and 2) @jart's request for considering an RGB (uint8[3][x]y) API/storage pixel format for the offscreen buffer.

Thank you!

PR #651, opened 2022-10-06
[metal] Fix sys_mmap_metal() issues for non-MAP_FIXED case

tkchia:
- correctly check that virtual region we want to use is unmapped, rather than accidentally clobbering existing pages
- avoid placing `mmap`'d memory at null virtual address


jart (approved):
Thanks!

PR #652, opened 2022-10-06
Update definitions

stellartux:
These changes update the definitions file with the new unix.Memory API and a few other things I'd missed before.


jart (approved):
Thank you!

PR #653, opened 2022-10-07
[metal] Push forward timing of VGA TTY initialization

tkchia:
- arrange to initialize the default VGA TTY data structure, at about the same time as the console stdin, stdout, & stderr descriptors are set up, rather than much later
- ~~tweak the text mode console code to rely less on `unbing()` to convert from Unicode to CP437~~

### DIFF
--- a/libc/vga/vga-init.greg.c
+++ b/libc/vga/vga-init.greg.c
@@ -60,7 +60,7 @@ void _vga_reinit(struct Tty *tty, unsigned short starty, unsigned short startx,
             chr_ht, chr_wid, vid_buf, init_flags);
 }
 
-__attribute__((__constructor__)) static textstartup void _vga_init(void) {
+textstartup void _vga_init(void) {
   if (IsMetal()) {
     struct mman *mm = (struct mman *)(BANE + 0x0500);
     unsigned short starty = mm->pc_video_curs_info.y,
--- a/libc/vga/rlinit-init-vga.S
+++ b/libc/vga/rlinit-init-vga.S
@@ -91,5 +91,12 @@
 vga_console:
 	.endobj	vga_console,globl,hidden
 	.previous
+	.init.start 305,_init_vga
+	push	%rdi
+	push	%rsi
+	call	_vga_init
+	pop	%rsi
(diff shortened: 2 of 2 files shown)


libc/vga/tty.greg.c
```diff
@@ -380,8 +381,19 @@ static void TtyTextDrawChar(struct Tty *tty, size_t y, size_t x, wchar_t wc) {
   uint8_t attr = TtyGetTextAttr(tty);
   struct VgaTextCharCell *ccs = (struct VgaTextCharCell *)tty->canvas;
   size_t i = tty->y * Xn(tty) + tty->x;
-  int c = unbing(wc);
-  if (c == -1) c = 0xFE;
+  int c;
+  /*
+   * This might be an emergency console, so do not rely too much on unbing()
```
jart, line 386: If unbing() is becoming mission critical then its size optimization needs to go away. I've fixed this for you in https://github.com/jart/cosmopolitan/commit/29d695f4c92bfe5c33b26faf7467e3f513f63464

jart (changes requested):

libc/vga/tty.greg.c
tkchia: Thanks @jart.  I have now removed the change to `tty.greg.c`.

libc/vga/vga-init.greg.c
```diff
@@ -60,7 +60,7 @@ void _vga_reinit(struct Tty *tty, unsigned short starty, unsigned short startx,
             chr_ht, chr_wid, vid_buf, init_flags);
 }
 
-__attribute__((__constructor__)) static textstartup void _vga_init(void) {
```
jart, line 63: For the record, it's fine to use `__attribute__((__constructor__))` too :-) I started out with `textstartup` but I've been warming up more to the underscore syntax.

jart (approved):
Thank you!

libc/vga/vga-init.greg.c
tkchia: Thank you @jart for the information!

PR #654, opened 2022-10-10
[metal] Fix bug: do not initialize IDT & TSS if unneeded

tkchia:
`libc/intrin/interrupts.S` should not be linked in unless an IDT (or TSS) is explicitly requested somewhere (i.e. it should probably not be a mandatory module).


tkchia:
It turns out that I misremembered how the `PROVIDE_HIDDEN()` linker script directive really works.

`PROVIDE_HIDDEN(_tss = 0);` means the linker would try very hard to look for a `_tss` symbol within all the code libraries, and only assign a default value if it cannot find it.

What I really wanted to do was to have the linker look for a `_tss` symbol only if it is really needed &mdash; which is better done with a weak symbol.

Thank you!

jart (approved):
Thank you! Rock on that it's possible to be non-mandatory.

PR #656, opened 2022-10-10
Integrate upstream dash changes into unbourne shell

notwa:
this merges the changes from the previous commit (as listed in the source), up to the latest commit to dash, with the changes seen here:
https://git.kernel.org/pub/scm/utils/dash/dash.git/diff/?id=057cd650a4edd5856213d431a974ff35c6594489&id2=fba95e9e4a5d0f1f1ac9f7d86557e47bc0e2656c&context=5&ignorews=0&dt=0
patching was done by hand, but I've done several passes over the past day to make sure I got everything. note that I've omitted the changes from e.g. `stat` to `stat64`.

---

having an Actually Portable shell is potentially useful to me. however, I was running into bugs that had already been fixed upstream — [this bug in particular, involving variable expansions.](https://git.kernel.org/pub/scm/utils/dash/dash.git/commit/?id=a29e9a1738a4e7040211842f3f3d90e172fa58ce) but after patching that, I figured I may as well apply the rest of the changes.

there are still some quirks on Windows, but my PR hasn't affected them at all:

* only `exec program` checks paths for `.com` and `.exe` extensions; simply running `program` will not find `program.exe` in `$PATH`. is this intentional?
* stderr does not play well when run from an MSYS2 terminal, either going silent or getting involuntarily captured(?). I can't even see the whole trace in this scenario because it gets eaten:
```
./unbourne.com --strace -c 'echo hello >&2'
```

### DIFF
--- a/examples/unbourne.c
+++ b/examples/unbourne.c
@@ -59,8 +59,8 @@
   for decades thanks to a spark of brilliance from Ken Thompson.
 
     git://git.kernel.org/pub/scm/utils/dash/dash.git
-    fba95e9e4a5d0f1f1ac9f7d86557e47bc0e2656c
-    Tue Jun 19 11:27:37 2018 -0400
+    057cd650a4edd5856213d431a974ff35c6594489
+    Fri Sep 03 15:00:58 2021 +0800
 
   The UNBOURNE SHELL pays HOMAGE to the Stewards of House California:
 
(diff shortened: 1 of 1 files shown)


jart (approved):
Wow a lot of work went into this. Thank you!  This is one of the first things I put in the codebase and the way everything is smooshed together sort of reflects that, since the build tools weren't as good back then.

If you're running it on Windows, then the biggest issue I've encountered so far is the newline issue. Two of us have tried to fix the `\n` \r\n` thing so it can work better on Windows and failed because the almquist code is nontrivial to change. 

I'd love to see that fixed somehow. In the meantime I've been using `o//examples/cosh.com` as my shell on Windows.

PR #657, opened 2022-10-11
Fix ParseHttpMessage failing to store >2 repeatable headers

malcolmseyd:
ParseHttpMessage had a bug where `HttpMessage.xheaders` would not exceed a capacity of `1` inside of `ParseHttpMessage`. This resulted in only 2 repeatable headers being stored, the first in `HttpMessage.headers`, and the second in the 1-capacity `HttpMessage.xheaders`.

This pull request implements a unit test to detect this bug and a fix that passes the unit test. Now, when `xheaders` has 0 capacity, it's set to 1, and when it's full, the capacity gets doubled.

Also, `clang-format` rearranged the headers in the test file.

### DIFF
--- a/net/http/parsehttpmessage.c
+++ b/net/http/parsehttpmessage.c
@@ -236,8 +236,12 @@ int ParseHttpMessage(struct HttpMessage *r, const char *p, size_t n) {
                 unsigned c2;
                 struct HttpHeader *p1, *p2;
                 p1 = r->xheaders.p;
-                c2 = r->xheaders.c + 2;
-                c2 = c2 >> 1;
+                c2 = r->xheaders.c;
+                if (c2 == 0) {
+                  c2 = 1;
+                } else {
+                  c2 = c2 * 2;
+                }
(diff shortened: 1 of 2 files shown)


jart (approved):
Thanks for spotting this. Looks like I forgot the `+=` in `c2 = c2 >> 1` . Very well executed by the way in the two commit style, creating a failing test case, and then the fix.

Welcome as a new contributor. We need to go through one quick hurdle before we can proceed. Could you please email Justine Tunney <jtunney@gmail.com> and say that you intend to assign her the copyright to the changes you contribute to Cosmopolitan? Please send the email under your legal name rather than anonymously. Please use a real email account associated with your identity. See [CONTRIBUTING.md](CONTRIBUTING.md) for further details. We make the ask because it helps us ship the tiniest most legally permissive prim and proper binaries. You also only need to do it one time, so any future changes you send us can be quickly merged.

Thanks!

malcolmseyd:
Email sent. I'm glad that I can contribute to this great project!

PR #660, opened 2022-10-11
vista: backport execve escaping and using cocmd as shell for system, etc.

G4Vi:


jart (approved):
Thank you! I've got more great changes coming for you in a change I'll push shortly after this gets merged.

PR #663, opened 2022-10-12
[metal] Allow more fine-grained control over page permissions

tkchia:
- use `PAGE_RSRV` bit (originally only for blinkenlights), rather than `PAGE_V` bit, to indicate that a virtual address page has been reserved — this should allow a program to create & reserve inaccessible "guard pages"
- mark page table entries for non-code pages with `PAGE_XD` bit, which should be supported on (circa) post-2004 x86-64 CPUs


jart (approved):
Nice!

PR #667, opened 2022-10-17
[metal] Try to get zipos and open(argv[0], ) to work on metal

tkchia:


tkchia:
`hellolua.c` now works on bare metal when built in tiny mode.
```
make -O -j4 MODE=tiny o/tiny/examples/hellolua.com
qemu-system-x86_64 -hda o/tiny/examples/hellolua.com -serial stdio
```

tkchia:
It does not work yet to run `hellolua.c` on bare metal if it is built in "non-tiny" mode, i.e.
```
make -O -j4 o//examples/hellolua.com
```
This is (partly?) because the Cosmopolitan bootloader currently only loads the first `0x70000` - `0x2000` = `0x6e000` bytes = 440 KiB of the program image into real-mode-addressable memory, and non-tiny `hellolua.com` exceeds that (496 KiB).

I see that `ape/ape.lds` defines `v_ape_highsectors`, which will be the number of sectors be loaded and transferred to higher (extended) memory, though this is not quite implemented yet.  This can probably be considered a separate issue.

Thank you!

ghaerr:
Hello @tkchia,

Nice work, I can see you're neck deep in the guts of bare metal virtual memory management and doing well! :)

> ape/ape.lds defines v_ape_highsectors, which will be the number of sectors to be loaded and transferred to higher (extended) memory

The current metal boot loader only loads ~440k of the .com file into lower main memory; can this be expanded to ~620k (or the amount of reportedly available main memory below 1M), or are you thinking it better to instead allow for execution of arbitrary-sized .com files and use real mode BIOS I/O combined with BIOS INT 15h XMS copy to get sectors into extended memory at load time? Most BIOS calls aren't operational in long mode, so the alternative of performing disk I/O in long mode is quite complicated, do you think?

With regards to `InitializeMetalFile`, I'm not quite following the "Do this only if needed" - your implementation will always prepare and create a full copy of the loaded .com file in additional memory from IMAGE_BASE_REAL, correct? Is using that memory always required? I'm asking mostly out of curiosity, since the amount of extra memory used frankly would seem to always be quite small, unless the executable were huge. I suppose the question may be what kinds of things the executable might change in its own data segment and overall correctness... just wondering if you could expand your thoughts on this. With more complexity, a "copy on write" approach could be used where no extra real memory is used, only virtual page table entries?

What is the purpose of the static yoink of "APE_COM_URI", what does that actually bring in?

Thank you!

tkchia:
Hello @ghaerr,

> With regards to `InitializeMetalFile`, I'm not quite following the "Do this only if needed"

Good point &mdash; I have pushed another commit to clarify this.  `STATIC_YOINK(APE_COM_URI)` is the "directive" that will ultimately rope in `InitializeMetalFile` and data structures for supporting `"/ape.com"`.

> just wondering if you could expand your thoughts on this. With more complexity, a "copy on write" approach could be used where no extra real memory is used, only virtual page table entries?

Well, yes, "copy on write" will be cool. :slightly_smiling_face: 

> The current metal boot loader only loads ~440k of the .com file into lower main memory; can this be expanded to ~620k (or the amount of reportedly available main memory below 1M), or are you thinking it better to instead allow for execution of arbitrary-sized .com files

Good question.  For now there seem to be pros and cons to either approach.  I am leaning more towards allowing arbitrary-sized programs, but the code to support this might be a bit more complicated.  In particular the page allocation logic
```
static noasan textreal void __normalize_e820(struct mman *mm) {
  ...
  mm->pdp = MAX(0x80000, mm->e820[0].addr);
  mm->pdpi = 0;
  mm->e820n = n;
}
```
will need a rework...

> and use real mode BIOS I/O combined with BIOS INT 15h XMS copy to get sectors into extended memory at load time?

I am thinking it might be easier for the bootloader to directly switch to unreal mode itself. :slightly_smiling_face:  The `int 0x15` interface is not exactly very developer-friendly (in my opinion).

Thank you!

tkchia:
Hello @ghaerr,

> With more complexity, a "copy on write" approach could be used where no extra real memory is used, only virtual page table entries?

I guess another option might be pull some shenanigans to sequester just the pages that contain the embedded `.zip` file, then leave the rest of `"/ape.com"` as effectively inaccessible.  This will _probably_ work if the main use of opening `"/ape.com"` or `"/proc/self/exe"` is only to access the embedded `.zip` anyway.

Thank you!

tkchia:
Hello @ghaerr,

I _think_ the intent of `zipos` 's design is to allow a user to modify the program's embedded `.zip`, even after build time.  This is probably why the `zipos` support code insists on reading the original executable, rather than relying on whatever data has been mapped into the ELF address space.  (The ELF mappings may become out of sync with the `.zip` contents if the latter is modified after build time.)

Unfortunately this arrangement starts to hit problems, when the Cosmopolitan program is not a normal filesystem file, but is instead "burnt onto" (say) a thumb drive:
  * The key metadata structure [for a `.zip` file](https://support.pkware.com/home/pkzip/developer-tools/appnote) is the "end of central directory record".  This is expected to be near the _end_ of the zip file.
  * When a program is treated as a disk image and written to a disk, the information about the program file's size is lost.
  * And there is no easy way to update the program binary &mdash; with third-party tools, anyway &mdash; to reflect any changes in an embedded `.zip` 's size, in a way that the program can use.

I am not quite sure how to address this.  For the moment I am mainly trying to keep the existing `zipos` support logic while getting it all to work.  But yes, probably a bit more needs to be done.

Thank you!

ghaerr:
Hello @tkchia,

Thank you for your insightful comments - I hadn't thought about the problem of the ELF sections not being updated properly along with any embedded .zip being updated; more to consider than I realized.

> When a program is treated as a disk image and written to a disk, the information about the program file's size is lost

I see - so the ELF sections (or linker symbols) can't tell us what the correct size is, and there is no filesystem tracking the overall image size - how is it then, that the bare metal startup code knows how many sectors to read, then, especially when the internal .zip file has been enlarged?

> For the moment I am mainly trying to keep the existing zipos support logic while getting it all to work. But yes, probably a bit more needs to be done.

Ok, agreed. My original concern was over having to make an internal copy though - doesn't this also mean that the internally-specified size of the image itself for that copy could be incorrect?

Thank you!

tkchia:
Hello @ghaerr,

> how is it then, that the bare metal startup code knows how many sectors to read, then, especially when the internal .zip file has been enlarged?

Actually the bare metal code only knows about the `.zip` file's size at build time.  Getting it to handle changes to the `.zip` contents  is still not quite a solved problem.

Thank you!

paulwratt:
TLDR; Is the real issue that the executable _does not know_ where the _end of itself is_.

(without having looked through the `zipos` code) I'm wondering why there is _any_ data _in_ the ELF headers regarding the ZIP part of the file. if a valid zip is _appended_ to a(ny) executable, the OS (ELF) headers store the correct offset to the ZIP without _knowing_ there is any ZIP (or any other format). The only issue I ever had (doing something simialr to `zipos`) was how does the executable know where the beginning of the (int his case) ZIP file is _without_ doing file access, and that was to _double build_ where the correct value is only available _after_ being build once (being that iv'e probably changed something since last build, hence the need to compile again). - specifically I used this technique to dump out "+appened sources", that can be _piped_ to a file (ZIP or RAW).

what ever load the executable _knows_ that there is something up, but only if it cross reference the SIZE in the header, with the SIZE on disk. If the executable itself knows _explicitly_ where the end of _itself_ is then _that +1_ is the beginning of the (in this case) ZIP file. NOTE that this too overcomes "what if zip changes", as ZIP is of size "arbitary data length"

tkchia:
Hello @paulwratt,

Actually, no, no, not quite.  We absolutely _do_ want to know where the _end_ of the `.zip` content is.  Specifically, we want to know the location of the "end of central directory record".  And this is at the end of the `.zip` content.

Thank you!

pkulchenko:
@tkchia, strictly speaking, the central directory record doesn't need to be at the end: https://taras.glek.net/post/optimized-zip-format/; this is just a convenience to do quick append operations. In fact, I don't see anything that would forbid having two central directory records (one before and one after the content), which should simplify the search and may work regardless of where the search is started. The flip size is that you may have to build the archive in a special way for that.

jart:
> hellolua.c now works on bare metal when built in tiny mode.

Wow! Great news. Thanks for the patience everyone. I'm reviewing this change now.

jart:
> @ghaerr As I read a little closer, it seems that the location of any embedded .zip can't be easily known at open("/ape.com") time, as it seems to be done through brute force search in GetZipCdir. However, given that the completely loaded .com executable will always be fully mapped to virtual addresses BANE+IMAGE_BASE_REAL, why not just use that address to start search, and/or if required another sys_mmap of the same with no copy?

I think we used to do that but it was removed.

> @tkchia  I think the intent of zipos 's design is to allow a user to modify the program's embedded .zip, even after build time.

That's the goal!

> When a program is treated as a disk image and written to a disk, the information about the program file's size is lost.

If we make zipos read-write with a disk image, what I'd do is upon the first mutation, I'd relocate the ZIP EOCD to the tail end of the disk image. So what we'd do at startup is search backwards 64kb from the end of the disk partition (or image) for the EOCD, and if it isn't there, then we can assume it's where it was originally linked.  In other words, the disk itself is the file. We immediately grow to consume all available space. That would let us support @ghaerr's proposal while allowing room for growth.

> @pkulchenko strictly speaking, the central directory record doesn't need to be at the end: https://taras.glek.net/post/optimized-zip-format/

The CD doesn't need to be at the end, but the EOCD *must* be in the last 64kb of the file.

>  I don't see anything that would forbid having two central directory records (one before and one after the content),

The CD entries need to be contiguous. You can have multiple CDs, for instance if you're writing a concurrent algorithm that transactionally rolls forward in an append-only fashion, but only one CD officially exists at a time.

libc/calls/metalfile.internal.h
```diff
@@ -9,6 +9,14 @@ struct MetalFile {
...
 };
 
+extern void *__ape_com_base;
+extern size_t __ape_com_size;
+
 COSMOPOLITAN_C_END_
 #endif /* !(__ASSEMBLER__ + __LINKER__ + 0) */
+
+#define APE_COM_NAME     "/ape.com"
+#define APE_COM_ALT_NAME "/proc/self/exe"
```
jart, line 19: I'd say make the alt name the only name.

libc/calls/metalfile.internal.h
```diff
@@ -9,6 +9,14 @@ struct MetalFile {
...
+#define APE_COM_URI      "file:/proc/self/exe"
```
jart, line 20: Why do we need a URI?

libc/calls/getfiledescriptorsize.c
```diff
@@ -60,7 +61,14 @@ ssize_t getfiledescriptorsize(int fd) {
       rc = -1;
     }
   } else if (IsMetal()) {
-    rc = -1;
+    if (fd < 0 || fd >= g_fds.n)
```
jart, line 64: Nit: Curlies when doing multiple lines.

libc/calls/getprogramexecutablename.greg.c
```diff
@@ -79,6 +79,7 @@ static inline void GetProgramExecutableNameImpl(char *p, char *e) {
   // if argv[0] exists then turn it into an absolute path. we also try
   // adding a .com suffix since the ape auto-appends it when resolving
   if (__argc && (((q = __argv[0]) && !sys_faccessat(AT_FDCWD, q, F_OK, 0)) ||
+                 IsMetal() ||
```
jart, line 82: Wouldn't it be simpler to just add a statement higher up that says:

```c
if (IsMetal()) {
  strcpy(p, "/proc/self/exe");
  return;
}
```

jart (changes requested):

libc/calls/metalfile.internal.h
tkchia: > Why do we need a URI?

@jart : I was following the `tool/build/lib/elfwriter_zip.c` behavior of using a URI as an ELF symbol which can be exported (or imported) as part of an ABI.  But other than that, there is probably no real need for a URI here.  Thank you!

libc/calls/getfiledescriptorsize.c
tkchia: @jart : I have fixed this.  Thank you!

libc/calls/metalfile.internal.h
tkchia: @jart : Fixed.  Thanks!

libc/calls/getprogramexecutablename.greg.c
tkchia: @jart : I have made the requested addition.  (Except, I am using `memccpy` rather than `strcpy`, in order to take the output buffer size into account.)  Thank you.

jart (approved):
Cool let me know when conflicts are resolved and it's ready to merge.

tkchia:
Hello @jart,

> Cool let me know when conflicts are resolved and it's ready to merge.

Thank you!  I have rebased this PR on top of the updated main branch.  It should now be ready to merge once the tests pass.

tkchia:
Hello @jart, hello @ghaerr,

> If we make zipos read-write with a disk image, what I'd do is upon the first mutation, I'd relocate the ZIP EOCD to the tail end of the disk image. So what we'd do at startup is search backwards 64kb from the end of the disk partition (or image) for the EOCD, and if it isn't there, then we can assume it's where it was originally linked. In other words, the disk itself is the file. We immediately grow to consume all available space. That would let us support @ghaerr's proposal while allowing room for growth.

This is possible, I suppose &mdash; though I am not yet sure how exactly it will play out in practice.  Currently what I normally do, when I wish test a program on bare metal, is to write it onto the first few sectors of a whole (USB) disk, like so:
```
$ ls -l o/tiny/examples/hellolua.com
-rwx------ 1 user user 335872 Nov  2 16:04 o/tiny/examples/hellolua.com
$ sudo dd if=o/tiny/examples/hellolua.com of=/dev/sdX conv=notrunc
```
(It i possible to "stretch" a program file such as `o/tiny/examples/hellolua.com` &mdash; including the embedded `.zip` &mdash; into a sparse file with a large size, say, 8 GiB.  But once Info-Zip modifies the file, it shrinks back again (to e.g. 382 KiB).  And, I could not get Info-Zip to correctly find the "end of central directory" within a `/dev/sdX` device special file.)

---

Hello @pkulchenko,

The `mozpack` tool which Taras Glek refers to, seems interesting indeed.  But it seems to be distributed only as part of the Mozilla build tree, rather than a separate tool in its own right?  This might pose a bit of a problem.

Thank you!

ghaerr:
Hello @tkchia,

> I could not get Info-Zip to correctly find the "end of central directory" within a /dev/sdX device special file.)

I assume you mean here on a Linux box, not on bare metal, right?

> But once Info-Zip modifies the file, it shrinks back again (to e.g. 382 KiB). 

I would think it likely best to stay in conformity with standard tools like you're using, rather than possibly add special requirements for tools used to add .zip entries.

That said, it seems to me perhaps considering splitting the presently discussed .zip issues into two, with separate solutions. Given that it seems that there are still possibly a number of things that need to be done to get bare metal supporting a graphics console with a self-contained bitmap or truetype font, it might be ok to initially only handle the simpler case of a makefile-built .com file with included font(s), then add support for zip-tool modified .com file contents later.

> That would let us support @ghaerr's proposal ...

Of the many items I have commented on (and thank you @tkchia for all your work!), my main concern/proposal remains only that of not requiring a `memcpy`-based copy of the already-loaded .com file, just in order to read .zip file contents. I had initially proposed copy-on-write, but then backed off on that thought when I couldn't understand the requirement for any copy at all - that is, why not implement a method where the already-loaded .com file can be searched, possibly with zipos modifications, and the EOCD found without any memory copy (i.e., just search `[BANE+IMAGE_BASE_REAL .. BANE+IMAGE_BASE_REAL+loaded_size]`. IMO, there are lots of good reasons for not copying. The enhancements within this PR that allow for a bare-metal memory copy with automatic page allocation for other reasons are super, and surely will find usefulness in the future for other reasons. If for some reason I am missing something major here, then it would still seem that a page-table-only copy would be better than a `memcpy`, which doubles the physical memory requirement for all bare metal graphics console programs. 

Thank you!

tkchia:
Hello @ghaerr,

> I assume you mean here on a Linux box, not on bare metal, right?

Yes &mdash; we cannot really run a separate Info-Zip process on bare metal (yet).

> Of the many items I have commented on (and thank you @tkchia for all your work!), my main concern/proposal remains only that of not requiring a `memcpy`-based copy of the already-loaded .com file, just in order to read .zip file contents.

Yes, I kind of agree.  If you ask me, a good solution would be to embed the `.zip` file in such a way, that it is not considered part of the "official" ELF executable and its virtual memory mappings, even though it is still read into physical memory.  Currently:
  * the `.zip` content is considered to be part of the second ELF program segment (`PF_R | PF_W` under `ape/ape.S`), which is read/write &mdash; it is included into `ape_ram_filesz`;
  * and the `.zip` content is followed by the `.tdata` section, which contains the initialization image for thread-local storage.

These should probably change, in my view.  Namely: leave the `.zip` content our of the ELF segments, and move `.tdata` somewhere before the `.zip`.

> (i.e., just search `[BANE+IMAGE_BASE_REAL .. BANE+IMAGE_BASE_REAL+loaded_size]`. 

(Regarding `IMAGE_BASE_REAL`, it seems to me &mdash; from earlier versions of the Cosmopolitan code &mdash; that the original plan was to get the program image positioned at `IMAGE_BASE_PHYSICAL` = `0x100000`, after loading (part of) it at `IMAGE_BASE_REAL`.  Perhaps we should get Cosmopolitan to do that again.  This will also offer a natural way to handle programs that are larger than 440 KiB.)

Thank you!

tkchia:
Hello @ghaerr,

> I would think it likely best to stay in conformity with standard tools like you're using, rather than possibly add special requirements for tools used to add .zip entries.

There is a trade-off here, from what I see.  Fewer requirements at `.zip` creation time, will mean more work for the bare metal loader at run time.  To _really_ be able to handle arbitrary `.zip` files of the kind that Info-Zip spews out, we will probably need to parse a `.zip` file's structures in the forward direction.  In particular, if memory serves me well, in some cases &mdash; depending on how the `.zip` was created &mdash; we might actually have to parse the Huffman codes of a compressed stream, just to know how long the stream is (!).

> Bit 3: If this [general purpose] bit is set, the fields crc-32, compressed size and uncompressed size are set to zero in the local header.  The correct values are put in the data descriptor immediately following the compressed data.  (Note: PKZIP version 2.04g for DOS only recognizes this bit for method 8 compression, newer versions of PKZIP recognize this bit for any compression method.)

Thank you!

ghaerr:
Hello @tkchia,

> (Regarding IMAGE_BASE_REAL, it seems to me — from earlier versions of the Cosmopolitan code — that the original plan was to get the program image positioned at IMAGE_BASE_PHYSICAL = 0x100000, after loading (part of) it at IMAGE_BASE_REAL. Perhaps we should get Cosmopolitan to do that again. This will also offer a natural way to handle programs that are larger than 440 KiB.)

Are you suggesting that perhaps the real-mode loader (i.e. the PC-compatible bootstrap sector) would load the first 440 KiB to IMAGE_BASE_REAL is it does now, then continue loading any additional sectors to IMAGE_BASE_PHYSICAL+0x6e000 (=0x16e000), then copy the lower 440 KiB to IMAGE_BASE_PHYSICAL (=0x100000) after the page tables are setup? That sounds like a neat idea.

What would the lower (non-extended main memory below 1M) be used for, would it be released from the active process memory and be later available for `mmap` or `malloc`, virtually mapped as normal but marked as free and not reserved by the VM management code?

Thank you!

ghaerr:
Hello @tkchia,

> a good solution would be to embed the .zip file in such a way, that it is not considered part of the "official" ELF executable and its virtual memory mappings, even though it is still read into physical memory. 

Yourself and @jart are way ahead of me on the internal details of .zip files and their associated central directory, but it sounds like there are problems having any process-required ELF data (particularly .tdata currently) after any .zip data within a .com file. I'm guessing this is because of the possible need for expansion, as well as your previous comments.

Thus, as you have indicated, moving the .zip contents to either always the end of a .com file (or perhaps a separate specially-identified ELF section?), and modifying the linker script to place .tdata and all other non-zip data prior, would work, right? The APE loader (for bare metal, and both internal and external APE loaders for and Linux/BSD etc) would be modified to explicitly allocate and load any ZIPELF section after the existing normal loading mechanism, and the memory address for that load could be anywhere, not necessarily contiguous. Or is there a problem with when using the host OS ELF loader to load binaries and requiring a "data" ELF section to be therefore used for the .zip data?

> we will probably need to parse a .zip file's structures in the forward direction. That is, go through the local file headers, file content, and central directory entries, and finally figure out where the "end of central directory" entry is.

If the .zip data/CD etc are always located at the end of the physical file, this would eliminate the need for this special processing, except for the modified loader(s), and standard tools could be used, and use the existing working zipos identification code, provided the ZIPELF/ZIPDATA was previously loaded somewhere, right?

> To really be able to handle arbitrary .zip files of the kind that Info-Zip (etc.) spews out

Perhaps the goal should be to allow a limited capability of updating .zip data in bare metal .com files, rather than leaving it completely arbitrary. It seems that the currently-allowed mechanism works well for redbean .com files to be easily updated, and we could accept the same limitations for bare metal .com files as well. Given that the bare-metal case(s) will likely be used far less frequently, it seems possibly a decent tradeoff, unless you think that users may want to change fonts based on the VESA implementation. This wouldn't be an issue with truetype, though.

Sorry for the many rambling questions, all things considered, it seems a good idea to have all Cosmo-required sections, other than .zip, to be contiguous in the makefile-built as well as zipinfo-built .com files, with the zip data always appended, if that can be depended on.

Thank you!

tkchia:
Hello @ghaerr,

> What would the lower (non-extended main memory below 1M) be used for, would it be released from the active process memory and be later available for `mmap` or `malloc`, virtually mapped as normal but marked as free and not reserved by the VM management code?

Probably &mdash; I hope to implement this.  It seems that the memory management front can use some further work &mdash; in particular the reclaiming of freed, `munmap` 'd pages is not yet implemented.

> Given that the bare-metal case(s) will likely be used far less frequently, it seems possibly a decent tradeoff, unless you think that users may want to change fonts based on the VESA implementation. This wouldn't be an issue with truetype, though.

Well, at the moment the more pressing (?) issue to me, would be the 440 KiB limit on the total program size (including the `.zip`) in bare metal mode.  This is a problem, considering that e.g. the Noto Sans TrueType font file easily takes up 150 KiB of space, even with compression.  I hope to work on removing the program size limit, once this PR is merged.

Thank you!

jart:
I think we're good to go now. Thanks again for sending this!

PR #668, opened 2022-10-17
minor nt fcntl and lseek improvements

G4Vi:
I found these trying to get my personal media server up and running on Windows with cosmopolitan and comparing the `--strace` output with it running on Linux.

`O_NONBLOCK` support would be really nice, for sockets a thread might not be needed with `ioctlsocket`.


jart (approved):
Great finds! Thank you!

PR #671, opened 2022-10-21
Redbean gzip more

pkulchenko:
@jart, redbean compression update per our discussion and a couple of documentation fixes.


jart (approved):
Thank you!

PR #672, opened 2022-10-25
check existence of IFNAMSIZ

ahgamut:
IFNAMSIZ is defined in:

- libc/sock/if.h
- libc/sock/struct/ifreq.h

we add a check in the latter to avoid a complaint when using the amalgamated header.

https://github.com/ahgamut/debug-cosmo-example/issues/1


jart (changes requested):
Thanks for sending this. Could you delete `IFNAMSIZ` from your choice of header instead? So we only have a single definition in the codebase?

ahgamut:
I removed IFNAMSIZ from `libc/sock/if.h`, which made it an empty header that's `#include`d in exactly one file, so I removed the `libc/sock/if.h` itself.

jart (approved):

vinodkd:
FYI, the zip downloadable from https://justine.lol/cosmopolitan/download.html (actual url: https://justine.lol/cosmopolitan/cosmopolitan.zip) still has this issue:
```
In file included from <command-line>:31:
./cosmopolitan.h:28115: warning: "IFNAMSIZ" redefined
28115 | #define IFNAMSIZ    IF_NAMESIZE
      | 
./cosmopolitan.h:28012: note: this is the location of the previous definition
28012 | #define IFNAMSIZ 16
```
 The fix is easy enough locally, just delete line  28115. However, for the first time attempt at using cosmo, it'd be great if the header file in the downloaded zip had this fixed already.

PR #678, opened 2022-11-03
bestline keybindings and redbean definitions

stellartux:
This adds keybindings for Ctrl-Left/Right for back/forward word, and a couple more redbean definitions I'd missed.


jart (approved):
Thank you!

PR #680, opened 2022-11-06
Add redbean sqlite session support

pkulchenko:
@jart, this implements SQLite session support as well as wal_hook and wal_checkpoint function. It can be disabled if needed by removing `-DSQLITE_ENABLE_SESSION`.


jart (approved):
Thanks!

PR #685, opened 2022-11-07
[metal] Allow programs larger than 440 KiB to run in bare metal mode

tkchia:
This PR tries to address the problem outlined at https://github.com/jart/cosmopolitan/pull/667#issuecomment-1283748049 : the bare metal bootloader only loads the first 440 KiB of the program image into memory, so programs larger than 440 KiB crash.

To do:
* [x] Get `o//examples/hellolua.com` to run correctly under QEMU in legacy BIOS mode
  * [x] Copy first 440 KiB of program pages to extended memory
  * [x] Load program pages beyond first 440 KiB into extended memory as well
  * [x] Reclaim memory pages used at boot time, and make them available for later application use
* [x] Ensure `o//examples/vga2.com` ~~(and/or `o//examples/hellolua.com`)~~ also run correctly when started in UEFI mode (&check; https://github.com/tkchia/cosmopolitan/commit/36b06ab73e107796d93f27e63e17e5a0655170db, https://github.com/tkchia/cosmopolitan/commit/9fe2f551139d08c23d17bc94e1cde45076de4794, https://github.com/tkchia/cosmopolitan/commit/8f698a13f570a2c41e46ea357e3db8b12af02855)
* [x] Allow `munmap` to reclaim dynamically allocated pages (&check; https://github.com/tkchia/cosmopolitan/commit/de295966b1f24dc5c2c70d3d36a60af9aab47313)
  * [x] Implement reference counting of physical pages, to decide when to reclaim them 
* [ ] (?) Maybe add integrity checks to guard against corruption in free list of reclaimed pages

### DIFF
--- a/ape/ape.S
+++ b/ape/ape.S
@@ -1091,6 +1091,42 @@ ape_idata_idt:
 ape_idata_idtend:
 	.previous
 
+	.section .piro.data.sort.metal_gdt,"aw",@progbits
+	.align	8
+_gdt:
+//	          ┌G:granularity (1 → limit *= 0x1000)
+//	          │┌D/B:default operation size (0 = 16|64bit, 1 = 32-bit)
+//	          ││┌L:long mode
+//	          │││┌AVL:this bit is thine (1<<52)
+//	          ││││    ┌P:present
+//	          ││││    │┌DPL:privilege
+//	          ││││    ││ ┌─────────data/code(1)
+//	          ││││    ││ │┌────data(0)──────code(1)
+//	          ││││    ││ ││┌───conforming───expand-down
+//	          ││││    ││ │││┌──writeable────readable
+//	          ││││    ││ ││││┌─accessed─────accessed
+//	          ││││    ││ │││││
+//	          ││││ ┌──││─│││││───────────────────────────────┐
+//	      ┌───││││─│──││─│││││───────────┐                   │
+//	  ┌───┴──┐││││┌┴─┐│├┐│││││┌──────────┴───────────┐┌──────┴───────┐
+//	  │      ││││││  ││││││││││          base address││ segment limit│
+//	  │      ││││││  ││││││││││               32 bits││       20 bits│
+//	  │      ││││││  ││││││││││                      ││              │
+//	  6666555555555544444444443333333333222222222211111111110000000000
+//	  3210987654321098765432109876543210987654321098765432109876543210
+//	  │      ││││││  ││││││││││                      ││              │
+.quad	0b0000000000000000000000000000000000000000000000000000000000000000 # 0
+.quad	0b0000000000001111100110100000000000000000000000001111111111111111 # 8
+.quad	0b0000000000001111100100100000000000000000000000001111111111111111 #16
+.quad	0b0000000011001111100110100000000000000000000000001111111111111111 #24
+.quad	0b0000000011001111100100100000000000000000000000001111111111111111 #32
+.quad	0b0000000010101111100110110000000000000000000000001111111111111111 #40
+.quad	0b0000000010101111100100110000000000000000000000001111111111111111 #48
+.tssdescstub _tss							#56,64
+_gdt_end:
+	.endobj	_gdt,global,hidden
+	.previous
+
 	.section .piro.data.sort.iat.1,"aw",@progbits
 	.type	ape_idata_iatend,@object
 	.type	ape_idata_iat,@object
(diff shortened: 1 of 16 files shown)


ghaerr:
Hello @tkchia,

Looks nice, great enhancements :)

Perhaps it's time to put the 0x79000 (and possibly 0x7f000) constants into defines and a common header somewhere?

Is IMAGE_BASE_REAL useful to an application anymore, or after the copying of low and high parts of the .com file, the system will only use IMAGE_BASE_PHYSICAL?

Thank you!

tkchia:
Hello @ghaerr,

> Perhaps it's time to put the 0x79000 (and possibly 0x7f000) constants into defines and a common header somewhere?

Probably. :slightly_smiling_face: 

> Is IMAGE_BASE_REAL useful to an application anymore, or after the copying of low and high parts of the .com file, the system will only use IMAGE_BASE_PHYSICAL?

Yes, after startup the system will only use `IMAGE_BASE_PHYSICAL`.

Thank you!

ghaerr:
Hello @tkchia,

Are you ready to have this committed, or is there more to do for this PR? It will be nice to (continue) to work on the bare metal enhancements using your enhanced loader and mem mgmt routines once this is committed.

It seems that almost everything in this PR affects only bare metal operation and/or real mode .com file startup, right?

Perhaps after commit more discussion can be had about copying versus page table allocation for bare metal applications (for embedded zip files). On a similar subject I was reading that CPUs may need a TLB cache invalidate after changing certain page table entries... is that being done now, or does that only need to happen when existing already allocated pages in table entries are modified or for multi-threaded systems?

Another quick question: is the reason a linked list has to be used for the PTE reclamation, rather than the Present bit, because we're having to use the BIOS E820 map that may have holes in it? It would otherwise seem that merely turning off the Present bit would be lots simpler (unless the underlying memory doesn't exist). Either that or a free memory bitmap could be used.

Super work on getting the UEFI boot and graphics init working :) Is UEFI available on most newer PCs, I'm mostly familiar with much older ELKS-compatible hardware.

Thank you!

tkchia:
Hello @ghaerr,

> Are you ready to have this committed, or is there more to do for this PR?

It is ready for merging.  The "integrity checks" thing in my to-do list is good to have, but not exactly essential.

> Another quick question: is the reason a linked list has to be used for the PTE reclamation, rather than the Present bit, because we're having to use the BIOS E820 map that may have holes in it?

Well, a linked list can quickly (_O_(1)) answer the question "where can I find a physical page that has been used and later freed up?"  Unfortunately a field of bits in a page table entry simply cannot do that.

> Is UEFI available on most newer PCs, I'm mostly familiar with much older ELKS-compatible hardware.

I am not sure, but I have come across at least two x86-64 PCs which are capable of both UEFI booting and legacy BIOS operation.  And, I strongly suspect that there are some newer machines which only allow UEFI booting, but I am not quite sure yet.

Thank you!

jart:
Apropos linked lists, I built a really nice double linked list abstraction for Blink (the project that's going to let cosmopolitan binaries run on different architectures) and I intend to bring it into the Cosmo codebase soon. 

- https://github.com/jart/blink/blob/master/blink/dll.c
- https://github.com/jart/blink/blob/master/blink/dll.h

It uses the same trick as the Linux kernel where `DLL_CONTAINER()` allows us to avoid having to store an additional pointer in the `struct Dll` data structure.

ape/ape.S
```diff
@@ -1405,10 +1437,88 @@ a20:	cli
...
+	pop	%es
+	ret
+	.endfn	cpyhi
+
+//	Disables interrupts and switches to "unreal mode".
+//
+//	@return	ds, es have same base addresses as before but can access
+//		4 GiB of memory
+//	@clob	eax
+unreal:	push	%ds
```
jart, line 1461: Unreal!!

ape/ape.lds
```diff
@@ -349,9 +349,30 @@ SECTIONS {
     HIDDEN(_etext = .);
     PROVIDE_HIDDEN(etext = .);
 /*END: Read Only Data (only needed for initialization) */
-/*END: Read Only Data */
   } :Rom
 
+/* initialization image for thread-local storage, this is copied */
```
jart, line 354: This is the hardest part of the linker script. Congratulations on successfully editing it. I've spent many hours scratching my head about how to make TLS work with the linker.

libc/nt/efi.h
```diff
@@ -97,6 +97,12 @@
       0x8E, 0x3F, 0x00, 0xA0, 0xC9, 0x69, 0x72, 0x3B \
     }                                                \
   }
+#define GRAPHICS_OUTPUT_PROTOCOL                     \
```
jart, line 100: You got EFI working too?! I love you.

jart (approved):
I'm breathless. This change is extraordinary. You're making my dreams come true. Thank you.

PR #699, opened 2022-11-24
Upgrade SQLite to 3.40

pkulchenko:
@jart, I upgraded SQLite to 3.40 and merged all the changes you applied on top of the previous version. I marked each version with a reference to its "original" version. There were a couple of changes that I didn't merge, primarily because I wanted to reduce the number of changes to file name to simplify the next upgrade. All the functional changes have been applied.

I've only renamed `geopoly.c` to `geopoly.inc`, but we can also filter it out in the makefile.

I've tested all the changes with my scripts in redbean and sqlite shell and everything is working so far (including some new functions, like memdb vfs, that were not present in 3.35).

[update] I relented and renamed several other files, as they had been causing errors for this build, even though sqlite itself compiles without issues. This is mostly related to the fact that those are not standalone .h files and depend on them being included in a certain manner, such that their dependencies are all satisfied. I ended up renaming them to .inc files to avoid this, but I'm hoping there is a better way to deal with that build issue.


jart (approved):

jart:
Thank you for doing this!

PR #706, opened 2022-12-20
Make the `curl` example into a complete tool and add it as a cocmd builtin

Ronsor:
This pull requests makes the `examples/curl.c` example into a complete tool located in `tool/curl/`. Additionally, it is now added as a cocmd builtin and can be used with `system()`/`popen()` by including it with `_STATIC_YOINK("_curl")`.


jart (approved):
Thank you!

Issue #713, opened 2023-01-02
Cannot set argv[0] on New Technology for execve


G4Vi:
`mkntcmdline` always ignores `argv[0]` and instead inserts the program path into the command line. This prevents Cosmopolitan from properly invoking programs that dispatch based on `argv[0]` as they will always be invoked with their path as the first token in the command line instead of `argv[0]`

This test program fork bombs when the `argv[0]` passed to `exec` is not set for the new process:
```C
#include "libc/calls/calls.h"
#include "libc/stdio/stdio.h"
#include "libc/str/str.h"
#include "libc/calls/weirdtypes.h"

int main(int argc, char *argv[]) {
    if(strcmp(argv[0], "child") == 0) {
        printf("child, argv[0], exiting\n");
    }
    else {
        pid_t cpid = fork();
        if(cpid == -1) {
            printf("fork failed\n");
        }
        else if(cpid == 0) {
            printf("I am child\n");
            char *newargv[] = {"child", NULL};
            execv(argv[0], newargv);
            printf("why am i still alive\n");
        }
        else {
            printf("waiting for %d\n", cpid);
            int wstatus;
            printf("waitpid return %d\n", waitpid(cpid, &wstatus, 0));
        }
    }
}
```

The `CreateProcessW` documentation states "C programmers generally repeat the module name as the first token in the command line." leading me to believe it is not a requirement for the first part in the command line to be the application name.

APPerl depends on this functionality in order to allow easily invoke multiple scripts backed into the same binary,

G4Vi:
Fixed by https://github.com/jart/cosmopolitan/commit/10d1c6da187e2c7b8d50535a7701c0b944ded8c8 . Thank you.

PR #715, opened 2023-01-03
add arg0 passing execve test

G4Vi:
Safeguard for verifying https://github.com/jart/cosmopolitan/issues/713 stays fixed.

### DIFF
--- a/test/libc/calls/execve_test.c
+++ b/test/libc/calls/execve_test.c
@@ -25,6 +25,7 @@
 #include "libc/testlib/testlib.h"
 
 #define N 16
+static const char *testArg0PassingArg = "bothArgsShouldBeThis";
 
 char *GenBuf(char buf[8], int x) {
   int i;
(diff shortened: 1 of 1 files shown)


jart (approved):
Thanks!

jart:
Any idea why the `MODE=tiny` tests are failing?

G4Vi:
I have no idea, I just ran `MODE=tiny`locally and it didn't fail for me. In the actions fail, somehow argv[0] was still program name in the failing test.

jart:
I don't think we're going to be able to merge this PR. There's a lot of different ways a Cosmo binary can get launched. Overriding argv[0] like this is pretty janky. Even if we worked out any possible bugs, breaking argv[0] breaks other things, like GetProgramExecutableName() which is needed for --strace and lots of other runtime magic. UNIX simply doesn't provide a guaranteed way to get the executable name without it. Even /proc/self/exe and AT_EXECFN aren't appropriate for this purpose, since executables can be interpreted. I'm also reasonably certain some operating systems like OpenBSD might not permit argv[0] munging at all, so our multi-system testing system wouldn't be able to run this test.

I recommend not using this as a coding technique if possible. I recommend considering any alternative, like an environment variable.

PR #717, opened 2023-01-10
Avoid matching directories when searching PATH on Windows

alganet:
The issue: `commandv()` would return directory names in the PATH as executable:

Here I'm testing in indirectly through `cocmd`:

```cmd
PS T:\Shared> .\cocmd.com --strace -c "Cursors"
SYS  15448            902'983 bell system five system call support 4'260'106'890 magnums loaded on the new technology
SYS  15448          1'015'706 __morph_begin()
SYS  15448          1'647'322 __morph_end()
SYS  15448          1'794'645 getenv("TERM") → NULL
SYS  15448          1'957'984 GetFileSecurity(u"C:\\Windows\\system32\\Cursors.com") failed:  errno=2
SYS  15448          2'044'688 faccessat(AT_FDCWD, "/C/Windows/system32/Cursors.com", 024000000000, 0) → -1 errno= 2
SYS  15448          2'161'393 GetFileSecurity(u"C:\\Windows\\Cursors.com") failed:  errno=2
SYS  15448          2'274'419 faccessat(AT_FDCWD, "/C/Windows/Cursors.com", 024000000000, 0) → -1 errno= 2
SYS  15448          2'423'230 GetFileSecurity(u"C:\\Windows\\system32\\Cursors.exe") failed:  errno=2
SYS  15448          2'541'540 faccessat(AT_FDCWD, "/C/Windows/system32/Cursors.exe", 024000000000, 0) → -1 errno= 2
SYS  15448          2'683'636 GetFileSecurity(u"C:\\Windows\\Cursors.exe") failed:  errno=2
SYS  15448          2'790'506 faccessat(AT_FDCWD, "/C/Windows/Cursors.exe", 024000000000, 0) → -1 errno= 2
SYS  15448          2'900'983 GetFileSecurity(u"C:\\Windows\\system32\\Cursors") failed:  errno=2
SYS  15448          3'014'082 faccessat(AT_FDCWD, "/C/Windows/system32/Cursors", 024000000000, 0) → -1 errno= 2
SYS  15448          3'189'455 faccessat(AT_FDCWD, "/C/Windows/Cursors", 024000000000, 0) → 0 errno= 2
SYS  15448          3'298'196 commandv("Cursors", 0x70000001fb60, 1'024) → "/C/Windows/Cursors"
SYS  15448          3'422'618 execve("/C/Windows/Cursors", {"Cursors", NULL}, {"=T:=T:\\", "ALLUSERSPROFILE=/C/ProgramData", "APPDATA=/C/Users/Enki/AppData/Roaming", "COMMONPROGRAMFILES=/C/Program Files/Common Files", "COMMONPROGRAMFILES(X86)=/C/Program Files (x86)/Common Files", "COMMONPROGRAMW6432=/C/Program Files/Common Files", "COMPUTERNAME=HARAM-TEPE", "COMSPEC=/C/Wi...) → ...
SYS  15448          3'867'212 panic: unrecoverable ntspawn("/C/Windows/Cursors") error:  errno=5
```

Without `--strace`, it fails silently:

```cmd
PS T:\Shared> .\cocmd.com  -c "Cursors && echo 1"
PS T:\Shared>
```

After the fix, it displays the error message from `execve()`:

```cmd
PS T:\Shared> .\cocmd.com  -c "Cursors && echo 1"
Cursors: execve() failed: No such file or directory (2)
```

My fix is naive, ~~I'm checking if the basename of the path contains a dot. Files without an extension are never executable on Windows. I've also changed the tests to reflect the change, I ran `commandv_test` on Windows as well and the entire suite on Linux.~~ (EDIT: I have added and additional `isdirectory_nt` for directory names containing a dot).

Ideally, we could improve this function checking for the `PATHEXT` Windows variable, which holds extensions for executable programs, but for my use case `.com` and `.exe` which are already implemented is enough.

Maybe these changes should go on `ntaccesscheck`? I'm not sure. Also, there is a lot of things there that I don't understand yet. Please let me know if that is the case and I'll work in it.

### DIFF
--- a/libc/calls/commandv.c
+++ b/libc/calls/commandv.c
@@ -16,16 +16,18 @@
 │ TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR             │
 │ PERFORMANCE OF THIS SOFTWARE.                                                │
 ╚─────────────────────────────────────────────────────────────────────────────*/
-#include "libc/intrin/bits.h"
-#include "libc/intrin/safemacros.internal.h"
 #include "libc/calls/calls.h"
-#include "libc/intrin/strace.internal.h"
+#include "libc/calls/struct/stat.h"
 #include "libc/dce.h"
 #include "libc/errno.h"
+#include "libc/intrin/bits.h"
+#include "libc/intrin/safemacros.internal.h"
+#include "libc/intrin/strace.internal.h"
 #include "libc/log/libfatal.internal.h"
 #include "libc/runtime/runtime.h"
 #include "libc/str/str.h"
 #include "libc/sysv/consts/ok.h"
+#include "libc/sysv/consts/s.h"
 #include "libc/sysv/errfuns.h"
 
 static bool IsExePath(const char *s, size_t n) {
@@ -48,6 +50,9 @@ static bool AccessCommand(const char *name, char *path, size_t pathsz,
                           size_t pathlen) {
   size_t suffixlen;
   suffixlen = strlen(suffix);
+  if (IsWindows() && suffixlen == 0 && !IsExePath(name, namelen) &&
+      !IsComPath(name, namelen))
+    return false;
   if (pathlen + 1 + namelen + suffixlen + 1 > pathsz) return false;
   if (pathlen && (path[pathlen - 1] != '/' && path[pathlen - 1] != '\\')) {
     path[pathlen] = !IsWindows()                  ? '/'
@@ -57,7 +62,16 @@ static bool AccessCommand(const char *name, char *path, size_t pathsz,
   }
   memcpy(path + pathlen, name, namelen);
   memcpy(path + pathlen + namelen, suffix, suffixlen + 1);
-  if (!access(path, X_OK)) return true;
+  if (!access(path, X_OK)) {
+    struct stat st;
+    if (!stat(path, &st)) {
+      if (S_ISREG(st.st_mode)) {
+        return true;
+      } else {
+        errno = EACCES;
+      }
+    }
+  }
   if (errno == EACCES || *err != EACCES) *err = errno;
   return false;
 }
--- a/test/libc/calls/commandv_test.c
+++ b/test/libc/calls/commandv_test.c
@@ -16,15 +16,16 @@
 │ TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR             │
 │ PERFORMANCE OF THIS SOFTWARE.                                                │
 ╚─────────────────────────────────────────────────────────────────────────────*/
-#include "libc/intrin/safemacros.internal.h"
 #include "libc/calls/calls.h"
 #include "libc/calls/struct/dirent.h"
 #include "libc/calls/struct/stat.h"
 #include "libc/dce.h"
+#include "libc/errno.h"
 #include "libc/fmt/fmt.h"
+#include "libc/intrin/safemacros.internal.h"
 #include "libc/log/check.h"
-#include "libc/mem/mem.h"
 #include "libc/mem/gc.internal.h"
+#include "libc/mem/mem.h"
 #include "libc/runtime/internal.h"
 #include "libc/runtime/runtime.h"
 #include "libc/stdio/stdio.h"
@@ -58,7 +59,12 @@ void TearDown(void) {
 
 TEST(commandv, testPathSearch) {
   EXPECT_NE(-1, touch("bin/sh", 0755));
-  EXPECT_STREQ("bin/sh", commandv("sh", pathbuf, sizeof(pathbuf)));
+  if (IsWindows()) {
+    EXPECT_EQ(NULL, commandv("sh", pathbuf, sizeof(pathbuf)));
+    EXPECT_EQ(errno, ENOENT);
+  } else {
+    EXPECT_STREQ("bin/sh", commandv("sh", pathbuf, sizeof(pathbuf)));
+  }
 }
 
 TEST(commandv, testPathSearch_appendsComExtension) {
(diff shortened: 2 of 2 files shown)


jart:
Thanks for your patience. @G4Vi could you take a look at this change? Have you run into this issue?

G4Vi:
@jart I tested  `cocmd` on Linux and the same issue is present there.  `commandv` is happy to return directories even though they are not `execve()'able paths`.  The issue is `access(path, X_OK)` is an insufficient check for an `execve()'able path` as directories usually have the executable bit set in order to allow accessing items in a directory.

@alganet Rather than calling `isdirectory_nt`, add a call to one of the `stat` variants to check if `st_mode` corresponds to a compatible file type to add some verification for all platforms.

jart:
@G4Vi Makes sense. 

I'm thinking this might be the best approach:

```c
  if (!access(path, X_OK)) {
    struct stat st;
    if (!stat(path, &st)) {
      if (S_ISREG(st.st_mode)) {
        return true;
      } else {
        errno = EISDIR;
      }
    }
  }
```

I think access() should be called first, because it is my belief (based on https://blog.aaronballman.com/2011/08/how-to-check-access-rights/ which is the origin of our access() implementation) that this will ensure no windows security popups happen in the event we traverse over some restricted directory that happens to be on path.

@alganet Could you send an email to jtunney@gmail.com saying "I intend to assign
you the copyright to the changes I contribute to Cosmopolitan" please?
See CONTRIBUTING.md. Please link to this GitHub pull request in your
email. Afterwards, please reply to this pull request saying "I sent
it!". I ask this of everyone. It only has to happen once. It only
applies to the changes you *choose* to contribute. We need it to protect
the project. The way we're protecting it is by keeping a record that
everyone's wishes are being respected. Please note that in order to
assign copyright, it has to be yours to assign. For example, if you have
a job where your employment contract states your employer owns
everything you do, then you can't assign copyright on your own. In that
case, what I recommend doing is writing an email to your manager telling
them what you're doing, and making sure he/she feels comfortable with
it. That way if an issue ever comes up, we'll have a record that you're
OK with the ownership of Cosmopolitan, and that your employer is OK with
it too.

libc/calls/commandv.c
```diff
@@ -57,6 +59,10 @@ static bool AccessCommand(const char *name, char *path, size_t pathsz,
   }
   memcpy(path + pathlen, name, namelen);
   memcpy(path + pathlen + namelen, suffix, suffixlen + 1);
+  if (IsWindows() &&
+          ((*suffix == '\0' && memchr(basename(name), '.', namelen) == NULL)) ||
```
G4Vi, line 63: Rather then determining the basename and searching the string. Why not compare with directly what we want:
```c
((suffixlen == 0) && !IsExePath(name, namelen) && !IsComPath(name, namelen))
```
Minor nitpick, checking suffixlen is a better test of an empty string as it doesn't require dereferencing a pointer first.

libc/calls/commandv.c
```diff
@@ -57,6 +59,10 @@ static bool AccessCommand(const char *name, char *path, size_t pathsz,
...
+      isdirectory_nt(path))
```
G4Vi, line 64: See comments in PR on adapting this check for all platforms.

G4Vi (changes requested):
Checking extension on Windows is a great idea for `commandv` as the `execve` implementation is unrecoverable and so it's more critical to get it right there.

If you want to check `PATHEXT` on Windows, that would be a great addition, but is not required now.

Thanks for adding more tests.

jart (approved):
Thanks for sending the email. Everything looks good to me. Waiting on @G4Vi to approve and merge.

libc/calls/commandv.c
```diff
@@ -16,16 +16,20 @@
 │ TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR             │
 │ PERFORMANCE OF THIS SOFTWARE.                                                │
 ╚─────────────────────────────────────────────────────────────────────────────*/
-#include "libc/intrin/bits.h"
-#include "libc/intrin/safemacros.internal.h"
 #include "libc/calls/calls.h"
-#include "libc/intrin/strace.internal.h"
+#include "libc/calls/struct/stat.h"
+#include "libc/calls/syscall_support-nt.internal.h"
```
G4Vi, line 21: I don't think this header is needed anymore: `#include "libc/calls/syscall_support-nt.internal.h"`

libc/calls/commandv.c
```diff
@@ -16,16 +16,20 @@
...
 #include "libc/dce.h"
 #include "libc/errno.h"
+#include "libc/fmt/conv.h"
```
G4Vi, line 24: `#include "libc/fmt/conv.h"` is also no longer needed.

libc/calls/commandv.c
```diff
@@ -57,7 +64,16 @@ static bool AccessCommand(const char *name, char *path, size_t pathsz,
...
   memcpy(path + pathlen, name, namelen);
   memcpy(path + pathlen + namelen, suffix, suffixlen + 1);
-  if (!access(path, X_OK)) return true;
+  if (!access(path, X_OK)) {
+    struct stat st;
+    if (!stat(path, &st)) {
+      if (S_ISREG(st.st_mode)) {
+        return true;
+      } else {
+        errno = EISDIR;
```
G4Vi, line 73: Nitpick: @jart did give `EISDIR` as an example, but `EACCES` looks like a better fit to me as `execve` uses `EACCES` for "The file or a script interpreter is not a regular file." and just because it isn't a regular file doesn't mean it's a directory. https://man7.org/linux/man-pages/man2/execve.2.html#ERRORS

G4Vi (changes requested):
Just a couple minor changes and this will be good to go, mostly just cleanup from the original fix that was replaced.

libc/calls/commandv.c
G4Vi: It isn't obvious to me what causes the difference in behavior, I can take a look at it either this evening or tomorrow. `EACCES` could be useful if they thought a executable with that name was in path, but there is only a directory, `ENOENT` seems wrong if anything exists by that name.

G4Vi: I don't think it's a showstopper, but I marked it for me to look into it #762

PR #722, opened 2023-01-28
Fix filenames in help.txt

jchulce:
Many functions in tool/net/help.txt incorrectly direct users to see a .c file
Reference the correct .S files instead


jart (approved):
Thanks! I'll update the website when I hit merge.

jart:
Website updated. Thanks again!

PR #725, opened 2023-02-01
Implement zipos mmap

G4Vi:
- Implemented via creating `MAP_ANONYMOUS` map and copying.
- `MAP_SHARED` disallowed for simplicity and because zipos files are read only
- Discovered some `mmap` tests didn't fail when the child process failed, added checks.

### DIFF
--- a/libc/zipos/mmap.c
+++ b/libc/zipos/mmap.c
@@ -48,14 +48,17 @@
  */
 void *__zipos_mmap(void *addr, size_t size, int prot, int flags,
                    struct ZiposHandle *h, int64_t off) {
-  if(!(flags & MAP_PRIVATE) || (flags & ~(MAP_PRIVATE | MAP_FILE | MAP_FIXED | MAP_FIXED_NOREPLACE)) ||
-  (!!(flags & MAP_FIXED) ^ !!(flags & MAP_FIXED_NOREPLACE))) {
-    STRACE("zipos mappings currently only support MAP_PRIVATE with select flags");
+  if (!(flags & MAP_PRIVATE) ||
+      (flags & ~(MAP_PRIVATE | MAP_FILE | MAP_FIXED | MAP_FIXED_NOREPLACE)) ||
+      (!!(flags & MAP_FIXED) ^ !!(flags & MAP_FIXED_NOREPLACE))) {
+    STRACE(
+        "zipos mappings currently only support MAP_PRIVATE with select flags");
     return VIP(einval());
   }
 
   const int tempProt = !IsXnu() ? prot | PROT_WRITE : PROT_WRITE;
-  void *outAddr = mmap(addr, size, tempProt, (flags & (~MAP_FILE)) | MAP_ANONYMOUS, -1, 0);
+  void *outAddr =
+      mmap(addr, size, tempProt, (flags & (~MAP_FILE)) | MAP_ANONYMOUS, -1, 0);
   if (outAddr == MAP_FAILED) {
     return MAP_FAILED;
   }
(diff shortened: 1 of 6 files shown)


test/libc/runtime/mmap_test.c
```diff
@@ -219,6 +221,86 @@ TEST(isheap, mallocOffset) {
   ASSERT_TRUE(_isheap(p + 100000));
 }
 
+static const char *ziposLifePath = "/zip/life.elf";
+TEST(mmap, ziposCannotBeAnonymous) {
```
jart, line 225: Thank you so much for putting in the additional effort to write tests.

libc/zipos/mmap.c
```diff
@@ -0,0 +1,79 @@
...
+ */
+void *__zipos_mmap(void *addr, size_t size, int prot, int flags,
+                   struct ZiposHandle *h, int64_t off) {
+  if (VERY_UNLIKELY(!!(flags & MAP_ANONYMOUS))) {
+    STRACE("MAP_ANONYMOUS zipos mismatch");
+    return VIP(einval());
+  }
+
+  // MAP_SHARED for non-writeable pages could be implemented, but would require
+  // keeping track of zipos pages to prevent mprotect(addr,len,PROT_WRITE)
```
jart, line 57: One thing we might consider doing in the future is checking if the zip local file content is 64kb aligned and uncompressed, in which case it can be mapped directly. That would require special tools though since the `zip` command for instance is unlikely to do this.

Strictly speaking, POSIX says that you have to call `msync()` if you want a file or shared mapping to propagate to disk (and possibly for other processes too?). So if you wanted to have shared mappings, you could have `msync()` be implemented to propagate changes.

libc/zipos/mmap.c
```diff
@@ -0,0 +1,79 @@
...
+  if (VERY_UNLIKELY(!!(flags & MAP_SHARED))) {
+    STRACE("MAP_SHARED on zipos");
+    return VIP(eacces());
+  }
+
+  const int tempProt = !IsXnu() ? prot | PROT_WRITE : PROT_WRITE;
+  void *outAddr = mmap(addr, size, tempProt, flags | MAP_ANONYMOUS, -1, 0);
```
jart, line 64: Passing flags here is definitely the wrong thing to do. I would say be much stricter about flags. For example:

```c
if (!(flags == (MAP_FILE | MAP_PRIVATE) ||
      (flags == MAP_FILE && (~prot & PROT_WRITE)))) {
  STRACE("zipos mappings currently only support MAP_PRIVATE or non-writabl0e MAP_FILE");
  return einval();
}
```

jart (changes requested):
Nice!

libc/zipos/mmap.c
jart: One additional note: You would want to support MAP_FIXED and MAP_FIXED_NOREPLACE. Those are the only other flags I could foresee being used. All other flags should be disallowed.

libc/zipos/mmap.c
G4Vi: That's a good idea for the future, supporting uncompressed writes would be a good starting point for writeable zipos.

I also thought about how zipos could decompress a file on demand instead of at open. Map out at least the size of the uncompressed file with `MAP_ANONYMOUS`, but with `PROT_NONE` and then on page faults decompress and `mprotect` as needed.

test/libc/runtime/mmap_test.c
G4Vi: It's amazing how much they ease development.

jart (approved):
Thank you!

PR #727, opened 2023-02-03
zipos fexecve

G4Vi:
Maps zipos to a real file descriptor for `fexecve`. You can now run ELF binaries embedded in `/zip`.

Supports Linux via `memfd_create`. FreeBSD via `shm_open(SHM_ANON` (available since FreeBSD 8.0), not tested.   (all `fexecve` platforms) 

Adding support for more sysv systems to the mapping function `__zipos_to_fd` is possible via various `shm_`.

### DIFF
--- a/libc/calls/fexecve.c
+++ b/libc/calls/fexecve.c
@@ -22,6 +22,7 @@
 #include "libc/calls/calls.h"
 #include "libc/calls/cp.internal.h"
 #include "libc/calls/execve-sysv.internal.h"
+#include "libc/calls/internal.h"
 #include "libc/calls/struct/stat.internal.h"
 #include "libc/calls/syscall-sysv.internal.h"
 #include "libc/dce.h"
@@ -105,13 +106,18 @@ static bool ape_to_elf(void *ape, const size_t apesize) {
   return false;
 }
 
-static int ape_fd_to_mem_elf_fd(const int infd, char *path) {
+/**
+ * Creates a memfd and copies fd to it.
+ *
+ * This does an inplace conversion of APE to ELF when detected!!!!
+ */
+static int fd_to_mem_fd(const int infd, char *path) {
   if (!IsLinux() && !IsFreebsd() || !_weaken(mmap) || !_weaken(munmap)) {
     return enosys();
   }
 
   struct stat st;
-  if (sys_fstat(infd, &st) == -1) {
+  if (fstat(infd, &st) == -1) {
     return -1;
   }
   int fd;
@@ -133,7 +139,7 @@ static int ape_fd_to_mem_elf_fd(const int infd, char *path) {
       ((space = _weaken(mmap)(0, st.st_size, PROT_READ | PROT_WRITE, MAP_SHARED,
                               fd, 0)) != MAP_FAILED)) {
     ssize_t readRc;
-    readRc = sys_pread(infd, space, st.st_size, 0, 0);
+    readRc = pread(infd, space, st.st_size, 0);
     bool success = readRc != -1;
     if (success && (st.st_size > 8) && IsAPEMagic(space)) {
       success = ape_to_elf(space, st.st_size);
@@ -156,7 +162,7 @@ static int ape_fd_to_mem_elf_fd(const int infd, char *path) {
  * Executes binary executable at file descriptor.
  *
  * This is only supported on Linux and FreeBSD. APE binaries are
- * supported.
+ * supported. Zipos is supported.
  *
  * @param fd is opened executable and current file position is ignored
  * @return doesn't return on success, otherwise -1 w/ errno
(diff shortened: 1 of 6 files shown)


libc/calls/fexecve.c
```diff
@@ -47,7 +51,11 @@ int fexecve(int fd, char *const argv[], char *const envp[]) {
   } else {
     STRACE("fexecve(%d, %s, %s) → ...", fd, DescribeStringList(argv),
            DescribeStringList(envp));
-    if (IsLinux()) {
+    if (__isfdkind(fd, kFdZip) &&
```
jart, line 54: Are you planning to implement execve() too?

libc/zipos/to_fd.c
```diff
@@ -0,0 +1,80 @@
...
+
+  const int64_t beforeOffset = __zipos_lseek(h, 0, SEEK_CUR);
+  struct stat st;
+  if ((beforeOffset == -1) || (__zipos_fstat(h, &st) == -1)) {
+    return -1;
+  }
+  int fd;
+  if (IsLinux()) {
+    fd = memfd_create(__func__, MFD_CLOEXEC);
+  } else /*if(IsFreebsd())*/ {
```
jart, line 56: I recommend having the `IsFreebsd()` here and perhaps returning `enosys()` in the else clause.

libc/zipos/to_fd.c
```diff
@@ -0,0 +1,80 @@
...
+  if (fd == -1) {
+    return -1;
+  }
+  void *space;
+  if ((ftruncate(fd, st.st_size) != -1) &&
+      ((space = mmap(0, st.st_size, PROT_READ | PROT_WRITE, MAP_SHARED, fd,
+                     0)) != MAP_FAILED)) {
+    const bool success =
+        (__zipos_read(h, &(struct iovec){space, st.st_size}, 1, 0) != -1) &&
+        (__zipos_lseek(h, beforeOffset, SEEK_SET) != -1);
```
jart, line 68: `__zipos_lseek` shouldn't be necessary since `__zipos_read` only changes the file position if the last argument is `-1`. Since you're specifying `0` as the last arg, `__zipos_read` will have `pread()` behavior instead.

libc/zipos/to_fd.c
```diff
@@ -0,0 +1,80 @@
...
+  }
+  if (fd == -1) {
+    return -1;
+  }
+  void *space;
+  if ((ftruncate(fd, st.st_size) != -1) &&
+      ((space = mmap(0, st.st_size, PROT_READ | PROT_WRITE, MAP_SHARED, fd,
+                     0)) != MAP_FAILED)) {
+    const bool success =
+        (__zipos_read(h, &(struct iovec){space, st.st_size}, 1, 0) != -1) &&
```
jart, line 67: [optional] You might want to have an `_npassert()` possibly to check that the result of `__zipos_read` is equal to `st.st_size`

test/libc/calls/fexecve_test.c
```diff
@@ -86,3 +88,12 @@ TEST(fexecve, memfd_create) {
...
   EXITS(42);
 }
+
+TEST(fexecve, zipos) {
+  if (!IsLinux() && !IsFreebsd()) return;
+  int fd = open("/zip/life.elf", O_RDONLY);
+  SPAWN(vfork);
+  if (fd == -1 && errno == ENOSYS) _Exit(42);
+  fexecve(fd, (char *const[]){0}, (char *const[]){0});
+  EXITS(42);
```
jart, line 98: sweeet

jart (changes requested):
Nice change! It's a shame we can't easily and safely support more platforms for this.

libc/calls/fexecve.c
G4Vi: Didn't even consider that it could be polyfilled with similar implementation to for `fexecve`. Added for elf on Linux and FreeBSD. Added (with fd leaking for APE on Linux).

libc/zipos/to_fd.c
G4Vi: That's handy and simplifies the `mmap` too

G4Vi:
Added `execve` and attempted to improve the zipos interface. You can now run APEs from zipos on Linux though it leaks a `fd`, maybe the APE loader could be setup to cleanup various shared memory at startup.

jart:
Leaking file descriptors is a big no-no for C library functions.

Where is it leaking?

1. Does the file descriptor count increase in the parent process after execve() is called? If so, why can't it be closed?
2. Does the file descriptor leak into the child process? If so, why can't it be made `O_CLOEXEC`?

G4Vi:
> Does the file descriptor count increase in the parent process after execve() is called? If so, why can't it be closed?

I don't think so.
> Does the file descriptor leak into the child process? If so, why can't it be made `O_CLOEXEC`?

It leaks into the child process. It cannot be made `O_CLOEXEC` as then when the APE loader is launched, the path to `proc` (something like `/proc/self/fd/4`) no longer exists so it can't launch.

Thinking about this more, with ELF or APE, the in-mem path to the child executable must be preserved for the child's zipos to work. Making that work without leaking resources isn't trivial, so I think I'll save that for another PR.

> Leaking file descriptors is a big no-no for C library functions.

I will remove the run APE from zipos proof of concept, saving it for the getting zipos working aforementioned PR.

On the topic of writing good library functions, what do you think of the interface to `__zipos_to_mem_file` / `__zipos_uri_to_mem_file`? In my own code, I'd probably use `char (*path)[PATH_MAX]` instead of the `char *` to avoid buffer overflows, but I didn't notice that being used in this code base. `NULL` is a valid to pass there.

jart:
One thing you could do for the child process is use `fcntl(fd, F_DUPFD, 100)` to at least ensure it's granted a number that's high enough to not cause problems with code that assumes fixed file descriptor numbers.

The array boundary type signature thing shouldn't make much of a difference. If you build in `MODE=dbg` or `MODE=asan` then the runtime gets that information by other means automatically. C compilers aren't very good at statically analyzing this sort of thing. So I'd say favor simpler code.

G4Vi:
Moving this to draft.  Planning on first adding converting APE to ELF in memory as in 2dc68bc2ddb8df2e3ec484c7fd788de9ea3a0f45

jart:
Oh nice! I'm glad you discovered that! I wish I'd thought to recommend it.

G4Vi:
I reworked this on-top of the APE `fexecve` stuff. The zipos `execve` implementation just passes to `fexecve` for simplicity.

Additionally `fexecve` should always work with APEs with `FD_CLOEXEC` set. Rather than the APE loader from `binfmt_misc` failing with no recovery, it creates a memfd, copies to it with APE to ELF conversion, and runs from it in that case too. When kernel support is added for APE, that shouldn't be necessary on supporting kernels.

jart (approved):
Thank you!

PR #733, opened 2023-02-06
APE fexecve

G4Vi:
Via by copying to `memfd_create` / `shm_open(SHM_ANON`.

This relies on parsing APE shell printf string.

Edit: TODO: async signal safe and cancellation safe.
vfork safety isn't likely possible due to the dependence on the memory mapping.


jart (approved):
Thank you!

PR #735, opened 2023-02-06
zipos mmap - run entirely under __mmi_lock, hide implementation strace

G4Vi:
Makes the zipos mmap safer occurring entirely under the `__mmi_lock` just like `Mmap` and `Munmap` normally runs under.

Fixes zipos mmap showing implementation STRACEs, it now appears like normal mmap.


jart (approved):
Thank you!

PR #740, opened 2023-02-09
add * and ? operators to cocmd via musl glob

G4Vi:
Tilda not added as the musl glob does not support `GLOB_TILDA`.

`globfree` is never called as all cocmd paths end in `execve` or `exit`.


jart (approved):

jart:
Thank you!

PR #742, opened 2023-02-14
Use Kahan summation for fsum and fsumf

skeeto:
This [compensated summation](https://en.wikipedia.org/wiki/Kahan_summation_algorithm) gives a more precise result than the original algorithm, as demonstrated through extra precision in the tests. It also drops recursion and the extra branching.

I used double precision internally for fsumf since I assume its purpose is to interface with single precision, not necessary constrain itself to only using single precision. If needed, it can be trivially changed to use single precision internally while still passing the tests.

In fsumf, naive summation using double precision is likely sufficient, and more precise than single precision Kaham summation. In other words, Kahan summation with double precision in fsumf may be overkill.

In the fsumf test, neither 10000.1 nor 1.1 can be represented exactly. The former is rounded down and the latter is rounded up (by less) to the nearest representable value. As a result, the most precise sum is just shy of the "obvious" result.

* * *

Changing the tests isn't strictly necessary. I wanted tests that fail under the old implementation but pass with the new.

### DIFF
--- a/libc/tinymath/fsum.c
+++ b/libc/tinymath/fsum.c
@@ -24,8 +24,14 @@
  */
 double fsum(const double *p, size_t n) {
   size_t i;
-  double s;
-  if (n > 8) return fsum(p, n / 2) + fsum(p + n / 2, n - n / 2);
-  for (s = i = 0; i < n; ++i) s += p[i];
-  return s;
+  double err, sum, t, y;
+
+  sum = err = 0;
+  for (i = 0; i < n; ++i) {
+    y = p[i] - err;
+    t = sum + y;
+    err = (t - sum) - y;
+    sum = t;
+  }
+  return sum;
 }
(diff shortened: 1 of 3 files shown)


jart (approved):
Thank you! It makes me so happy to see a change to Cosmo from one of my favorite bloggers!

Could you send an email to jtunney@gmail.com from your [wellons@nullprogram.com](mailto:wellons@nullprogram.com) address saying "I intend to assign you the copyright to the changes I contribute to Cosmopolitan" please? See CONTRIBUTING.md. I ask it of everyone. It only has to happen once. It only applies to the changes you *choose* to contribute. We need it to protect the project. The way we're protecting it is by keeping a record that everyone's wishes are being respected. Please note that in order to assign copyright, it has to be yours to assign. For example, if you have a job where your employment contract states your employer owns everything you do, then you can't assign copyright on your own. In that case, what I recommend doing is writing an email to your manager telling them what you're doing, and making sure he/she feels comfortable with it. That way if an issue ever comes up, we'll have a record that you're OK with the ownership of Cosmopolitan, and that your employer is OK with it too.

PR #744, opened 2023-02-17
Fix nt path format check

JoshuaWierenga:
I think that's what the check should be, just wondering if it should be a letter check instead of an alphanumeric one since I don't think windows supports mounting to anything other than drive*letters* and ntfs directories.


jart (approved):
Thank you!

PR #749, opened 2023-02-24
Redbean sqlite fix multi close

pkulchenko:
@jart, this implements the fix for the issue of closing db after the statements are finalized that you came up with. I made it work for the current setup with the weak keys, as we discussed in the chat. This also implements additional config options as well as an improvement for session handling (as those shouldn't be GC collected given that they are released when the DB is closed).


jart (approved):
Amazing change! Thank you!

Issue #739, opened 2023-02-08
Configurable cache directive


johnmuhl:
When serving static assets using a cache-busting pattern (e.g. `/{timestamp}/file.ext`) it is better to use the `immutable` directive than `must-revalidate`. Attached is a patch to add `ProgramCacheDirective()` function which can be used to change it.

[0001-make-cache-directive-configurable.txt](https://github.com/jart/cosmopolitan/files/10689576/0001-make-cache-directive-configurable.txt)

pkulchenko:
I agree that this setting may need to be more configurable, but `immutable` is only better than `must-revalidate` for specific use cases (and also appears to be not widely supported yet: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control).

@jart, are you good with the proposed patch? Would it be better to use `ProgramCache` to set up both time and the directive?

johnmuhl:
Sorry I jumbled that up and agree `immutable` isn't strictly better. I meant that sometimes you want something other than `must-revalidate` and it'd be nice to have a way to configure that.

I think I'd prefer `ProgramCache(600, "stale-while-revalidate=86400")` but it's not much of a preference.

pkulchenko:
@johnmuhl, I submitted #750 based on your patch and our discussion.

jart:
Sounds good to me.

johnmuhl:
Thanks again.

PR #750, opened 2023-02-24
Make redbean cache directive configurable (thanks to @johnmuhl)

pkulchenko:
@jart, this implements setting Cache-Control directive as discussed in #739.

I've extended the logic a bit to allow an empty string to be used to remove any value and a null value to use the default (so using `ProgramCache(num)` keeps the currently set directive for compatibility). I also updated `-c` to accept "num", "num," (to set an empty directive), and "num,directive".


jart (approved):
LGTM

PR #752, opened 2023-02-25
memfd fexecve zipos support

G4Vi:
Uses environment variable `COSMOPOLITAN_INIT_ZIPOS` to pass fd. If provided, `__zipos_get` uses that to initialize instead of looking up  `GetProgramExecutableName` and `open`ing. The advantage to using a memfd rather than an `shm_open` file for this, is that it will be automatically cleaned up without an `unlink` though for other operating systems an `unlink` will be required.

One thing to note is https://github.com/G4Vi/cosmopolitan/blob/7ae7d0f1b327df52f1d87fc8172af5fda462bc70/libc/calls/fexecve.c#L236
I'm not sure if it's reasonable to use that much stack? `cocmd` uses that much space for `envp` as a `static`.

### DIFF
--- a/libc/calls/fexecve.c
+++ b/libc/calls/fexecve.c
@@ -190,6 +190,7 @@ int fexecve(int fd, char *const argv[], char *const envp[]) {
   } else {
     STRACE("fexecve(%d, %s, %s) → ...", fd, DescribeStringList(argv),
            DescribeStringList(envp));
+    int savedErr = 0;
     do {
       if (!IsLinux() && !IsFreebsd()) {
         rc = enosys();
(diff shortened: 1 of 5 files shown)


G4Vi:
Requesting review as this blocks polyfilling `fexecve` for other platforms.

ahgamut:
Is there some context for the upper bound of `envp` being 500? I'm asking because when we implemented `getservbyname`, we ended up with `DNS_NAME_MAX` (253) as an upper bound, allocating two `static char[DNS_NAME_MAX]` in the implementation.

G4Vi:
I switched the `envs` buffer to use `static _Thread_local` instead of stack.

jart (approved):
Looks good! Thanks!

PR #760, opened 2023-03-04
Fix redbean SQLite to report results on failure to open db

pkulchenko:
A small patch to fix the issue with not returning `nil` as the first parameter after a failure to open DB.


jart (approved):

jart:
Thank you!

PR #761, opened 2023-03-05
UPDATE: serve .mjs files as text/javascript

imawaki:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules#aside_%E2%80%94_.mjs_versus_.js


jart (approved):
LGTM Thanks.

PR #764, opened 2023-03-09
commandv: fix dir tests on non-NT

### LINKED ISSUE #762: commandv: Inconsistent behavior between NT and non=NT.
Adding this here for me to look into it later and not hold up the merge.
-----------------------------------------------------------------------
This test got really strange:

```c
TEST(commandv, test_DirPaths_wontConsiderDirectoriesExecutable2) {
  EXPECT_NE(-1, mkdir("this_is_a_directory.com", 0755));
  EXPECT_EQ(NULL,
            commandv("this_is_a_directory.com", pathbuf, sizeof(pathbuf)));
  if (IsWindows()) {
    EXPECT_EQ(errno, EACCES);
  } else {
    EXPECT_EQ(errno, ENOENT);
  }
}
```

I had to put the `if` to make it pass on linux and NT.

I'm not sure how to make non-NT return `EACCES` for this case, and I'm not comfortable with the errors being different across plaforms. Maybe `commandv` should always return `ENOENT`? If so, maybe we can remove this line:

```c
  if (errno == EACCES || *err != EACCES) *err = errno;
```

I'm kind of a newbie in C stuff. What benefit a caller would have by calling `commandv` and getting an `EACCES`? I mean, it's hard to point which one of the paths was not a regular file anyway. Maybe I'm missing something here!

_Originally posted by @alganet in https://github.com/jart/cosmopolitan/pull/717#discussion_r1126074481_

G4Vi:
Fixes #762

### DIFF
--- a/test/libc/calls/commandv_test.c
+++ b/test/libc/calls/commandv_test.c
@@ -118,18 +118,20 @@ TEST(commandv, testSameDir_willHappenWithColonBlank2) {
 }
 
 TEST(commandv, test_DirPaths_wontConsiderDirectoriesExecutable) {
+  CHECK_NE(-1, setenv("PATH", ":bin", true));
   EXPECT_NE(-1, mkdir("Cursors", 0755));
   EXPECT_EQ(NULL, commandv("Cursors", pathbuf, sizeof(pathbuf)));
-  EXPECT_EQ(errno, ENOENT);
+  if (IsWindows()) {
+    EXPECT_EQ(errno, ENOENT);
+  } else {
+    EXPECT_EQ(errno, EACCES);
+  }
 }
 
 TEST(commandv, test_DirPaths_wontConsiderDirectoriesExecutable2) {
+  CHECK_NE(-1, setenv("PATH", ":bin", true));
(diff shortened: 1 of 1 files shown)


G4Vi:
The previous tests were not failing for me, just when I ran on Debian Linux with `--strace` I discovered that it wasn't even checking the location where the created directory was, so I adjusted `PATH` and that fixes the issue of the unexpected inconsistent behavior between Windows and Linux on `test_DirPaths_wontConsiderDirectoriesExecutable2`. Only on Windows, `commandv` is supposed to search the current directory even if it's not in `PATH` according to the other tests in the test file. That behavior allowed Windows to fail with `EACCES` (as expected)  without the `PATH` modification.

On the new `test_DirPaths_wontConsiderDirectoriesExecutable`, the early return on Windows causes it to fail with the default errno `ENOENT` instead of `EACCES`. I don't think Windows failing with `EACCES` would be right as the check is done before querying the file system and `commandv` failing fast is probably better behavior then unnecessarily querying the file system.

jart (approved):

Issue #772, opened 2023-03-23
The printf family of functions doesn't pad numbers properly when the minus flag is present


GabrielRavier:
The following program:

```cpp
#include <stdio.h>

int main()
{
    printf("<%-4.3d>\n", 42);
}
```

outputs `<042 >` on the 4 libcs I have tested it on, whereas cosmopolitan instead outputs `<42  >`. This appears to be caused by cosmopolitan assuming that the presence of the `-` flag means printed numbers never need to be padded with 0s (see the code at the beginning of `__fmt_ntoa_format`), even though the C standard specifies as much:

> The precision specifies the minimum number of digits to appear; if the value being converted can be represented in fewer digits, **it is expanded with leading zeros**.

...and there is no indication anywhere that the minus flag changes this in any way.

GabrielRavier:
I've written a patch that *should* fix this (I'm not fully confident of it but it fixes this specific issue and passes cosmopolitan's tests along with the libc testsuiite I'm writing (in so far as it doesn't cause any regressions and fixes the tests that were failing due to this)).

```patch
From b7873d80eee6243d7c56b42838366d13157e5876 Mon Sep 17 00:00:00 2001
From: Gabriel Ravier <gabravier@gmail.com>
Date: Thu, 23 Mar 2023 18:22:05 +0100
Subject: [PATCH] Fix padding+minus flag on numbers for printf-family functions

The C standard states, for conversions using the d, i, b, B, o, u, x or X conversion specifiers:
> The precision specifies the minimum number of digits to appear; if
> the value being converted can be represented in fewer digits, it is
> expanded with leading zeros.
- C standard, 7.23.6.1. The fprintf function

However, cosmopolitan currently suppresses the addition of leading
zeros when the minus flag is set. This is not reflected by anything
within the C standard, meaning that behavior is incorrect.

This patch fixes this.
---
 libc/fmt/ntoa.c | 21 ++++++++++-----------
 1 file changed, 10 insertions(+), 11 deletions(-)

diff --git a/libc/fmt/ntoa.c b/libc/fmt/ntoa.c
index 8bf081768..116a9ed7a 100644
--- a/libc/fmt/ntoa.c
+++ b/libc/fmt/ntoa.c
@@ -34,17 +34,16 @@ static int __fmt_ntoa_format(int out(const char *, void *, size_t), void *arg,
                              unsigned char flags) {
   unsigned i;
   /* pad leading zeros */
[...]

PR #787, opened 2023-03-25
Fix issue (padding+minus flag on numbers for printf-family functions)

GabrielRavier:
(see also https://github.com/jart/cosmopolitan/issues/772)


G4Vi (approved):
LGTM, thank you!

Issue #778, opened 2023-03-24
The scanf family of functions doesn't implement the x conversion specifier properly


GabrielRavier:
The following program:

```cpp
#include <stdio.h>
#undef NDEBUG
#include <assert.h>

int main()
{
    int v = 0;
    assert(sscanf("0", "%x", &v) == 1);
}
```

crashes under cosmopolitan, when it should never do so. This appears to be due to an implementation error in `vcscanf` that is specifically triggered by the combination of attempting to parse `"0"` with `"%x"`.

Note: I'll submit a PR for this bug after doing so for the patches I've sent.

GabrielRavier:
Fixed by https://github.com/jart/cosmopolitan/pull/793

PR #793, opened 2023-03-29
Fix issue 778 (scanf of 0 with x conversion specifier)

GabrielRavier:
(see also https://github.com/jart/cosmopolitan/issues/778)


jart (approved):
Thank you!

Issue #783, opened 2023-03-24
wcscmp incorrectly handles large differences


GabrielRavier:
The following program:

```cpp
#include <stdio.h>
#include <wchar.h>
#include <limits.h>

int main()
{
    wchar_t str1[] = { WCHAR_MIN, L'\0' };
    wchar_t str2[] = { WCHAR_MAX, L'\0' };
    printf("%d\n", wcscmp(str1, str2));
}
```

is defined by the standard as printing a number that is smaller than 0, as the standard states:
> The wcscmp function returns an integer greater than, equal to, or less than zero, accordingly as the wide string pointed to by s1 is greater than, equal to, or less than the wide string pointed to by s2.
- C Standard, 7.31.4.4.1. The wcscmp function

and:
> Unless explicitly stated otherwise, the functions described in this subclause order two wide characters the same way as two integers of the underlying integer type designated by wchar_t.
- C Standard, 7.31.4.4. Wide string comparison functions

...and WCHAR_MIN is smaller than WCHAR_MAX, meaning the comparison should return a negative number. cosmopolitan instead prints a `1`.

GabrielRavier:
PS: wcscnmp has the same issue.

G4Vi:
Fixed by #795.

PR #795, opened 2023-03-29
Fix issue 783 (wcscmp difference larger than WCHAR_MAX)

GabrielRavier:
(see also https://github.com/jart/cosmopolitan/issues/783)


G4Vi (approved):
LGTM. Thank you.

Issue #784, opened 2023-03-24
fprintf doesn't handle write errors properly


GabrielRavier:
The following program:

```cpp
#include <stdio.h>

int main()
{
    FILE *fp = fopen("/dev/full", "w");

    setbuf(fp, NULL);
    printf("fprintf: %d\n", fprintf(fp, "test\n"));
    printf("fflush: %d\n", fflush(fp));
}
```

should print something like:

```
fprintf: -1
fflush: 0
```

or, at least

```
fprintf: 4
fflush: -1
```

cosmopolitan instead outputs this:

```
fprintf: 0
fflush: 0
```

It appears that fprintf is ignoring the write errors it gets from `fwrite`.

G4Vi:
Fixed by #798.

PR #798, opened 2023-03-31
Fix issue 784 (fprintf write error handling)

GabrielRavier:
(see also https://github.com/jart/cosmopolitan/issues/784)


G4Vi (approved):
LGTM. Thank you.

Issue #781, opened 2023-03-24
The printf family of functions doesn't handle large precisions properly


GabrielRavier:
The following program:

```cpp
#include <stdio.h>
#include <string.h>

int main()
{
    char buf2[512] = {};
    
    int i = snprintf(buf2, sizeof(buf2), "%.9999u", 10);
    printf("%d %zu\n", i, strlen(buf2));
}
```

Should always print `9999 511`, whereas cosmopolitan prints `144 144`. This is incorrect, and appears to be caused by the limited buffer size in the `__fmt_ntoa2` function of 144, which is easily hit by the large precision specified here.

PR #799, opened 2023-03-31
Fix issue 781 (printf large precisions for integers)

GabrielRavier:
(see also https://github.com/jart/cosmopolitan/issues/781)


G4Vi (approved):
LGTM. Thank you.

Issue #794, opened 2023-03-29
Support for unshare() mount() prctl() in Lua


ducktype:
Will be possibile ti add Lua api for this system functions: unshare() mount() prctl()
in the Linux platform insider the Unix lua module?

jart:
We've mostly been focusing our resources on the APIs we're able to make available on all our supported platforms. I'm love to have outstanding support for Linux-specific APIs too, since that will make things better for the people who only care about Linux. Contributions are welcome if anyone wants to tackle this.

konata-chan404:
It seems like a lovely addition for the API.
I'll take a look at it in the upcoming days🐈

PR #801, opened 2023-04-08
Redbean: Support for unshare() mount() prctl() in Lua 

konata-chan404:
These changes should extend Redbean's Lua's functionality to add additional unix syscalls per requested here #794.

I wasn't able to find any direct support for the ``unshare`` syscall in the Cosmopolitan C lib, so for now I am opening this pull request as incomplete without support for it. 
Few notes I have to leave for the maintainers:
* I made sure to add fields for the unix module for *all* flags and options for these syscalls, something that I perceived as something that pollutes the ``unix`` module so please let me know if there's any other way to implement it.
* Where can I add these syscalls to the documentation? And is there any need to add tests for them?
* I apologize in advance for any mistake whether it be a grave error or a styling mistake, this is my first PR for this project (and one the first for an open source project 💖)

### DIFF
--- a/third_party/lua/lunix.c
+++ b/third_party/lua/lunix.c
@@ -82,6 +82,7 @@
 #include "libc/sysv/consts/limits.h"
 #include "libc/sysv/consts/log.h"
 #include "libc/sysv/consts/map.h"
+#include "libc/sysv/consts/mount.h"
 #include "libc/sysv/consts/msg.h"
 #include "libc/sysv/consts/nr.h"
 #include "libc/sysv/consts/o.h"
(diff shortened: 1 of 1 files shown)


pkulchenko:
Overall, this PR looks good to me, but there are a couple of things that need to be fixed. This looks strange:

```
       if (lua_isnumber(L, 2)) {
            arg2 = lua_tointeger(L, 2);
        } else if (lua_isstring(L, 2)) {
            arg2 = (unsigned long) lua_tostring(L, 2);
        }
```

Why would the result of lua_tostring be converted to `unsigned long` if it's `char*`? I think what you really want to do is to use [lua_tointegerx](https://www.lua.org/manual/5.4/manual.html#lua_tointegerx) (or [lua_tonumberx](https://www.lua.org/manual/5.4/manual.html#lua_tonumberx) if needed), as it would convert to a number and tell you if the conversion was successful or not.

Also the closing bracket in the comments should have a slightly different character:

```
      └─→ nil, unix.Errno
```

> Where can I add these syscalls to the documentation? And is there any need to add tests for them?

Tests would be ideal, but these two may be difficult to test (although you may still be able to test for some basic parameters).

The documentation is in tool/net/help.txt; I can update it after this PR is merged or you can add it there yourself as part of the same PR.

konata-chan404:
Thank you so much for the lovely and relevant feedback 🐈
The Lua comment is really nice to learn and much better for sure, will get on all of the changes tomorrow.

konata-chan404:
I added corrections for the relevant feedback. 
Although I wasn't able to mitigate the ``(unsigned long)`` cast, the ``prctl`` syscall accepts arguments in the form of unsigned longs that are polymorphic in the sense they can be either a pointer or an integer. It does work in both cases so for now and I'm leaving it as it is and if there's something that is missing then I'll correct as needed🎇
Thank you so much.

jart (approved):
Approval for copyright.

third_party/lua/lunix.c
```diff
@@ -3196,6 +3199,71 @@ static int LuaUnixFdopendir(lua_State *L) {
...
+  int olderr = errno;
+  return SysretBool(
+    L, "mount", olderr,
+    mount(luaL_checkstring(L, 1), luaL_checkstring(L, 2), luaL_checkstring(L, 3),
+        luaL_optinteger(L, 4, 0), luaL_optstring(L, 5, NULL)));
+}
+
+// unix.umount(target: str[, flags: int)
+//     ├─→ true
+//     ├─→ nil, unix.Errrno
```
pkulchenko, line 3216: Minor tweak, but the character should be `└` (as n `prctl`).

third_party/lua/lunix.c
```diff
@@ -3196,6 +3199,71 @@ static int LuaUnixFdopendir(lua_State *L) {
   }
 }
 
+
+// unix.mount(source:str, target:str, filesystemtype: str[, mountflags: int[, data: str]])
+//     ├─→ true
+//     ├─→ nil, unix.Errno
```
pkulchenko, line 3205: Minor tweak, but the character should be `└` (as n `prctl`).

third_party/lua/lunix.c
```diff
@@ -3196,6 +3199,71 @@ static int LuaUnixFdopendir(lua_State *L) {
...
+static int LuaUnixMount(lua_State *L) {
+  int olderr = errno;
+  return SysretBool(
+    L, "mount", olderr,
+    mount(luaL_checkstring(L, 1), luaL_checkstring(L, 2), luaL_checkstring(L, 3),
+        luaL_optinteger(L, 4, 0), luaL_optstring(L, 5, NULL)));
+}
+
+// unix.umount(target: str[, flags: int)
```
pkulchenko, line 3214: Missing closing `]`.

third_party/lua/lunix.c
```diff
@@ -3196,6 +3199,71 @@ static int LuaUnixFdopendir(lua_State *L) {
...
+  int olderr = errno;
+  return SysretBool(
+    L, "unmount", olderr,
+    unmount(luaL_checkstring(L, 1), luaL_optinteger(L, 2, 0)));
+}
+
+// unix.prctl(option: int[, arg2: int[, arg3: int[, arg4: int[, arg5: int]]]])
+//     ├─→ result: int
+//     └─→ nil, unix.Errno
+static int LuaUnixPrctl(lua_State *L) {
```
pkulchenko, line 3227: None of the get-like prctl variants are going to work, as there is no way to get the values back to the Lua layer. These values either need to be returned are prctl results (it seems like most of them are stored as arg2, so may be need to be returned as a string or number or need to be taken as one table parameter, in which case the table can be modified to pass the values back. The first option is going to be more consistent with how the rest of unix.* functions operate.

I'm not familiar with the implementation details of prctl to access its coverage in cosmopolitan, but I'm sure @jart can comment on that.

pkulchenko (changes requested):

PR #803, opened 2023-04-16
unveil: Added truncate support on Linux 6.2+

GabrielRavier:
(see also commit message for more details).

Note that I have only been able to test this on a kernel with Landlock ABI version 3 support - my testing of kernels without it has been limited to manually modifying `landlock_create_ruleset`'s return statement to `return (flags == LANDLOCK_CREATE_RULESET_VERSION) && rc > 2 ? 2 : rc;` to test that the backup codepaths for the old ABI work properly, but this doesn't ensure that they also work on old kernels.

### DIFF
--- a/libc/calls/landlock.h
+++ b/libc/calls/landlock.h
@@ -26,6 +26,15 @@
  */
 #define LANDLOCK_ACCESS_FS_REFER 0x2000ul
 
+/**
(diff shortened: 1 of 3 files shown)


jart (approved):
Thank you!

PR #807, opened 2023-04-25
Fix the behavior of `redbean -i -e CODE`

jart:
When redbean is functioning as a Lua interpreter, the `-e` flag should behave the same way as other open source language interpreters. Namely it should exit after evaluating the code rather than showing the REPL.


pkulchenko (approved):
Looks good to me.

PR #813, opened 2023-05-11
Fix redbean to avoid compressing the first yielded chunk.

pkulchenko:
This may happen if no Status or Content-Type is set and if the written content is longer than 100B, the compression is going to be applied after coroutine.yield, even though it shouldn't be.

This patch checks for `isyielding` directly instead of keeping `istext` in sync with the state when compression is not needed.


jart (approved):
Thank you!

PR #817, opened 2023-05-14
Remove unused features from RedPajama Chat.

ingenieroariel:
In preparation of replacing them with the llama.com approach.


jart (approved):
Thank you!

PR #818, opened 2023-05-18
Redbean fetch keepalive

pkulchenko:
This change extends Fetch() to allow keep-alive requests. If requested, the socket will be kept open and stored in a table that can be passed to the next call to indicate which socket needs to be re-used for the same host connection. If the connection is configured to be closed or is closed by the server, then it's closed and removed from the socket table, so that any subsequent call will open it anew.


jart (approved):
Looks good. Doesn't seem like this will impact anyone who doesn't enable it. Thanks for the change!

pkulchenko:
> Doesn't seem like this will impact anyone who doesn't enable it.

Correct; it would need to be explicitly enabled. I also tested the server explicitly closing the connection, even when the client expects it to continue; should be handled correctly.

PR #819, opened 2023-05-22
Fix rgbtoxterm tool to output the proper colors

notwa:
I apologize for how random and minor this change is, but I was idly prodding through the executables from `make all` and noticed this tool wasn't working correctly. I'm guessing `rgb2tty` returned a byte in the past, but it presently returns a struct, so this printf has been pulling the value of the red channel instead of the xterm color code.


jart (approved):
Thank you!

PR #822, opened 2023-05-25
[metal] Fix some issues that were breaking UEFI boot

tkchia:


tkchia:
`examples/vga2.c` should now work again when booted from UEFI.

jart (approved):

PR #827, opened 2023-06-03
Extend Pledge with anet (same as inet, but with no connect)

pkulchenko:
This is useful for configurations where it's desirable to start redbean (or another socket listening application), but not to allow `connect` socket calls.

I moved `GetHostIps` in redbean before `GetOpts`, as this allows to pass `unix.pledge` as a command line option that may disable `connect` and still run `GetHostIps`, as it requires `connect`. It may still fail under external `pledge`, but that's by design.

I also slightly tweaked the message shown to make it a bit more clear on what may need to be done:

```
error: protected syscall connect (ord=02a); pledge promise 'inet' to allow
```

There is also similar "unix" promise that also includes connect, but I'm not sure if it also needs a corresponding promise to be added without connect for consistency. Maybe we can add one when there is a read use case for it...

### DIFF
--- a/libc/calls/pledge-linux.c
+++ b/libc/calls/pledge-linux.c
@@ -811,6 +811,22 @@ static const uint16_t kPledgeInet[] = {
     __NR_linux_getsockname,            //
 };
 
+// anet is similar to init, but without connect;
+// this allows to accept, but not initiate socket connections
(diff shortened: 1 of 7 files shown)


jart (approved):
Thank you!

PR #830, opened 2023-06-10
Fix syscall(2) returning -errno instead of using POSIX errno scheme

GabrielRavier:
(see commit message for more details)


jart (approved):

jart:
Thank you!

PR #832, opened 2023-06-12
Add path information in logged error messages

pkulchenko:
@jart, we discussed this some time ago and I think it should provide necessary information in the logs.


pkulchenko:
This doesn't add path to the logged messages from `ServeError` Lua calls. Can be extended to those call as well, but (1) it will be applied to all returned statuses and (2) it may not reflect the problem with the URL itself, but just some behavior of the Lua code, so I thought it's better to leave it this way.

jart (approved):
Thanks!

PR #837, opened 2023-06-23
introduce xxhash into third_party

fzakaria:
Introduce xxhash the Extremely fast non-cryptographic hash algorithm
https://github.com/Cyan4973/xxHash
commit 5b45185d0419f63eb5911133f3dc578a163d162f

```console
 ./o/third_party/xxhash/xxhsum.com -h
xxhsum.com 0.8.1 by Yann Collet 
Print or verify checksums using fast non-cryptographic algorithm xxHash 

Usage: xxhsum.com [options] [files] 

When no filename provided or when '-' is provided, uses stdin as input. 
```

### DIFF
--- a/third_party/xxhash/cli/README.md
+++ b/third_party/xxhash/cli/README.md
@@ -1,0 +1,5 @@
+// clang-format off
+This directory contains source code dedicated to the `xxhsum` command line utility,
+which is a user program of `libxxhash`.
+
+Note that, in contrast with the library `libxxhash`, the command line utility `xxhsum` ships with GPLv2 license.
--- a/third_party/xxhash/cli/xsum_config.h
+++ b/third_party/xxhash/cli/xsum_config.h
@@ -1,0 +1,243 @@
+// clang-format off
+/*
+ * xxhsum - Command line interface for xxhash algorithms
+ * Copyright (C) 2013-2021 Yann Collet
+ *
+ * GPL v2 License
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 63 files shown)


third_party/xxhash/cli/README.md
```diff
@@ -0,0 +1,5 @@
+// clang-format off
```
fzakaria, line 1: the python third-party script included these clang-format off to every file

third_party/xxhash/cli/xsum_config.h
```diff
@@ -0,0 +1,243 @@
...
+#endif
+
+/**
+ * 
+ * fzakaria: remove this
+ * 
+ * #if !defined(S_ISREG)
+ * #  define S_ISREG(x) (((x) & S_IFMT) == S_IFREG)
+ * #endif
+ */
```
fzakaria, line 113: I had to hand-edit the source here -- this macro becomes redefined by the libc the way the includes work.

third_party/xxhash/tests/Makefile
```diff
@@ -0,0 +1,132 @@
+// clang-format off
```
fzakaria, line 1: TODO: do we want to support tests in third_party?

jart (changes requested):
Could you create a README.cosmo file describing this package including where it came from? Also if it wouldn't be too much trouble, could you make this pull request consist of two commits: (1) the first commit should basically be copy/pasted upstream sources verbatim, and (2) the second commit should be what you have now. That way it's clearer where the third party code starts and where our local modifications begin.

fzakaria:
I noticed lz4cli also uses xxhash but has the source within itself -- that can be fixed after this.

third_party/xxhash/tests/Makefile
jart: Generally speaking, yes. It'd be very much welcome if you went to the extra effort to do that.

fzakaria:
Added the README.cosmo.

Trying to figure out the cleanest `git rebase` to introduce the commit earlier as you mentioned.

jart (approved):
Let's make progress on getting Mold into Cosmopolitan, now that you've got the process nailed down with zstd. Thanks for the contribution!

Issue #839, opened 2023-06-25
Windows: UDPv4 multicast/setsockopt() for IP_ADD_MEMBERSHIP does not work


jtru:
I am trying to implement a minimal UDP(v4) multicast receiver loop as an actually portable executable, and cannot get the resulting binary that ought to prove this is doable to work on Windows (major version 10, x86_64, last patched today). The same binary works fine on GNU/Linux and also OS X. I am using the [**Hermetic Monorepo Snapshot**](https://justine.lol/cosmopolitan/cosmopolitan.tar.gz) to build my executable.

I am actually not sure that it is actually cosmopolitan what's somehow at fault, or if it is me (who doesn't know anything relevant about Windows programming in general, or winsock in particular) holding something wrong... The problem arises as soon as I call `setsockopt()` to make the listener join an IP multicast group by means of `IP_ADD_MEMBERSHIP` in the following code:

```c
/*
 * mcast_recv.c, massaged for cosmopolitan libc, based on
 * listener.c -- joins a multicast group and echoes all data it receives from
 *              the group to its stdout...
 *
 * Antony Courtney,     25/11/94
 * Modified by: Frédéric Bastien (25/03/04)
 * to compile without warning and work correctly
 * Modified by: Johannes Truschnigg (2023-06-25)
 * to compile into an actually portable executable (https://justine.lol/ape.html)
 */


#include "libc/calls/calls.h"
#include "libc/macros.internal.h"
#include "libc/runtime/runtime.h"
#include "libc/sock/struct/ip_mreq.h"
#include "libc/sysv/consts/so.h"
#include "libc/sysv/consts/sock.h"
#include "libc/sysv/consts/ip.h"
#include "libc/sock/sock.h"
#include "libc/sysv/consts/inaddr.h"
#include "libc/stdio/stdio.h"
#include "libc/str/str.h"
#include "libc/sysv/consts/af.h"
#include "libc/sysv/consts/ipproto.h"
#include "libc/sysv/consts/sol.h"


#define HELLO_PORT 12345
#define HELLO_GROUP "239.255.255.250"
#define MSGBUFSIZE 2048

int main(int argc, char *argv[])
{
     struct sockaddr_in addr;
     int fd, nbytes;
     uint32_t addrlen;
     struct ip_mreq mreq;
     char msgbuf[MSGBUFSIZE];

     int yes=1;

     memset(msgbuf, 0, MSGBUFSIZE);

[...]

PR #842, opened 2023-06-26
Update Windows IP_MULTICAST constants to use Ws2tcpip.h values (#839)

pkulchenko:
@jart, this fixes the Windows IP_MULTICAST* constants to use Ws2tcpip.h values, as described by Microsoft in https://learn.microsoft.com/en-US/troubleshoot/windows/win32/header-library-requirement-socket-ipproto-ip. I also changed one 0x10 to 16 for consistency.


jart (approved):
Thanks!

Issue #841, opened 2023-06-25
Windows: recvfrom() support broken


jtru:
As originally discovered/reported in #839, `recvfrom()` appears broken on Windows hosts, and any code calling it on a UDP socket results in an error like this:

```
C:\Users\QEMU\Downloads>echoserver.com udp server 0 6000
udp server 0.0.0.0 6000
error:examples/echoserver.c:64:echoserver.com: check failed on DESKTOP-2485S1V pid 8860
        CHECK_NE(-1, (rc = recvfrom(sock, buf, sizeof(buf), 0, (struct sockaddr *)&addr2, &addrsize2)));
                 → -1 (-1)
                != -1 ((rc = recvfrom(sock, buf, sizeof(buf), 0, (struct sockaddr *)&addr2, &addrsize2)))
        EFAULT/487/Bad address/10014/The system detected an invalid pointer address in attempting to use a pointer argument in a call
        echoserver.com udp server 0 6000
note: won't print addr2line backtrace because no debug binary
70000003fe90 4048c5 NULL+0
70000003ff50 40456d NULL+0
70000003ffa0 403a4b NULL+0
70000003ffd0 403cb1 NULL+0
70000003ffe0 44e5ae NULL+0
```

Furthermore, I think the [`recvfrom()` API/spec](https://justine.lol/cosmopolitan/documentation.html#recvfrom) should explicitly support and document having both `struct sa* opt_out_srcaddr` and `unsigned int* opt_inout_srcaddrsize ` assume `NULL`, since `recvfrom(3p)` does that, too:

```man
[...]

SYNOPSIS
       #include <sys/socket.h>

       ssize_t recvfrom(int socket, void *restrict buffer, size_t length,
           int flags, struct sockaddr *restrict address,
           socklen_t *restrict address_len);

DESCRIPTION
[...]
       address     A  null  pointer,  or points to a sockaddr structure in which the sending address is to be stored.
                   The length and format of the address depend on the address family of the socket.

       address_len Either a null pointer, if address is a null pointer, or a pointer to a socklen_t object  which  on
                   input  specifies the length of the supplied sockaddr structure, and on output specifies the length
                   of the stored address.
[...]
```

pkulchenko:
[...]

PR #844, opened 2023-06-27
Fix recvfrom (#841)

pkulchenko:
This fixes a typo in the order of `recvfrom` parameters.

Just in case UDP client and server examples:

```
-- client
local sock = assert(unix.socket(unix.AF_INET, unix.SOCK_DGRAM, unix.IPPROTO_UDP))
assert(unix.connect(sock, ParseIp("127.0.0.1"), 12345))
print(assert(unix.send(sock,"this is a message")))

-- server
local sock = assert(unix.socket(unix.AF_INET, unix.SOCK_DGRAM, unix.IPPROTO_UDP))
assert(unix.bind(sock, ParseIp("127.0.0.1"), 12345))
print(assert(unix.recvfrom(sock)))
```

### DIFF
--- a/libc/intrin/wsarecvfrom.c
+++ b/libc/intrin/wsarecvfrom.c
@@ -66,8 +66,8 @@ textwindows int WSARecvFrom(
   }
 #else
   rc = __imp_WSARecvFrom(s, inout_lpBuffers, dwBufferCount,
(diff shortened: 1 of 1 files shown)


jart (approved):
Thank you!

Issue #334, opened 2021-11-29
Dependency in documentation


angeletakis:
Following the directions for the hello.com program in the README.md produces an error "run-detectors: unable to find an interpreter for ./hello.com"

<pre>(base) <font color="#26A269"><b>alex@alex-Z490-AORUS-PRO-AX </b></font><font color="#12488B"><b>~/down/cosmo $</b></font> <font color="#26A269">gcc</font> -g -Os -static -nostdlib -nostdinc -fno-pie -no-pie -mno-red-zone \
  -fno-omit-frame-pointer -pg -mnop-mcount \
  -o hello.com.dbg <u style="text-decoration-style:single">hello.c</u> -fuse-ld=bfd -Wl,-T,ape.lds \
  -include <u style="text-decoration-style:single">cosmopolitan.h</u> <u style="text-decoration-style:single">crt.o</u> <u style="text-decoration-style:single">ape.o</u> <u style="text-decoration-style:single">cosmopolitan.a</u>
/usr/local/bin/ld.bfd: warning: .note.gnu.build-id section discarded, --build-id ignored
(base) <font color="#26A269"><b>alex@alex-Z490-AORUS-PRO-AX </b></font><font color="#12488B"><b>~/down/cosmo $</b></font> <font color="#26A269">objcopy</font> -S -O binary <u style="text-decoration-style:single">hello.com.dbg</u> hello.com
(base) <font color="#26A269"><b>alex@alex-Z490-AORUS-PRO-AX </b></font><font color="#12488B"><b>~/down/cosmo $</b></font> <font color="#26A269">bash</font> -c <font color="#A2734C">&quot;./hello.com&quot;</font>                         
run-detectors: unable to find an interpreter for ./hello.com</pre>

I did some research and found https://issueexplorer.com/issue/jart/cosmopolitan/171
I executed
```bash
sudo sh -c "echo ':APE:M::MZqFpD::/bin/sh:' >/proc/sys/fs/binfmt_misc/register"
```
And it worked.
<pre>(base) <font color="#26A269"><b>alex@alex-Z490-AORUS-PRO-AX </b></font><font color="#12488B"><b>~/down/cosmo $</b></font> <font color="#26A269">bash</font> -c <font color="#A2734C">&quot;./hello.com&quot;</font>
hello world</pre>

I feel as this should be documented.

meiskam:
[...]

PR #846, opened 2023-07-02
Added missing dependency to readme install info

xet7:
Added missing dependency mentioned at https://github.com/jart/cosmopolitan/issues/334
to get installing Cosmopolitan working on Ubuntu 23.04 64bit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ It's recommended that Cosmopolitan be installed to `/opt/cosmo` and
 contains your non-monorepo artifacts.
 
 ```
+sudo sh -c "echo ':APE:M::MZqFpD::/bin/sh:' >/proc/sys/fs/binfmt_misc/register"
 sudo mkdir -p /opt
 sudo chmod 1777 /opt
 git clone https://github.com/jart/cosmopolitan /opt/cosmo


JoshuaWierenga:
This may also require adding ape/apeuninstall.sh right before ape/apeinstall.sh as it doesn't register with binfmt if it has already been done and specifically warns already having an entry that refers to /bin/sh instead of ape but that is the only option before running make. Can ape/apeinstall.sh be run before building works? If so we could move the mention of it in the readme to before make.

JoshuaWierenga:
I might be missing something but I just setup a Ubuntu 23.04 VM and with the following commands based on the readme I was able to produce a working binary without messing with binfmt:
```sh
sudo apt install -y git
sudo mkdir -p /opt
sudo chmod 1777 /opt
git clone https://github.com/jart/cosmopolitan /opt/cosmo
cd /opt/cosmo
build/bootstrap/make.com -j$(($(nproc) + 1)) toolchain
mkdir /opt/cosmos
cd ~
cat > hello.c << EOF
#include <stdio.h>

int main(void) {
  puts("Hello World");
}
EOF
/opt/cosmo/tool/scripts/cosmocc hello.c -o hello.com
```
The other commands about setting up /opt/cosmos/bin also work, they are just optional.

xet7:
@JoshuaWierenga 

For me, this additional line was needed to run first, so that I would not get errors when compiling.

JoshuaWierenga:
Something like this?
firefox_WS6s392RYv

JoshuaWierenga:
I just checked and on a newly checked out cosmopolitan repo, `ape/apeinstall.sh` works even before running make so I would recommend that over registering /bin/sh as the handler for ape binaries if for whatever reason linux refuses to run ape binaries. If you have previously built cosmo and then somehow break ape support then you will have to run `rm o/depend` in order to get `ape/apeinstall.sh` to work again as if it finds that file it assumes building works and will call make.

xet7:
@JoshuaWierenga 

Thanks! I added those improvements and moved this to new PR at https://github.com/jart/cosmopolitan/pull/850

PR #848, opened 2023-07-03
add support for __cxa_demangle

fzakaria:
This adds support for __cxa_demangle through the cxxabi.h file.
At the moment it is the only symbol included.

The source was taken from FreeBSD:
contrib/libcxxrt/libelftc_dem_gnu3.c
2176c9ab71c85efd90a6c7af4a9e04fe8e3d49ca

FreeBSD does say this is also almost verbatim from ELFToolkit

### DIFF
--- a/third_party/libcxx/libcxx.mk
+++ b/third_party/libcxx/libcxx.mk
@@ -67,6 +67,7 @@ THIRD_PARTY_LIBCXX_A_HDRS =					\
 	third_party/libcxx/ctype.h				\
 	third_party/libcxx/cwchar				\
 	third_party/libcxx/cwctype				\
+	third_party/libcxx/cxxabi.h				\
 	third_party/libcxx/deque				\
 	third_party/libcxx/errno.h				\
 	third_party/libcxx/exception				\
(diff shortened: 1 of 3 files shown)


third_party/libcxx/libcxx.mk
```diff
@@ -212,6 +214,16 @@ $(THIRD_PARTY_LIBCXX_A_OBJS): private				\
 			-ffunction-sections			\
 			-fdata-sections
 
+# Super gross! But I found no way to do -fno-error=permissive
+# this steps from the fact that demangle is a C++ file extension
+# but the source I copied was written in C.
+# The fix would be to go through all the casts and make them correct.
+o/$(MODE)/third_party/libcxx/demangle.o: private				\
```
fzakaria, line 221: Two options here:
1. is there a way to compile this as a C file while still including the other libcxx files?
I guess I could include the libc equivalent and just change this to a .c file?

2. edit the source to add casts to not emit the permissive error anymore.

jart: With something like this, the dependency footprint is so small that the best place for it is probably libc/stdio/. I just plopped the file in there, made sure the license was at the top, and then added a single line to stdio.h.

image

Those `COSMOPOLITAN_C_END_` etc. defines are what automate the `extern "C"` thing, so that C++ code can include C headers. I'm going to cherry pick this change onto the master branch manually in a moment.

PR #851, opened 2023-07-06
Add OnServerReload to replace /.reload.lua with a bit more flexibility

pkulchenko:
I kept `/.reload` in place and only marked it as "deprecated" in the documentation.

The documentation states that it's only called after `SIGHUP` (when running as a daemon), but it should also be called after `SIGUSR1` in all cases. I've reflected that in `OnServerReload` description, but haven't modified the `.reload` text.


jart (approved):
Thank you!

PR #854, opened 2023-07-08
lz4cli: consolidate xxhash to third_party version

fzakaria:
Remove the internal xxhash within lz4cli and consolidate to the one true xxhash within third_party.


third_party/lz4cli/lz4cli.mk
```diff
@@ -42,7 +41,8 @@ THIRD_PARTY_LZ4CLI_DIRECTDEPS =				\
 	LIBC_INTRIN					\
 	LIBC_STDIO					\
 	LIBC_LOG					\
-	LIBC_TIME
+	LIBC_TIME					\
+	THIRD_PARTY_XXHASH
```
fzakaria, line 45: Doesn't seem to be picking it up through
```
o/third_party/gcc/bin/x86_64-linux-musl-ld.bfd: o//third_party/lz4cli/bench.o: in function `BMK_benchMem':
third_party/lz4cli/bench.c:228: undefined reference to `XXH64'
o/third_party/gcc/bin/x86_64-linux-musl-ld.bfd: third_party/lz4cli/bench.c:333: undefined reference to `XXH64'
```

jart: You probably need to fix the `include foo.mk` lines in `Makefile` so they're topologically ordered again. xxhash.mk should come before `lz4cli.mk`.

PR #856, opened 2023-07-09
Add EncodeBase32() to Redbean

pkulchenko:
@jart, this implements EncodeBase32, but I tweaked the function from google-authenticator-libpam to (1) allow the alphabet to be configurable and (2) work with any power of 2. I added a couple of tests, so it seems to do what's needed. I don't need DecodeBase32, so it's not included, but let me know if you think it's worth adding for consistency.

### DIFF
--- a/tool/net/lfuncs.c
+++ b/tool/net/lfuncs.c
@@ -605,19 +605,28 @@ int LuaEncodeLatin1(lua_State *L) {
   }
 }
 
-int LuaEncodeBase32(lua_State *L) {
+dontinline int LuaBase32Impl(lua_State *L,
+    char *B32(const char *, size_t, const char *, size_t, size_t *)) {
   char *p;
   size_t sl, al; // source/output and alphabet lengths
   const char *s = luaL_checklstring(L, 1, &sl);
   // use an empty string, as EncodeBase32 provides a default value
   const char *a = luaL_optlstring(L, 2, "", &al);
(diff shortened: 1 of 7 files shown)


tool/net/lfuncs.c
```diff
@@ -602,6 +602,64 @@ int LuaEncodeLatin1(lua_State *L) {
...
 }
 
+int tobits(int b) {
+  int bits = 0; while (b && (b >>= 1)) bits++;
+  return bits;
+}
+
+// the next function is based on
+// https://github.com/google/google-authenticator-libpam/blob/master/src/base32.c
+// Copyright 2010 Google Inc.; Author: Markus Gutschke
```
jart, line 612: Could you put it in its own file, with its own license header?

jart (changes requested):
Sure having the decoder would be cool.

tool/net/lfuncs.c
pkulchenko: @jart, I reorganized all the code and added decoding (along with the documentation).

pkulchenko:
@jart, I followed the same style/convention as is used in Base64 encoding/decoding with one exception: the decoding with the default alphabet accepts/skips whitespaces and dash, but stops at the first character outside of the alphabet (returning the produced result). I can change it to continue (by updating the default map), but I like it a bit better.

jart (approved):
Thank you!

PR #858, opened 2023-07-12
Let lz4toasm accept LZ4 files that lack extracted-size field

tkchia:
`lz4toasm` should now more easily accept LZ4 files output by compressor programs that do not support the extracted-size field, such as Stephan Brumme's `smallz4`.

This patch also proposes to add a new `lz4len()` function to the `libc`: it parses an LZ4 compressed block to compute the unpacked content size, without really unpacking the block.


jart (approved):
Looks good to me. Thank you!

PR #863, opened 2023-07-23
simplify cosmocc builds

ahgamut:
- tcgetpgrp(STDIN_FILENO) should be equal to getpgrp() on Windows also, found while reading wget source code which uses this check to decide whether to print to stderr or to a file
- IN6_ADDR_ARE_EQUAL is a comparison macro used when IPV6 is allowed, found while reading CPython3.11 source code
- the changes in signal.h and addition of ucontext.h are because CPython3.11 source code expect sigaltstack to be available
- the sqlite3.mk change is because CPython3.11 requires sqlite3 to be built with -DOMIT_SHARED_CACHE
- unistd.h has getopt.h now, because some libraries like it there


jart (approved):
Thank you!

Issue #862, opened 2023-07-21
`ProgramDirectory` appends to package.path; should prepend instead?


michaellenaghan:
Add this to `.init.lua`:

```lua
Log(kLogVerbose, " *** package.path: " .. package.path)
```

Now run `redbean -D something`:

```
V2023-07-21T15:23:03+000226:/.init.lua:3:redbean-unsecure-2:13611]  *** package.path: /zip/.lua/?.lua;/zip/.lua/?/init.lua
```

The directory wasn't added to `package.path`.

Now add this to `.init.lua`, before the `Log()` line:

```lua
ProgramDirectory("something")
```

The directory still wasn't added.

The problem is in `UpdateLuaPath()`:

https://github.com/jart/cosmopolitan/blob/18536950b3db37da45641bc5805d82ea6056d283/tool/net/redbean.c#L1087-L1100

I'll submit a pull request once I work out how to do a build. In the meantime, there's a workaround of sorts: set `package.path` manually in `.init.lua`, like this:

```lua
package.path = "something/.lua/?.lua;something/.lua/?/init.lua;" .. package.path
```

===

On a related note, `LuaStart()` calls `GetDefaultLuaPath()`:

https://github.com/jart/cosmopolitan/blob/18536950b3db37da45641bc5805d82ea6056d283/tool/net/redbean.c#L5353

`g_lua_path_default` is never actually used; both it and `GetDefaultLuaPath()` can safely be removed.

pkulchenko:
@michaellenaghan, are you sure you're using the latest master branch. This works for me:

```
> redbean.com -D backup -e "print(package.path); unix.exit()"
I2023-07-21T16:05:58.182708:tool/net/redbean.c:1111:redbean:12636] (cfg) program directory: backup
/zip/.lua/?.lua;/zip/.lua/?/init.lua;backup/.lua/?.lua;backup/.lua/?/init.lua
```

It also produces exactly the same output with `redbean.com -e "ProgramDirectory('backup'); print(package.path); unix.exit()"` (assuming `backup` directory exists), so it all appears to work as designed. What version are you using?

> g_lua_path_default is never actually used; both it and GetDefaultLuaPath() can safely be removed.

It's used in lua/loadlib.c to set the initial `package.path` value.

michaellenaghan:
Ah.

[...]

PR #864, opened 2023-07-24
Update ProgramDirectory to prepend Lua path instead of appending (#862).

pkulchenko:
This update checks the default path in the current path and prepends the added path in front of it. If the path is not found, then it's appended at the end (as the current implementation does). In either case it preserves the order of the additions, but allows to prepend, thus making `require` to check in added paths first, before checking the internal archive.

### DIFF
--- a/tool/net/redbean.c
+++ b/tool/net/redbean.c
@@ -4584,6 +4584,12 @@ static int LuaProgramBrand(lua_State *L) {
 }
 
 static int LuaProgramDirectory(lua_State *L) {
+  struct stat st;
+  char *path = luaL_checkstring(L, 1);
+  // check to raise a Lua error, to allow it to be handled
+  if (stat(path, &st) == -1 || !S_ISDIR(st.st_mode)) {
+    return luaL_argerror(L, 1, "not a directory");
+  }
   return LuaProgramString(L, ProgramDirectory);
 }
 
--- a/tool/net/redbean.c
+++ b/tool/net/redbean.c
@@ -5355,22 +5355,11 @@ static void LuaSetConstant(lua_State *L, const char *s, long x) {
   lua_setglobal(L, s);
 }
 
-static char *GetDefaultLuaPath(void) {
-  char *s;
-  size_t i;
-  for (s = 0, i = 0; i < stagedirs.n; ++i) {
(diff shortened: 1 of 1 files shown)


michaellenaghan:
This is, strictly speaking, independent of this change, but might be worth considering.

* Remove the iteration through `stagedir`s in [`GetDefaultLuaPath()`](https://github.com/jart/cosmopolitan/blob/e0c2b91b3ec6f88d060d6059210a993a0ef46b08/tool/net/redbean.c#L5338); or
* Remove `GetDefaultLuaPath()` and replace the [one place it's called](https://github.com/jart/cosmopolitan/blob/e0c2b91b3ec6f88d060d6059210a993a0ef46b08/tool/net/redbean.c#L5353) with the new `DEFAULTLUAPATH` constant

In the current codebase, nothing can add `stagedir`s before `LuaStart()` is called:

* [`GetDefaultLuaPath()`](https://github.com/jart/cosmopolitan/blob/e0c2b91b3ec6f88d060d6059210a993a0ef46b08/tool/net/redbean.c#L5338) is only called by [`LuaStart()`](https://github.com/jart/cosmopolitan/blob/e0c2b91b3ec6f88d060d6059210a993a0ef46b08/tool/net/redbean.c#L5349)
* `LuaStart()` is called [before `GetOpts()`](https://github.com/jart/cosmopolitan/blob/e0c2b91b3ec6f88d060d6059210a993a0ef46b08/tool/net/redbean.c#L7421)
* `GetOpts()` is called [before `LuaInit()`](https://github.com/jart/cosmopolitan/blob/e0c2b91b3ec6f88d060d6059210a993a0ef46b08/tool/net/redbean.c#L7427)

`GetOpts()` is the first thing that can add `stagedir`s.

pkulchenko:
I'm okay with the proposed changes and can implement them tomorrow, but I'd also change the behavior of `ProgramDirectory`, as it's not a Lua error and hence doesn't allow catching/ignoring the error. I propose to keep the error as is (so it's a fatal one if used from the command line), but to trigger an (earlier) Lua error from `LuaProgramDirectory` to allow catching and ignoring it if needed. @jart, does that work for you?

pkulchenko:
Updated with the changes discussed above.

Issue #861, opened 2023-07-19
tool/redbean: time machinery uses a different timezone


mataha:
When e.g. logging, `redbean` doesn't use the system's timezone, preferring a different one instead:

```batchfile prompt
C:\Users\user> wmic os get LocalDateTime /format:list
LocalDateTime=20230720012811.734000+120

C:\Users\user> redbean
I2023-07-19T16:28:13.079556:tool/net/redbean.c:7057:redbean:22652] (srvr) listen http://127.0.0.1:8080
>:
I2023-07-19T16:28:14.525606:tool/net/redbean.c:7082:redbean:22652] (srvr) received SIGINT
I2023-07-19T16:28:14+002001:tool/net/redbean.c:7089:redbean:22652] (srvr) shutdown complete
```

UTC-08:00 looks like Lower California - does it use a compile-time timezone?

Diagnostics:

 - `redbean` 2.2 (from [here](https://redbean.dev/#download))
 - Windows 10.0.19045.3208
 - system TZ: UTC+01:00

Possibly related: #816

pkulchenko:
It does match for me, but I'm in the same timezone as CA (I'm in WA), so it's not a surprise:

```
>wmic os get LocalDateTime /format:list
LocalDateTime=20230719165305.129000-420

>redbean.com
I2023-07-19T16:53:07.613616:/zip/.lua/fullmoon.lua:60:redbean:12108] (fm) started redbean/2.2.0 fullmoon/0.376
I2023-07-19T16:53:07+000000:tool/net/redbean.c:7090:redbean:12108] (srvr) listen http://192.168.1.12:8080
I2023-07-19T16:53:07+000000:tool/net/redbean.c:7090:redbean:12108] (srvr) listen http://127.0.0.1:8080
I2023-07-19T16:53:08.676212:tool/net/redbean.c:7115:redbean:12108] (srvr) received SIGINT
I2023-07-19T16:53:08+000000:tool/net/redbean.c:7122:redbean:12108] (srvr) shutdown complete
```

I checked the logging and it's supposed to use the [local time](https://github.com/jart/cosmopolitan/blob/18536950b3db37da45641bc5805d82ea6056d283/libc/log/vflogf.c#L115):

```
  localtime_r(&t2.tv_sec, &tm);
  strcpy(iso8601(buf32, &tm), sign);
```

michaellenaghan:
I had to do this to get the correct local time on my Mac:

    TZ=New_York ./redbean.com

[...]

PR #867, opened 2023-07-24
Change the default `TZ` value from `GST` to `GMT` #861

michaellenaghan:
With this change, `unix.localtime()` will match `unix.gmtime()` unless `TZ` has been (successfully) defined.

First, that seems like a less surprising "not defined" default; see #861.

Second, it matches `localtime()`'s current behaviour when `TZ` has been defined but can't be found.

If that doesn't fly, it might at least be worth documenting the fact that the default is `GST`.***

There are two related changes that might also make sense: 

* Change [`TZDEFULES`](https://github.com/jart/cosmopolitan/blob/e0c2b91b3ec6f88d060d6059210a993a0ef46b08/libc/time/tzfile.internal.h#L32) to `GMT`
* Change [`TZDEFRULESTRING`](https://github.com/jart/cosmopolitan/blob/e0c2b91b3ec6f88d060d6059210a993a0ef46b08/libc/time/localtime.c#L122) to, effectively, `GMT` (i.e., no DST)

The thinking is the same: that seems like a less surprising "not defined" default.

If you're good with that I could add those changes to this pull request?

===

***I saw a comment somewhere saying that "GST" is "Google Standard Time." But it looks like there's a different (and official) "GST" timezone: [Gulf Standard Time](https://www.timeanddate.com/time/zones/gst). When I google for "google standard time gst", "Gulf Standard Time" is the only thing that shows up.

### DIFF
--- a/libc/time/tzfile.internal.h
+++ b/libc/time/tzfile.internal.h
@@ -25,7 +25,7 @@
 #endif /* !defined TZDIR */
 
 #ifndef TZDEFAULT
-#define TZDEFAULT	"GST"
+#define TZDEFAULT	"GMT"
 #endif /* !defined TZDEFAULT */
 
 #ifndef TZDEFRULES


jart (approved):
Sounds great. Let's change it. 

I just need one process hurdle from you first before I can merge. Can you please send an email to jtunney@gmail.com saying "I intend to assign you the copyright to the changes I contribute to Cosmopolitan"? It only applies to the changes you choose to contribute. It only needs to happen once. If your employer owns everything you do, then you'll need to clear it with them first. Please do that if you need to and get it on the record. Thanks!

michaellenaghan:
Is that a flakey test? There's no obvious connection to `TZ`.

michaellenaghan:
Should I add changes to `TZDEFULES` and `TZDEFRULESTRING`? Or maybe put them in a separate pull request?

jart:
I honestly don't care much about using epoll from Python 3.6 so I'd say we should simple comment out that test for now. This looks good enough for me to merge. I'll take care of mopping up optlinux later today.

PR #871, opened 2023-07-26
Don’t lowercase the test filename.

michaellenaghan:
This test fails if Cosmo builds from a path that contains an uppercase character.

Paths with uppercase characters aren’t so common in server Linux. But they are in *desktop* Linux. Guess how I…?

But I digress.

The real problem is that the path is lowercased on one line, but not the next:

```
        self.file_name = support.TESTFN.lower()
        self.file_path = FakePath(support.TESTFN)
```

Given that no other test in the suite lowercases `support.TESTFN`, I opted to remove it from the first line rather than add it to the second.


jart (approved):
Thanks!

PR #872, opened 2023-07-27
Re-sync to Lua 5.4.3.

michaellenaghan:
This prepares the ground for bringing Lua up to 5.4.6. It adds some missing copyrights, removes some overlooked copyrights, makes vertical whitespace consistent across files and adds `[jart]` comments in various places to make subsequent Lua comparisons easier. Well. Hopefully.

In theory, nothing here is controversial.

If we merge this, the 5.4.4 update will be “all signal, no noise.” But the noise level isn’t all that high, so it probably doesn’t matter all that much.

### DIFF
--- a/third_party/lua/lapi.c
+++ b/third_party/lua/lapi.c
@@ -27,6 +27,7 @@
 ╚─────────────────────────────────────────────────────────────────────────────*/
 #define lapi_c
 #define LUA_CORE
+
 #include "third_party/lua/lapi.h"
 #include "third_party/lua/ldebug.h"
 #include "third_party/lua/ldo.h"
@@ -42,6 +43,7 @@
 #include "third_party/lua/lua.h"
 #include "third_party/lua/lundump.h"
 #include "third_party/lua/lvm.h"
+
 // clang-format off
 
 asm(".ident\t\"\\n\\n\
@@ -1524,7 +1526,7 @@ LUA_API int lua_error (lua_State *L) {
   else
     luaG_errormsg(L);  /* raise a regular error */
   /* code unreachable; will unlock when control actually leaves the kernel */
-  __builtin_unreachable();
+  __builtin_unreachable();  // [jart]
 }
 
 
--- a/third_party/lua/lapi.h
+++ b/third_party/lua/lapi.h
@@ -4,7 +4,7 @@
 #include "third_party/lua/llimits.h"
 #include "third_party/lua/lstate.h"
 
-/* clang-format off */
+// clang-format off
 
 /* Increments 'L->top', checking for stack overflows */
 #define api_incr_top(L)   {L->top++; api_check(L, L->top <= L->ci->top, \
--- a/third_party/lua/lauxlib.c
+++ b/third_party/lua/lauxlib.c
@@ -27,6 +27,7 @@
 ╚─────────────────────────────────────────────────────────────────────────────*/
 #define lauxlib_c
 #define LUA_LIB
+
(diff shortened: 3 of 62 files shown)


pkulchenko:
> If we merge this, the 5.4.4 update will be “all signal, no noise.” But the noise level isn’t all that high, so it probably doesn’t matter all that much.

@michaellenaghan, looks good to me, but we actually don't need to merge this yet, as you can keep adding 5.4.4, 5.4.5, and 5.4.6 as separate commits. We'll then test and merge the branch.

I did the same thing you did with .inc files in the SQLite upgrade, as renaming them was difficult to manage with the upgrades (so I had to rename them back).

pkulchenko:
@michaellenaghan, it's better to rebase on top of the current master instead of merging master, as it will allow you to keep the branch history clean.

michaellenaghan:
@pkulchenko Roger that.

michaellenaghan:
I didn't realize that building Lua didn't run Lua's tests.

Five test files are ISO-8859-1 encoded. At the beginning of this process I converted them to UTF-8. The problem is that two of the test files do some byte counting, and of course changing the encodings changes the byte counts. Given that, I reverted all five test files to ISO-8859-1 encoding.

Pity.

All Lua tests now pass in all branches.

pkulchenko:
Thank you for these updates! FWIW, I'd keep all the commits in one PR (like you have in your last one) as it's easier to read/process that way. If someone needs to check on a subset, they can check out a particular version.

Just to confirm: these changes upgrade Lua to 5.4.6 and othercosmo-local changes that are covered in the README still need to be added, right?
```
  Lua now uses a bestline REPL with bash-style code completion.
  Integer literals such as `033` will now be interpreted as octal.
  Integer literals such as `0b10` will now be interpreted as binary.
  The `\e` string literal escape sequence has been added, which is
  equivalent to `\27` (the Lua version of `\033`) or the ASCII ESC
  character. It may be used for teletypewriter control like having
  bold text, which can be encoded elegantly as `\e[1mHELLO\e[0m`.
  Added luaL_traceback2() for function parameters in traceback.
  Added Python-like printf modulus operator for strings.
  Added Python-like printf multiply operator for strings.
```

michaellenaghan:
> FWIW, I'd keep all the commits in one PR (like you have in your last one) as it's easier to read/process that way. If someone needs to check on a subset, they can check out a particular version.

It depends on how you merge, no? Many projects squash everything into one commit. If you do that you'll have 5.4.6, but you'll loose the intervening steps. Either way, your call.

> Just to confirm: these changes upgrade Lua to 5.4.6 and othercosmo-local changes that are covered in the README still need to be added, right?

As far as I know, everything is there and working. :-)

```sh
# o//third_party/lua/lua.com
Lua 5.4.6  Copyright (C) 1994-2023 Lua.org, PUC-Rio
>: print(0744)
484
>: print(0b010)
2
>: print("\e[1mHELLO\e[0m")
HELLO
>: print("Hello, %s" % {"world"})
Hello, world
>: print("Hello, %s" % {"world"} * 2)
Hello, worldHello, world
```

(I'm not sure how to trigger `luaL_traceback2()`. But the code is all there.)

pkulchenko:
> Many projects squash everything into one commit. If you do that you'll have 5.4.6, but you'll lose the intervening steps. Either way, your call.

Yes, that's how @jart usually does merges.

> As far as I know, everything is there and working. :-)

That's great; thank you!

michaellenaghan:
Resolved in #1214.

PR #878, opened 2023-08-01
fix x-axis blurring in nesemu1 tty conversion

ahgamut:
before this change:
before

after this change:
after

Screenshots taken while rendering on `kitty` terminal, with the lowest allowed font size.
Aspect ratio is off because I am on fullscreen, here's a square-shaped screenshot after applying this change:

after2

I am not sure if `2` is the correct value, but it is definitely better than zero, because the value ends up being used in a division.


jart (approved):
Great find!

PR #880, opened 2023-08-04
Paginate by writing data to pager via stdin

mataha:
On Windows, `more` is unable to open the file generated by `mkstemp()` due to not recognizing the first argument as a path, but as a parameter. In contrast to *nix, arguments on Windows are, by convention, prefixed with `/`, which `mkntcmdline()` happily transforms into path-like strings.

This commit sidesteps this issue entirely by writing data to the pager in question via stdin, dumping text to the console as a last resort.

### DIFF
--- a/libc/calls/mkntcmdline.c
+++ b/libc/calls/mkntcmdline.c
@@ -90,26 +90,13 @@ textwindows int mkntcmdline(char16_t cmdline[32767], char *const argv[]) {
         }
       }
       if (!x) break;
-      if (x == '/' || x == '\\') {
-        if (!i) {
-          // turn / into \ for first argv[i]
-          x = '\\';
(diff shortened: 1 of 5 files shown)


mataha:
Recent changes seem to have broken `PATH` discovery - don't merge this.

mrdomino:
@G4Vi is this still a going concern? It looks like this may have been addressed by one of your recent PRs?

G4Vi:
I forgot about this, sorry this fell through. I ended up solving pagination on windows with path translation in #1148.

Closing this as `__paginate` was almost rewritten. @mataha if you still feel strongly about using stdin instead, feel free to remake this. Respecting `$PAGER` could be nice addition if we switch over to using stdin instead of creating temp files.

PR #883, opened 2023-08-11
Linux add definition for madvise flag

michalbiesek:
@jart This PR provides only extend the definition without the logic behind it, please let me know if You are ok with the TODO comment


jart (approved):
LGTM. Thank you!

PR #886, opened 2023-08-17
get rsync to build

ahgamut:
- added some IPV6 comparison macros
- the `MSG_INFO` in `msg.h` clashes with an enum declaration that `rsync` makes


jart (approved):
Thanks!

PR #894, opened 2023-09-12
Do not format `int` as a non-printable character

mataha:
Output gets truncated otherwise.

### DIFF
--- a/libc/testlib/formatint.c
+++ b/libc/testlib/formatint.c
@@ -29,7 +29,7 @@ char *testlib_formatint(intptr_t x) {
   char *str = sbufi_ < ARRAYLEN(sbufs_) ? sbufs_[sbufi_++] : malloc(256);
   char *p = str;
   p += sprintf(p, "%ld\t(or %#lx", x, x);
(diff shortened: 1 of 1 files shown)


G4Vi:
@mataha Thanks for the fix.

Closing as I believe this was addressed in recent changes https://github.com/jart/cosmopolitan/blob/ec480f5aa0c047944cef305e8f33b1caae049436/libc/testlib/formatint.c 

Feel free to rebase and reopen/remake the PR if this is not the case.

mataha:
Can confirm it's been addressed indeed:

```c
error:test/libc/calls/access_test.c:82: access_textFileIsntExecutable() on wsl2buntu pid 11202 tid 11202
        ASSERT_EQ(-1, access("foo.txt", R_OK | X_OK))
                need -1 (or 0xffffffffffffffff) =
                 got 0 (or 0x0 or '\0')
        EUNKNOWN/0/No error information
        o//test/libc/calls/access_test.com @ wsl2buntu
1 / 33 tests failed
```

PR #895, opened 2023-09-13
ignore trailing slash when sorting zipos entries

ahgamut:
before running the binary search in zipos-find.c, the query string is checked for a trailing slash, which if found is ignored during the memcmp as part of the search.

however, '/' (ASCII 47) is after '-' (ASCII 45),  so the sorted array could contain entries like "/zip/x-" followed by "/zip/x/". Now, if the query string is "/zip/x/", the trailing slash will be removed, but then "/zip/x-" will be incorrectly matched, leading to an error.

this change adds a similar check to ignore the trailing slash during the construction of the sorted array to ensure "/zip/x-" occurs after "/zip/x/", and avoid such errors.


ahgamut:
the trailing slash is ignored when performing the comparisons for the binary search, so it should be ignored when performing the comparisons for sorting as well.

ahgamut:
closing without merging because this error appears to be fixed by the changes in https://github.com/jart/cosmopolitan/commit/ececec4c943bc644186091d156d43e1475589bc5

PR #900, opened 2023-09-19
[metal] Fix regression causing early crash in __new_page( )

tkchia:
There was a glitch in the refactoring of `__map_phdrs( )` in commit ec480f5aa0, which caused it to try to map the `PT_NOTE` program segment into virtual memory, which in turn caused the memory page at `BANE` to be wrongly remapped to the start of the program image (`0x100000`) rather than physical address 0. This affected subsequent page allocation operations because the `struct mman` was located at `BANE + 0x0500`.


jart (approved):
Thank you!

PR #901, opened 2023-09-22
Fix strtod NaN handling / fix SIGSEGV in testlib/showerror

G4Vi:
We could change the `NAN_WORD0`/ `NAN_WORD1` macros instead. I chose to hardcode the constants to match the style of the `infinity` and other `nan` case.

On error in `strtod_test.c` `tinyprint` crashed without a `NULL` arg at the end being passed.


jart (approved):
Thank you!

PR #905, opened 2023-10-07
[metal] Add a uprintf() routine, for non-emergency boot logging

tkchia:


libc/intrin/kprintf.h
```diff
@@ -9,25 +9,49 @@
...
 bool kisdangerous(const void *);
+
 void klog(const char *, size_t);
+void _klog_serial(const char *, size_t);
 long kloghandle(void);
 
+void uprintf(const char *, ...);
+void uvprintf(const char *, va_list);
+
+#define KINFOF(FMT, ...)                                      \
```
jart, line 34: could you use `#ifndef TINY` for these instead and remove the include

jart (changes requested):

libc/intrin/kprintf.h
jart: It's mostly an issue for header files. I try to keep include complexity down. It also made your macros more elegant.

jart (approved):

PR #909, opened 2023-10-09
Make `args` replace `...` only at the end of file

mataha:
So that the string '...' can be used verbatim elsewhere in the file.

---

I have a question regarding my PRs, including this one: what stops them from being merged? Is it the e-mail from which I've made the commits (which is kept private by GitHub)?

If that's the case, I can include my public mail in all of my commits as well as send a proper disclaimer from it - I just need to know what exactly do I need to do.


jart (approved):
Looks good. Thanks!

PR #912, opened 2023-10-15
add tsearch from musl

ahgamut:
from https://git.musl-libc.org/cgit/musl/snapshot/musl-1.2.4.tar.gz

uses `#include "libc/isystem"` :sweat:


ahgamut:
ok I changed the headers to not `libc/isystem`

jart (approved):
As you wish.

PR #914, opened 2023-10-21
Add database and statement readonly checks to sqlite in redbean

pkulchenko:
Also made a small cleanup of de/serialize calls to only compile when `SQLITE_ENABLE_DESERIALIZE` is set.


jart (approved):
Thanks!

Issue #456, opened 2022-06-28
sscanf doesn't know about floating-point numbers?


parasti:
Basic test case:

```c
int main()
{
    float f = 0;
    sscanf("0.3715", "%f", &f);
    printf("%f\n", f);
    return 0;
}
```

Results in
```
$ ./hello.com.dbg
0.000000
```

matheusmoreira:
I'm working on it. So far I've managed to make it pass your test case:

```shell
$ ./o/examples/parsefloat.com
0.371500
```

Need to test it more and polish it up for a pull request.

PR #924, opened 2023-11-02
Floating point parsing support for scanf family

matheusmoreira:
I have implemented floating point parsing logic in the `vcscanf` function and consequently all the other functions which call it.

There is a small state machine in the code which attempts to recognize the floating point number before passing it to the `strtod` function from the `gdtoa` third party library. I attempted to use `double-conversion` but it's a C++ library.

I'm open to review and any feedback. Support for the maximum length specifier is missing, I'll add it when I figure out a better implementation. Meanwhile I thought it was a good idea to contribute this and see what everyone thinks.

Relevant issue: #456

### DIFF
--- a/examples/parsefloat.c
+++ b/examples/parsefloat.c
@@ -1,0 +1,31 @@
+#include <stdio.h>
+
+#define PARSE_AND_PRINT(type, scan_fmt, print_fmt, str)       \
+  do {                                                        \
+    type val; int ret;                                        \
+    ret = sscanf(str, scan_fmt, &val);                        \
+    printf("\"%s\" => " print_fmt " = %d\n", str, val, ret);  \
+  } while (0)
+
+int main()
+{
+    PARSE_AND_PRINT(float, "%f", "%f", "0.3715");
+    PARSE_AND_PRINT(float, "%f", "%f", ".3715");
+    PARSE_AND_PRINT(float, "%f", "%f", "3715");
+    PARSE_AND_PRINT(float, "%f", "%f", "111.11");
+    PARSE_AND_PRINT(float, "%f", "%f", "-2.22");
+    PARSE_AND_PRINT(float, "%f", "%f", "Nan");
+    PARSE_AND_PRINT(float, "%f", "%f", "nAn(2)");
+    PARSE_AND_PRINT(float, "%f", "%f", "-NAN(_asdfZXCV1234_)");
+    PARSE_AND_PRINT(float, "%f", "%f", "-nan");
+    PARSE_AND_PRINT(float, "%f", "%f", "+nan");
+    PARSE_AND_PRINT(float, "%f", "%f", "inF");
+    PARSE_AND_PRINT(float, "%f", "%f", "iNfINiTy");
+    PARSE_AND_PRINT(float, "%f", "%f", "+inf");
+    PARSE_AND_PRINT(float, "%f", "%f", "-inf");
+    PARSE_AND_PRINT(float, "%f", "%f", "0X1.BC70A3D70A3D7P+6");
+    PARSE_AND_PRINT(float, "%f", "%f", "1.18973e+4932zzz");
+    PARSE_AND_PRINT(float, "%f", "%.10f", "   -0.0000000123junk");
+    PARSE_AND_PRINT(float, "%f", "%f", "junk");
+    return 0;
+}
--- a/libc/stdio/vcscanf.c
+++ b/libc/stdio/vcscanf.c
@@ -35,17 +35,17 @@
     c;                       \
   })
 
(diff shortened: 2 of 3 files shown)


libc/stdio/vcscanf.c
```diff
@@ -34,6 +35,21 @@
     c;                       \
   })
 
+#define BUFFER_GROW 48
+#define BUFFER                                         \
+  ({                                                   \
+    int c = READ;                                      \
+    if (bufcur >= bufsize - 1) {                       \
+      bufsize = bufsize + BUFFER_GROW;                 \
+      buf = realloc(buf, bufsize);                     \
```
jart, line 44: Do we need dynamic memory if we're calling strtod()? Would it be possible to simply save a pointer to first character of the number, and then pass that to strtod()?

libc/stdio/vcscanf.c
```diff
@@ -293,6 +332,150 @@ int __vcscanf(int callback(void *),    //
...
+	      case FRACTIONAL:
+		if (isdecdigit || (hexadecimal && ishexdigit)) {
+		  goto Continue;
+		} else if (state == INTEGER && ispoint) {
+		  state = FRACTIONAL;
+		  goto Continue;
+		} else if (isdecexp || (hexadecimal && ishexp)) {
+		  state = SIGN;
+		  goto Continue;
+		} else goto Break;
```
jart, line 441: Add curlies for this please.

examples/parsefloat.c
```diff
@@ -0,0 +1,31 @@
...
+#define PARSE_AND_PRINT(type, scan_fmt, print_fmt, str)       \
+  do {                                                        \
+    type val; int ret;                                        \
+    ret = sscanf(str, scan_fmt, &val);                        \
+    printf("\"%s\" => " print_fmt " = %d\n", str, val, ret);  \
+  } while (0)
+
+int main()
+{
+    PARSE_AND_PRINT(float, "%f", "%f", "0.3715");
```
jart, line 12: Please add test cases to test/libc/stdio/sscanf_test.c

One way you can test floating point numbers is to do what the test/libc/tinymath/ tests do, and round trip it back to a string.

```c
  EXPECT_STREQ(".962423650119207", gc(xdtoa(_acosh(1.5))));
```

jart (changes requested):
Thanks for sending this.

libc/stdio/vcscanf.c
matheusmoreira: Probably! I'd really prefer to avoid memory allocation if at all possible.

I chose this approach because of the callback interface which provides one character at a time. Is there a way to get a pointer to the string being scanned?

examples/parsefloat.c
matheusmoreira: Will do. I realized there was a test suite minutes after I sent the PR.

libc/stdio/vcscanf.c
jart: I'm sorry I take this back. I forgot that this function has a character reader function. Sorry!

matheusmoreira: No problem. I'm wondering if there's a better way to do this too. The parsing logic got somewhat complex due to the single character nature of the function.

matheusmoreira:
There seems to be some bug lurking about. Only shows up if I parse several numbers from the same string one after the other. Hunting it down now.

By the way, please let me know if you'd like me to clean up the git history later!

examples/parsefloat.c
matheusmoreira: Added the tests!

matheusmoreira:
Alright! All tests pass.

jart (approved):
Looks good to me. Thank you so much for doing this!

matheusmoreira:
It appears the test segfaulted during the continuous integration build process.

```
make MODE=tinylinux -j2 o/tinylinux/test/libc/stdio/sscanf_test.com.runs` terminated by SIGSEGV
```

Any indication as to why? Tests pass on my machine but I did not test the `tinylinux` mode.

jart (changes requested):
After checking the change out locally and taking a closer look, could you please refactor all usages of the `buf` variable? `free(buf)` is crashing later in the program. I think the issue is that `buf` is supposed to be used for the growing array, but it's also being assigned in two other places for seemingly unrelated purposes.

matheusmoreira:
Indeed, I was only able to reproduce the segfault when building in `tinylinux` mode. Refactored as requested. Created separate buffers for the floating point parser and string decoder. Now tests pass in both build modes.

jart (approved):
Looks good. You're making a lot of people happy with this change, which is something we've really wanted for a while.

PR #941, opened 2023-11-10
Close Lua state before exiting redbean worker to allow GC to collect

pkulchenko:
@jart, this is a small patch to destroy the Lua state and allow Lua GC handlers to be called before workers exit. The only other way to get close to this is to call `os.exit(true, true)` from `OnWorkerStop`, but obviously it's not ideal, as it exits the process a bit too early not allowing redbean to finish its work.


jart (approved):
Thanks!

PR #942, opened 2023-11-10
ape: Fix uname -m redirection

tavianator:
This was leading to errors like

    .../tool/build/fixupobj.com: this ape binary lacks  support


jart:
Thanks for the report. I'm going to fix this locally so a fix goes out sooner than the copyright process would take. What I'd like to see happen is some kind of contribution to our GitHub Actions config that'll let us test the APE shell script for environments that don't have `binfmt_misc`. Right now it's hard for us to test the shell script because our build config is hermetic and won't run files from the system.

PR #948, opened 2023-11-12
fixes for building superconfigure

ahgamut:


ahgamut:
apart from this PR, some other small changes that need to be added:

- `tool/cosmocc/package.sh` needs to copy `libc/integral/*.inc` to the right spot
- `EXIT_SUCCESS` and `EXIT_FAILURE` need to be in some other public header that the `llvm` build is looking for

libc/isystem/future
```diff
@@ -0,0 +1,4 @@
+#ifndef COSMOPOLITAN_LIBC_ISYSTEM_FUTURE_
+#define COSMOPOLITAN_LIBC_ISYSTEM_FUTURE_
+#include "third_party/libcxx/future"
```
jart, line 3: Can you add this to `LIBC_ISYSTEM` in `libc/libc.mk` please?

jart (changes requested):

jart (approved):

PR #949, opened 2023-11-12
Fix session changeset procesing (reverting some of 0d748ad5 changes)

pkulchenko:
This removes strdup from Lua SQLite wrapper, as it's applied to binary changeset strings, thus cutting them short (because of strlen). It also doesn't need strdup in that context, as lua strings should be anchored as long as their values are in the stack (and in this case they should be until the function completes).


jart (approved):
Thanks!

PR #955, opened 2023-11-15
[metal] Make deathstar.com draw onto both VGA and serial consoles

tkchia:
In addition, `deathstar.com` can now be booted up via UEFI.


jart (approved):
Thanks!

PR #956, opened 2023-11-15
Make test/libc/calls/pledge_test.c work with "nonstandard" umask

tkchia:
The test was failing if the process's umask happened to be `0077`, for example.  The file `foo` was then created with a file mode of `0100600`, rather than the expected `0100644`.


jart (approved):
Thanks!

PR #957, opened 2023-11-16
Fix regression when building deathstar.com for aarch64

tkchia:
It turns out my earlier commit ddc08dc974 caused a build with `MODE=aarch64` to fail.  The commit changed `deathstar.c` to link in code to support a VGA console, but this is not implemented yet for AArch64.  Thanks to @ahgamut for spotting this issue.


jart (approved):
Thanks!

Issue #439, opened 2022-06-22
[FR] Expect redbean to support websocket


lost4git:

jart:
I agree. Redbean should support websockets. It's something we don't have planned immediately but we certainly intend to do it in the not too distant future.

javajosh:
I am also quite interested in websocket support. Any leads on this in terms of libraries? Or do you intent to implement something in the `redbean.c` file?

jart:
We intend to implement it since the mechanics of websockets are pretty deeply intertwined with the workings of a web server. If there's any off the shelf helpers we can use for some key aspects of it, then we'd use them. But overall the process is just drilling through the RFC and writing tests.

javajosh:
Today I've forked and built `jart/cosmopolitan`, started reading through `redbean.c` and also `rfc6455.txt` (each of which, interestingly enough, are about 7k lines). I have little experience with c, but I find the code quite readable. Justine, can you recommend any resources that might proactively ease my pain with c and its tooling, especially debugging with gdb? TIA

jart:
Join our Discord #help channel and I'll mentor you in the art of C. Like you I have a background in Java (particularly Guava) so I could draw upon some of that experience to help clarify anything you need. https://discord.gg/FunsGsDG

jonnybrooks:
I'd be very intertested in contributing to the implementation of WebSockets in redbean! I wrote a very simple (non spec-compliant) WebSocket server in C a couple years back and since discovering cosmo was wondering if there was something I could contribute to!

JakeBamrah:
Forgive me if this i a naïve question here. I've been looking into this (along with [RFC6455](https://datatracker.ietf.org/doc/html/rfc6455)) for the last few days. I am a beginner C programmer but would love to give this a shot (although I think it will take a few months). Could this be broken down into two sub-task:
1. Implementing websocket api + tests
2.  _Then_ integrating it into `redbean` and `greenbean` 

or does / should these two tasks be worked on in tandem?

pkulchenko:
@JakeBamrah, there is an implementation being worked on in #967, so you may want to review and discuss it there.

PR #967, opened 2023-11-27
Add websocket support to redbean

wingdeans:
I would like websocket support in redbean #439.
I think it can be done most easily with the following Lua API, which builds on the coroutine support implemented for streaming:

```lua
function OnHttpRequest()
    if GetPath() == '/ws' then
        UpgradeWS() -- Set status/headers for handshake
        -- Additional headers can be set after UpgradeWS and before yield
        coroutine.yield() -- Conduct handshake

        local fds = {[GetClientFd()] = unix.POLLIN}
        while true do
            unix.poll(fds)
            print(ReadWS()) -- Do stuff with packet

            Write("Received data") -- Write data
            coroutine.yield() -- Flush
        end
    end
end
```

This PR implements the handshake and framing, so that the above example functions.

TODO:
- [x] Fragmentation
- [x] Ping
- [x] Text frame
- [x] Tests (via [autobahn-testsuite](https://github.com/crossbario/autobahn-testsuite/))
- [ ] Redbean tests
- [ ] `SetHeader('Upgrade': 'websocket')`

Autobahn Test Results:
```json
["FAILED",1]
["INFORMATIONAL",3]
["NON-STRICT",4]
["OK",293]
["UNIMPLEMENTED",216]
```

Please feel free to give criticism of the API or code; there are definitely improvements to be made.

### DIFF
--- a/tool/net/redbean.c
+++ b/tool/net/redbean.c
@@ -5058,13 +5058,15 @@ static bool LuaRunAsset(const char *path, bool mandatory) {
   return !!a;
 }
 
-static int LuaUpgradeWS(lua_State *L) {
+static int LuaWSUpgrade(lua_State *L) {
   size_t i;
   char *p, *q;
   bool haskey;
   mbedtls_sha1_context ctx;
   unsigned char hash[20];
-  OnlyCallDuringRequest(L, "UpgradeWS");
+
+  if (cpm.generator)
+    luaL_error(L, "Cannot upgrade to websocket after yielding normally");
 
   if (!HasHeader(kHttpWebsocketKey))
     luaL_error(L, "No Sec-WebSocket-Key header");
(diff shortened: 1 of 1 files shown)


pkulchenko:
@wingdeans, thank you for the updates! Do you have any feedback on my earlier comments?

coderofsalvation:
exciting!

```
local fds = {[GetClientFd()] = unix.POLLIN}
```
 this is setting a dynamic key while declaring a table? no idea lua could do that ❤️

pkulchenko:
> upgrading the connection on Upgrade: websocket could work.

I like that! I agree that it should work and can't think of any reason why someone would *not* want that.

We may need to be careful though, as another `SetStatus` call can reset the headers, so this may need to be taken into account in `SetStatus` to set `iswebsocket` back to `false`.

pkulchenko:
I like the API (including moving it into its own module). I was thinking about adding `ws.Write` (for consistency with `ws.Read`), but maybe we can leverage it and trigger `SetHeader('Upgrade', 'websocket')` from `ws.Write` if `cpm.wstype` is not set, which can also check for `Sec-WebSocket-Key`? This way the user only needs to be mostly concerned with `ws.Read` and `ws.Write` (and `coroutine.yield()`).

pkulchenko:
Nice! Would it make sense to imply `TEXT` by default for `ws.Write`?

Also, what's this value https://github.com/jart/cosmopolitan/pull/967/files#diff-a937cc5818818785e079eab14ff76fa5db080fe9bd902d1ab21aa546aa79c7d1R5089 and why is it hardcoded?

pkulchenko:
> Right now it defaults to not changing the mode.

Got it; it changes the current output and all subsequent ones.

pkulchenko:
@wingdeans, `bool haskey` can be removed, as it's no longer used. I can add the docs; do you plan to add a couple of tests?

stet:
Is this the only work on websockets for redbean?

PR #970, opened 2023-11-29
Get tool/cosmocc/package.sh working on M1

mrdomino:
On aarch64 hosts, MODE= is changed to MODE=aarch64, so o// targets don't real. So On aarch64, get apelink.com out of o/aarch64/. Also prepend ape when calling it. And finally, fetch with curl when wget isn't installed.


mrdomino:
Done.

PR #972, opened 2023-11-30
Implement __zipos_dup

mrdomino:
Makes ZiposHandle reference-counted by an `rc` field in a union with its
freelist `next` pointer. The functions `__zipos_free` and `__zipos_keep`
function as incref/decref for it. Adds `__zipos_postdup` to fix metadata
on file descriptors after dup-like operations, and adds zipos support to
`sys_dup_nt` + `sys_close_nt`.

### DIFF
--- a/libc/runtime/zipos.internal.h
+++ b/libc/runtime/zipos.internal.h
@@ -20,8 +20,9 @@ struct ZiposHandle {
   struct Zipos *zipos;
   size_t size;
   size_t mapsize;
-  size_t pos;
   size_t cfile;
+  _Atomic(int) refs;
+  size_t pos;         // TODO atomic
   uint8_t *mem;
   uint8_t data[];
 };
@@ -38,6 +39,7 @@ struct Zipos {
 
 int __zipos_close(int);
 void __zipos_free(struct ZiposHandle *);
+struct ZiposHandle *__zipos_keep(struct ZiposHandle *);
 struct Zipos *__zipos_get(void) pureconst;
 size_t __zipos_normpath(char *, const char *, size_t);
 ssize_t __zipos_find(struct Zipos *, struct ZiposUri *);
@@ -45,6 +47,7 @@ ssize_t __zipos_scan(struct Zipos *, struct ZiposUri *);
 ssize_t __zipos_parseuri(const char *, struct ZiposUri *);
 uint64_t __zipos_inode(struct Zipos *, int64_t, const void *, size_t);
 int __zipos_open(struct ZiposUri *, int);
+void __zipos_postdup(int, int);
 int __zipos_access(struct ZiposUri *, int);
 int __zipos_stat(struct ZiposUri *, struct stat *);
 int __zipos_fstat(struct ZiposHandle *, struct stat *);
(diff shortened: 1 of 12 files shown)


libc/runtime/zipos.internal.h
```diff
@@ -16,7 +16,10 @@ struct ZiposUri {
 };
 
 struct ZiposHandle {
-  struct ZiposHandle *next;
+  union {
+    struct ZiposHandle *next;
+    ssize_t rc;
```
mrdomino, line 21: @jart should this be atomic?

jart: I'm assuming `rc` is a reference count (note: `rc` usually means "result code" in the cosmopolitan codebase). Cosmo has also assumed until now that any given `fd` has a single owner thread. With your change it appears that changes, at least for the `ZiposHandle` data structure. So I'd say, yes we want this to be atomic. We might even need a mutex too (but hopefully not since mutexes aren't a great thing to have in async-signal-safe functions).

Here's how cosmo currently implements reference counting for PosixThread objects.

```c
__funline void _pthread_ref(struct PosixThread *pt) {
  atomic_fetch_add_explicit(&pt->pt_refs, 1, memory_order_relaxed);
}

static bool _pthread_deref(struct PosixThread *pt) {
  int refs = atomic_load_explicit(&pt->pt_refs, memory_order_acquire);
  return !refs || !atomic_fetch_sub(&pt->pt_refs, 1);
}

void _pthread_unref(struct PosixThread *pt) {
  if (_pthread_deref(pt)) {
    unassert(_weaken(_pthread_free));
    _weaken(_pthread_free)(pt, false);
    _weaken(_pthread_decimate)();
  }
}
```

jart (commented):
Looking good so far. Thanks for the early glance.

libc/runtime/zipos.internal.h
mrdomino: Hmm. At least zipos is read-only, so IIUC the only thing that could be contended for besides the refcount is `pos`.

My inclination would be to see if we can get away with making just the two of them atomic. I'm foggy on atomics though, so not sure what memory order makes sense for `pos`.

mrdomino: Renamed `rc` to `refs` and made it atomic. Added a TODO for `pos`.

mrdomino: I've kept the quirks of `__zipos_keep` / `__zipos_free` for now — different naming convention and using 0 for initial refcount. Can change if desired, or change the pthread implementation (I think 0 for initial ref is slightly nicer.)

jart (approved):
Looks good to me. This is an impressive change. I cloned your branch locally and ran it on my test fleet of VMs (freebsd rhel7 xnu win10 openbsd netbsd pi silicon) using runit and runitd. I was amazed to see it all green.

I've contacted you about the copyright thing. Once that happens, I'll merge this.

mrdomino:
Sadly the commit message fell out of sync with the actual changes, and I had a couple more changes in flight (implementing atomic reads and seeks even though POSIX says they don't exist because why not). What's the best approach - put the new changes on a new branch or force-push to main?

jart:
It's probably too late to force push at this point. I'll make a mental note of it for the release notes.

PR #973, opened 2023-12-01
Better zipos refcounts and atomic reads/seeks

mrdomino:


libc/runtime/zipos-open.c
```diff
@@ -79,13 +83,19 @@ static void *__zipos_mmap_space(size_t mapsize) {
 }
 
 struct ZiposHandle *__zipos_keep(struct ZiposHandle *h) {
-  atomic_fetch_add_explicit(&h->refs, 1, memory_order_relaxed);
+  int refs = atomic_fetch_add_explicit(&h->refs, 1, memory_order_relaxed);
+  if (VERY_UNLIKELY(refs > MAX_REFS)) {
```
jart, line 87: I recommend `unassert(!VERY_UNLIKELY(refs > MAX_REFS));` for core library stuff like this, since it uses `__builtin_trap()` which is nicer and lighter weight than `abort()` which leads to hairier backtraces. There's also `npassert()` which guarantees the branch is never DCE'd in `NDEBUG` mode.

libc/runtime/zipos-open.c
```diff
@@ -79,13 +83,19 @@ static void *__zipos_mmap_space(size_t mapsize) {
...
+    abort();
+  }
   return h;
 }
 
 static bool __zipos_drop(struct ZiposHandle *h) {
-  int refs = atomic_load_explicit(&h->refs, memory_order_acquire);
-  return -1 == refs || -1 == atomic_fetch_sub(&h->refs, 1);
+  if (!atomic_fetch_sub_explicit(&h->refs, 1, memory_order_release)) {
```
jart, line 94: Interesting unref design. For posterity, I'd be interested in hearing where this is written.

jart (changes requested):
Beautiful, I'm loving it.

libc/runtime/zipos-open.c
mrdomino: Sources cited in the commit message.

jart (approved):
LGTM

PR #977, opened 2023-12-03
$prog.ape support

mrdomino:


ape/ape-m1.c
```diff
@@ -964,7 +988,20 @@ int main(int argc, char **argv, char **envp) {
   sp = sp2;
 
   /* interpret command line arguments */
-  if ((M->ps.literally = argc >= 3 && !StrCmp(argv[1], "-"))) {
+  if ((M->ps.indirect = GetIndirectOffset(argv[0]))) {
+    M->ps.literally = 0;
+    /* if argv[0] is $prog.ape, then we strip off the .ape and run
```
jart, line 993: Could you explain the use case a bit more?

> [Existing comment]. This lets you use symlinks to trick the OS at times when a native executable is required. For example, let's say you want to put the APE binary `/opt/cosmos/bin/bash` in your `/etc/shells`, or perhaps you want to use it as a shebang interpreter (i.e. `#!/opt/cosmos/bin/bash`). That won't work with APE normally, but it will if you say `ln -sf /usr/local/bin/ape /opt/cosmos/bin/bash.ape` and then use `#!/opt/cosmos/bin/bash.ape` instead.

Speaking of which, we might want to do this for `ape/loader.c` too.

ape/ape-m1.c
```diff
@@ -964,7 +988,20 @@ int main(int argc, char **argv, char **envp) {
...
+       $prog. */
+    prog = (char *)sp[1];
+    argc = sp[0];
+    argv = (char **)(sp + 1);
+    /* when we are started as a login shell, a '-' is prepended to
```
jart, line 998: Could you explain this to me?

ape/ape-m1.c
```diff
@@ -975,6 +1012,7 @@ int main(int argc, char **argv, char **envp) {
   } else if (argc < 2) {
     Emit("usage: ape   PROG [ARGV1,ARGV2,...]\n"
          "       ape - PROG [ARGV0,ARGV1,...]\n"
+         " ($0 = PROG.ape) [ARGV1,ARGV2,...]\n"
```
jart, line 1015: please insert an additional space on the left so the `\[` lines up vertically.

ape/ape-m1.c
mrdomino: Done out of band + in comments

ape/ape-m1.c
```diff
@@ -948,23 +959,63 @@ int main(int argc, char **argv, char **envp) {
...
   if ((long)sp2 & 15) ++sp2;
   for (; n > 0; n -= pagesz) {
     ((char *)sp2)[n - 1] = 0;
   }
-  memmove(sp2, sp, (auxv - sp) * sizeof(long));
+  if (islogin) {
+    memmove(sp2, sp, 2 * sizeof(long));
+    memmove(dash_l, "-l", 3);
+    *((char **)sp2 + 2) = dash_l;
+    memmove(sp2 + 3, sp + 2, (auxv - sp - 2) * sizeof(long));
```
jart, line 987: This is way too much ugliness just to support such a narrow feature. There's a better way. Do `setenv("_", getenv("SHELL"))`. Cosmo's `GetProgramExecutableName()` really likes to discover the executable path using `argv[0]` but it'll fall back to trying `$_` if it doesn't work. So if you use my suggestion, runtime services shouldn't be impacted if you just leave `argv` the way the kernel supplied it.

jart (changes requested):

ape/ape-m1.c
mrdomino: Do you mean a literal `setenv(3)`? That does not modify `envp`, so we will need to switch to getting `envp` out of `environ`. (We do not have `_` at all when called as a login shell.)

I'll take another look in a little while and try to either use `environ` or permute `envp` appropriately.

jart: Is there any reason why we can't just do nothing? Having a broken `argv[0]` only means you can't do things like `--strace` your login shell. But you wouldn't have been able to pass that flag anyway. I'm perfectly fine with trading that away if it means we can make as few changes to APE Loader as possible. What's the simplest change we can do that makes zsh work?

mrdomino: By nothing, you mean don't change `argv[0]` and don't setenv `_`?

If we do (don't do?) that, we get a very broken zsh. It does not know about anything under `/zip`, and does not have the ability to do things like e.g. remove characters from the display when you type backspace.

Now that I look at it, much of this brokenness is probably that zipos is not working; `__zipos_init` depends on `GetProgramExecutableName()`.

I was actually leaning in the opposite direction re changing APE loader. I was eyeing all those juicy stack bytes at the top of `main` and in `PathSearcher`, thinking about making everything not needed by the loaded program `static` and having a separate `Configure` method that would figure out what the invocation was supposed to be.

ape/ape-m1.c
mrdomino: Filed #979 for `ape/loader.c`.

ape/ape-m1.c
mrdomino: Also I looked at the code again and realized that I was incorrect about cosmopolitan's behavior earlier — `$_` supersedes `argv[0]` if it is present, so everything should be fine on that end. Only the loader would have to change.

jart (approved):

PR #980, opened 2023-12-04
Keep argv[0], add COSMOPOLITAN_PROGRAM_EXECUTABLE

mrdomino:
`argv[0]` was previously being used as a communication channel between the loader and the binary, giving the binary its full path for use e.g. in `GetProgramExecutableName`. But `argv[0]` is not a good channel for this; much of what made 2a3813c so gross is due to that.

This change fixes the issue by preserving `argv[0]` and establishing a new communication channel: `COSMOPOLITAN_PROGRAM_EXECUTABLE`.

The M1 loader will always set this as the first variable. Linux should soon follow. On the other side, `GetProgramExecutableName` checks that variable first. If it sees it, it trusts it as-is.

A lot of the churn in `ape/ape-m1.c` in this change is actually backing out hacks introduced by 2a3813c; the best comparison is:

https://github.com/jart/cosmopolitan/compare/2a3813c6%5E...mrdomino:cosmopolitan:underscore.

### DIFF
--- a/ape/ape-m1.c
+++ b/ape/ape-m1.c
@@ -16,6 +16,7 @@
 │ TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR             │
 │ PERFORMANCE OF THIS SOFTWARE.                                                │
 ╚─────────────────────────────────────────────────────────────────────────────*/
+#include <assert.h>
 #include <dispatch/dispatch.h>
 #include <dlfcn.h>
 #include <errno.h>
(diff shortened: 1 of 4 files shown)


mrdomino:
As mentioned out of band, the right diff for this change is `git diff $(git merge-base master HEAD)^..` since most of the deletions are lines that were added just previously.

mrdomino:
Apparently `_` is like `AT_EXECFN` except it is set by the shell, not the kernel. It is the string of the shell invocation that led to this command being executed. It does not always exist, even on M1; e.g. for a login shell, it is absent. Overall, it is not a reliable way of getting the executable path.

`COSMOPOLITAN_PROGRAM_EXECUTABLE` is supposed to always be the absolute path of the actual binary file that is being executed.

The reason for it to be the first environment variable is just performance and simplicity of the code on the other side. Is there some reason not to do it that way?

mrdomino:
Had a walk and thought about it, I'm gonna change `GetProgramExecutableName` to scan the whole env but keep injecting it into the first position in the loader. I may also try to actually benchmark the code to see how much it matters.

jart (approved):

jart:
Great work. Thank you!

ghaerr:
> The reason for it to be the first environment variable is just performance and simplicity of the code on the other side. Is there some reason not to do it that way?

No, and the code swapping the variables is kinda cool. Agreed probably a good idea to scan the whole env in `GetProgramExecutableName` so that it's less dependent on an ape loader specific setup of the environment array.

PR #981, opened 2023-12-04
GetProgramExecutableName fixes for `_` and -$SHELL

mrdomino:
- `_` was getting clobbered by a lack of a return at the end of the CopyString block.

- `_` may need to be absolutized.

- Handle the special case of `argv[0]` equal to -$SHELL.

### DIFF
--- a/libc/calls/getprogramexecutablename.greg.c
+++ b/libc/calls/getprogramexecutablename.greg.c
@@ -23,6 +23,7 @@
 #include "libc/cosmo.h"
 #include "libc/dce.h"
 #include "libc/errno.h"
+#include "libc/intrin/getenv.internal.h"
 #include "libc/serialize.h"
 #include "libc/limits.h"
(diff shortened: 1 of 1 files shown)


libc/calls/getprogramexecutablename.greg.c
```diff
@@ -49,6 +49,22 @@ static inline int IsAlpha(int c) {
   return ('A' <= c && c <= 'Z') || ('a' <= c && c <= 'z');
 }
 
+static inline const char *BaseName(const char *path) {
+  const char *b = strrchr(path, '/');
+  return b ? b + 1 : path;
+}
+
+static inline char *GetEnv(char** envp, const char* var) {
```
jart, line 57: Please delete this function and use `__getenv()`

jart (changes requested):

mrdomino:
Closing in favor of #980.

PR #985, opened 2023-12-05
ape/loader.c: $prog.ape, argv[0]/environment

mrdomino:


mrdomino:
(Closing because there is no way I can find to keep this PR open against this repo but base it off of my m1-register branch.)

PR #995, opened 2023-12-09
Use `gid_t` and `uid_t` in `passwd.h`

tempoz:
`group.gr_gid` was the wrong type (`int32_t` when `gid_t` is `uint32_t`). Just replaced all the types for uids and gids with the appropriate `uid_t` or `gid_t`, which incidentally fixes that inconsistency.

### DIFF
--- a/third_party/musl/passwd.h
+++ b/third_party/musl/passwd.h
@@ -8,8 +8,8 @@ struct FILE;
 struct passwd {
   char *pw_name;
   char *pw_passwd;
-  uint32_t pw_uid;
-  uint32_t pw_gid;
+  uid_t pw_uid;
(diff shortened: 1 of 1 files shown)


jart (commented):
The integration tests will most likely fail on the `$(THIRD_PARTY_MUSL_A_HDRS:%=o/$(MODE)/%.ok)` check due to the header not including `libc/call/weirdtypes.h`. I'd recommend only changing `gr_gid` to be `uint32_t`. I like it that way in the mono repo. Less include complexity.

jart (approved):
Nevermind. I see passwd.h somehow already includes weirdtypes.h

Issue #978, opened 2023-12-03
after vfork and closing zipos file, __isfdkind is incorrect


G4Vi:
I made a test:

```c
TEST(zipos, close_after_vfork) {
  if (!IsLinux()) return;
  int zip_fd = open("/zip/life.elf", O_RDONLY);
  ASSERT_NE(-1, zip_fd);
  SPAWN(vfork);
  EXPECT_EQ(1, __isfdkind(zip_fd, kFdZip));
  close(zip_fd);
  int fd = open("/etc/hosts", O_RDONLY);
  ASSERT_NE(-1, fd);
  EXPECT_EQ(0, __isfdkind(fd, kFdZip));
  exit(0);
  EXITS(0);
}
```

Fails on  `EXPECT_EQ(0, __isfdkind(fd, kFdZip));`

@mrdomino could you please take a look? I'm trying to revive https://github.com/jart/cosmopolitan/pull/888 and maintain vfork compat if possible.

mrdomino:
@jart what should happen here? Is it safe to lock `g_fds` after vfork? I'm thinking maybe if `__vforked` then we don't touch refcounts but do clear `g_fds`?

mrdomino:
(No, that probably won't work as such. I'll think about it some more.)

jart:
I would imagine that it is safe to lock a mutex after vfork, provided you unlock it before calling execve().

mrdomino:
Ok, took a bit to look at this.

Looks like the problem is straightforward. `close()` does not call `__releasefd()` if `__vforked`.

The comment says that it wants to avoid locking so `close()` doesn't need to block signals.

So one proposal is: if `__vforked`, have `close()` block signals, lock fds, and call `__releasefd()`.

Unfortunately, this does not work: https://github.com/jart/cosmopolitan/compare/master...0bc5e63d19bb2569aa316f1b3d546cf2a2968415 dies with `nsync panic: attempt to nsync_mu_unlock() an nsync_mu not held in write mode`.

jart:
[...]

PR #1004, opened 2023-12-12
zipos / vfork cleanup

mrdomino:
Remaining followup from #978.

### DIFF
--- a/libc/calls/close.c
+++ b/libc/calls/close.c
@@ -49,7 +49,7 @@ static int close_impl(int fd) {
   }
 
   if (__isfdkind(fd, kFdZip)) {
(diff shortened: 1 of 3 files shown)


libc/calls/close.c
```diff
@@ -49,7 +49,7 @@ static int close_impl(int fd) {
   }
 
   if (__isfdkind(fd, kFdZip)) {
-    if (!__vforked && _weaken(__zipos_close)) {
+    if (_weaken(__zipos_close)) {
```
mrdomino, line 52: Actually, question here: why not `npassert(_weaken(__zipos_close))`? How are we winding up with a `kFdZip` file without `__zipos_close`?

PR #1007, opened 2023-12-13
fix path handling in bind/connect

### LINKED ISSUE #1000: Fix Unix socket on windows (when path is absolute path like /tmp/socket)
currently using unix socket on windows get ENETDOWN, it is because winsock api requires a Windows-style path, but cosmopolitan pass the original filename into the application, which cause above error

codehz:
fix #1000
I didn't patch the getsockname for unix socket, so it would be win32 path instead of unix path, I don't think anyone really cares about the sockname of a unix socket...

### DIFF
--- a/libc/sock/connect-nt.c
+++ b/libc/sock/connect-nt.c
@@ -20,6 +20,7 @@
 #include "libc/atomic.h"
 #include "libc/calls/struct/fd.internal.h"
 #include "libc/calls/struct/sigset.internal.h"
+#include "libc/calls/syscall_support-nt.internal.h"
 #include "libc/cosmo.h"
 #include "libc/errno.h"
 #include "libc/mem/mem.h"
@@ -34,6 +35,7 @@
 #include "libc/sock/struct/sockaddr.h"
 #include "libc/sock/syscall_fd.internal.h"
 #include "libc/sock/wsaid.internal.h"
+#include "libc/sysv/consts/af.h"
 #include "libc/sysv/consts/o.h"
 #include "libc/sysv/errfuns.h"
 
@@ -140,6 +142,14 @@ static textwindows int sys_connect_nt_impl(struct Fd *f, const void *addr,
 
 textwindows int sys_connect_nt(struct Fd *f, const void *addr,
                                uint32_t addrsize) {
+  struct sockaddr_un *sun, nt_sun;
+  if (f->family == AF_UNIX && ((struct sockaddr *)addr)->sa_family == AF_UNIX &&
+      addrsize >= sizeof(struct sockaddr_un)) {
+    sun = (struct sockaddr_un *)addr;
+    nt_sun.sun_family = AF_UNIX;
+    if (__mkwin32_sun_path(sun->sun_path, nt_sun.sun_path) == -1) return -1;
+    addr = &nt_sun;
+  }
   sigset_t mask = __sig_block();
   int rc = sys_connect_nt_impl(f, addr, addrsize, mask);
   __sig_unblock(mask);
--- a/libc/sock/connect-nt.c
+++ b/libc/sock/connect-nt.c
@@ -147,7 +147,7 @@ textwindows int sys_connect_nt(struct Fd *f, const void *addr,
       addrsize >= sizeof(struct sockaddr_un)) {
     sun = (struct sockaddr_un *)addr;
     nt_sun.sun_family = AF_UNIX;
-    if (__mkwin32_sun_path(sun->sun_path, nt_sun.sun_path) == -1) return -1;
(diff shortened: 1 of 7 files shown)


libc/sock/connect-nt.c
```diff
@@ -140,6 +142,14 @@ static textwindows int sys_connect_nt_impl(struct Fd *f, const void *addr,
...
 textwindows int sys_connect_nt(struct Fd *f, const void *addr,
                                uint32_t addrsize) {
+  struct sockaddr_un *sun, nt_sun;
+  if (f->family == AF_UNIX && ((struct sockaddr *)addr)->sa_family == AF_UNIX &&
+      addrsize >= sizeof(struct sockaddr_un)) {
+    sun = (struct sockaddr_un *)addr;
+    nt_sun.sun_family = AF_UNIX;
+    if (__mkwin32_sun_path(sun->sun_path, nt_sun.sun_path) == -1) return -1;
+    addr = &nt_sun;
+  }
```
G4Vi, line 152: Rather than duplicating this `sockaddr_un` conversion code between `bind-nt.c` and here, it would be better to add `__convert_nt_to_sockaddr_un` and `__convert_sockaddr_un_to_nt` functions similar to the  [`__convert_bsd_to_sockaddr` and `__convert_sockaddr_to_bsd` functions](https://github.com/jart/cosmopolitan/blob/6e6fc38935054db0534d5af4fb99c6193305b946/libc/sock/sockaddr.c#L43-L62), but specialized to  `sockaddr_un`.

test/libc/sock/unix_test.c
```diff
@@ -116,6 +119,29 @@ TEST(unix, stream) {
   munmap(ready, 1);
 }
 
+TEST(unix, stream_absolute) {
+  int ws;
+  if (IsWindows() && !IsAtLeastWindows10()) return;
+  atomic_bool *ready = _mapshared(1);
+  // TODO(jart): move this line down when kFdProcess is gone
```
G4Vi, line 126:
```suggestion
```

test/libc/sock/unix_test.c
```diff
@@ -70,17 +70,20 @@ TEST(unix, datagram) {
...
+  struct sockaddr_un addr = {AF_UNIX};
+  strcpy(addr.sun_path, path);
   ASSERT_SYS(0, 3, socket(AF_UNIX, SOCK_STREAM, 0));
+  unlink(path);
+  errno = 0;
   ASSERT_SYS(0, 0, bind(3, (void *)&addr, len));
   bzero(&addr, sizeof(addr));
   ASSERT_SYS(0, 0, getsockname(3, (void *)&addr, &len));
-  ASSERT_EQ(2 + 8 + 1, len);
+  if (check_path) ASSERT_EQ(2 + strlen(path) + 1, len);
```
G4Vi, line 84: Assuming `getsockname` and `getpeername` is implemented for AF_UNIX on Windows, implementing them for Windows AF_UNIX in Cosmo would be greatly appreciated for both the functionality and to avoid making tests platform specific as much as possible.

G4Vi (changes requested):
Sorry for the long delay. I've recently been working on improving Cosmo's Windows networking and would like to get this in too! This looks great, just needs some cleanup and potentially `getsockname` and `getpeername` implementations. 

Also have you assigned copyright to Justine as described here: https://github.com/jart/cosmopolitan/blob/master/CONTRIBUTING.md#copyright-assignment ?

test/libc/sock/unix_test.c
codehz: the problem is the path is no longer starts with /tmp (because it has been rewritten). and there is no way to recover it(unable to distinguish whether it is a real path manually set by the user)

codehz:
> Also have you assigned copyright to Justine as described here: https://github.com/jart/cosmopolitan/blob/master/CONTRIBUTING.md#copyright-assignment ?

just sent the email..

test/libc/sock/unix_test.c
G4Vi: Sounds like there's no easy way to implement these as Linux does, we can skip these for now. I noticed in various situations on Linux and FreeBSD when unimplemented return a `sun_path` of `\0` and out length of `3`.

G4Vi (changes requested):
One last thing, sorry I didn't spot this before. Since this PR was created our `.clang-format` has been updated with `AllowShortIfStatementsOnASingleLine: false` among other settings. Could you please rebase and rerun the autoformatter?

Edit: also update includes to fix ci build.

jart:
This project won't accept your contributions, for reasons that should be perfectly clear to you @codehz.

PR #1012, opened 2023-12-15
Loader path security

### LINKED ISSUE #991: Loader path security
There is currently a security issue in the way the loader passes paths to the binary.

For argument's sake, consider:

- user installs a setuid root `sudo.com` that reads its configuration out of /zip/etc/sudoers
- user does `ln -sf /usr/local/bin dir`
- user runs `dir/../bin/sudo.com bad.sh`
- user wins a race with the loader, changing `dir` to point to `bad` from which `../bin/sudo.com` is a zip archive with a user controlled sudoers file in it, between when the loader finds the binary and when the binary opens itself

There are a few mitigations that can be done:

1. If the loader detects that the binary it has been asked to run is setuid, it can refuse to run it from a non-canonical path (i.e. a path containing `..`.)
2. The loader can always do (the moral equivalent of) `realpath` on the path it resolves before passing it to the binary.
3. More speculatively, binaries can embed their own hash at a known label. Whenever they open themselves, they can compare the hash of the binary they open against their own embedded value.

1 is a no-brainer. 2 may be a good idea but I'm not sure it buys much beyond 1. 3 is a very neat sounding idea to me; thoughts?

EDIT: actually 1 doesn't help that much. Consider: `ln -sf /usr/local dir; ~/dir/bin/sudo.com`. There is probably some adjacent mitigation that is similar in spirit though.

mrdomino:
The ape loader now passes the program executable name directly as a register. `x2` is used on aarch64, `%rdx` on x86_64. This is passed as the third argument to `cosmo()` (M1) or `Launch` (non-M1) and is assigned to the global `__program_executable_name`.

`GetProgramExecutableName` now returns this global's value, setting it if it is initially null. `InitProgramExecutableName` first tries exotic, secure methods: `KERN_PROC_PATHNAME` on FreeBSD/NetBSD, and `/proc` on Linux. If those produce a reasonable response (i.e., not `"/usr/bin/ape"`, which happens with the loader before this change), that is used. Otherwise, if `issetugid()`, the empty string is used. Otherwise, the old argv/envp parsing code is run.

The value returned from the loader is always the full absolute path of the binary to be executed, having passed through `realpath`. For the non-M1 loader, this necessitated writing `RealPath`, which uses `readlinkat` of `"/proc/self/fd/[progfd]"` on Linux, `F_GETPATH` on Xnu, and the `__realpath` syscall on OpenBSD. On FreeBSD/NetBSD, it punts to `GetProgramExecutableName`, which is secure on those OSes.

With the loader, all platforms now have a secure program executable name. With no loader or an old loader, everything still works as it did, but setuid/setgid is not supported if the insecure pathfinding code would have been needed.

Fixes #991.


libc/calls/getprogramexecutablename.greg.c
```diff
@@ -51,6 +52,12 @@ static inline int IsAlpha(int c) {
 }
 
 static inline void InitProgramExecutableNameImpl(void) {
+  if (__program_executable_name) {
+    /* already set by the loader */
+    return;
+  }
+  npassert(!issetugid());   // TODO(mrdomino): pathname security
```
mrdomino, line 59: This could easily go after `IsWindows()` and `IsMetal()`... should it, or should this feature be the same on all platforms?

I know of no other problems with setuid/setgid programs, so this function seems like a fine place for the check.

I decided against printing a friendlier error message here for two reasons: one, this assert will hopefully be short-lived. Two, with assertions on, the assert failure should be clear enough already to someone who has wound up with a setuid/setgid ape program on their system.

mrdomino:
(I somewhat doubt this is ready to merge, but just in case, please hold off until the loader `realpath()` change is in.)

libc/calls/getprogramexecutablename.greg.c
jart: How is making the binary crash a better option? Can't we just use the exotic kernel guaranteed paths only when it's setuid?

mrdomino: You mean actually implement the correct change rather than punt to later? Fine....

jart: Thank you!

mrdomino: Ok, fixed.

jart (approved):
Looks good. Thanks!

PR #1013, opened 2023-12-15
Mint APE Loader v2.0

mrdomino:
I opted for 2.0 rather than 1.10 for two reasons: one, believe it or not, it saves a byte (on x86_64 at least.) Two, (and less jokingly,) the change to no longer modify `argv[0]` is breaking.


mrdomino:
How much of a use case is there for `$prog.ape` on non-M1? Is it needed in x86-64 xnu as well? Maybe BSD?

I don't have a use case for it yet, so I thought to leave it out for now.

mrdomino:
Investigating why the binary sizes are so different for `*.com`; I will try rebuilding against `MODE=tiny` or somesuch.

mrdomino:
tiny cocmd.com is 14K *smaller* than master... gonna hold off on doing anything further with that and investigate the `GetProgramExecutableName` test failure.

mrdomino:
😲

```
       If  the  pathname argument of execve() specifies an interpreter script, then in‐
       terpreter will be invoked with the following arguments:

           interpreter [optional-arg] pathname arg...

       where pathname is the pathname of the file specified as the  first  argument  of
       execve(),  and arg...  is the series of words pointed to by the argv argument of
       execve(), starting at argv[1].  Note that there is no way  to  get  the  argv[0]
       that was passed to the execve() call.
```

`execve` not going through `binfmt_misc` would explain the test case behavior…

mrdomino:
But [the source](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/exec.c) does appear to be checking binfmt...

mrdomino:
Interesting, the test fails for `MODE=tiny` but not `MODE=tinylinux`...

mrdomino:
#1014 describes the bug, #1015 has the workaround and has been merged into this PR.

mrdomino:
Do this later after the realpath stuff is in.

PR #1015, opened 2023-12-15
Disable argv[0] tests on non-XnuSilicon

mrdomino:


mrdomino:
This can be merged as-is to make the tests pass after the new loader has been minted, or `sys_execve` can be modified and this change dropped.

mrdomino:
The correct feature test is now spelled out in a comment, and the de facto `!IsXnuSilicon()` more closely matches reality.

mrdomino:
Waffled on merging this, but I think I will push it to reduce noise on BSD tests, and just rewrite it in terms of something like the hypothetical `cosmo_execve` soon.

PR #1016, opened 2023-12-15
Fix loader usage, shave off a few bytes

mrdomino:
The `-f` functionality was removed in ape loader version 1.5.

Shaves some bytes off here and there; functionality remains the same modulo usage string, but each change reduces binary size slightly.


mrdomino:
Calling it; this is ready to go if we want it. If I understand correctly it's not going to make a difference with everything being 4k-aligned, but I believe this is a (slight) improvement.

jart (approved):

PR #1018, opened 2023-12-16
Make ape-m1 not die on realpath error

mrdomino:
The loader was dying after the first PATH component if the command was not found. :(


mrdomino:
We probably don't want `realpath` at all anymore, but I'm holding off on making that change until I have a firmer idea of what we do want.

mrdomino:
Superseded by #1024.

PR #1021, opened 2023-12-17
Update redbean ProgramDirectory to return a list of previously set directories

pkulchenko:


mrdomino (commented):
Can you give me a bit of context on why this is desired?

Also, what does ProgramDirectory return if called with other than nil?

pkulchenko:
@mrdomino, sure. Right now external directories can be added by calling `ProgramDirectory`, but it's impossible to get a list of currently assigned directories (those that are added with `-D` or `ProgramDirectory`), which is useful in getting those directories re-scanned to detect new resources. There have been several requests for this functionality and it's difficult to implement without having the list of directories.
When `ProgramDirectory` is called to add a new value, nothing is returned, as I wouldn't expect the developer to need a value and didn't want to unnecessarily create and populate a table.

mrdomino (approved):
LGTM

PR #1022, opened 2023-12-17
Cover more cases for detecting ape loader

mrdomino:
Fixes a regression in GetProgramExecutableName on Linux against old loaders. In the loader case, /proc/self/exe gives the loader's path. We tried to detect this by checking for `/usr/bin/ape`. But that is only one of the possible places the loader could be.


pkulchenko:
@mrdomino, confirmed working; thank you!

pkulchenko (approved):

PR #1026, opened 2023-12-17
Claim that /zip itself is user-modifiable

mrdomino:
This is in line with what I have learned about the TOCTOU stuff regarding ape executables. If there is ever a check for whether we are running in a secure set-id context, this can be disabled by that check.

Adds basic sanity test for zipos stat as well.


mrdomino:
I think generally `/zip` should reflect the (transitive most-permissive) owner/mode of the binary we are running as. If we used `KERN_PROC_PATHNAME` or friends, this is knowable and can be set appropriately. If not, it is unknowable. "You own it and can write to it" is one option; maybe a better one is "root, 01777" ignoring umask.

mrdomino:
I do not like this change anymore. Something like it should probably be done, but I'm not sure what yet.

If we have used a secure pathfinding mechanism (always true if `issetugid()` assuming #1029 goes through), then I think we can just report `/zip`’s owner/mode as the owner/mode of `GetProgramExecutableName()`. Otherwise "root, 0777" is probably best.

PR #1029, opened 2023-12-18
sanity-check set-id interpreter script paths

mrdomino:
This actually implements proposal 2a, rolled back from #1024, of this:

https://gist.github.com/mrdomino/2222cab61715fd527e82e036ba4156b1

From the first commit:

> There is an argument to be made for trapping in [a] constructor here,
> and there is also an argument to be made for failing open like we did
> previously. We pick a middle ground: GetProgramExecutableName returns
> empty string. Logic being that ensuring the sanity of whatever set-id
> script mechanism user is running is above our pay grade, but the path
> TOCTOU between loader and binary is not.


mrdomino:
Tested on M1 Mac as so:

```console
% chmod u+s /usr/local/bin/ape
% chown root /usr/local/bin/ape
% ape - o/aarch64/examples/printargs.com printargs.com
[… GetProgramExecutableName = "" …]
% ape - /dev/fd/3 printargs.com 3<o/aarch64/examples/printargs.com
[… GetProgramExecutableName = "/dev/fd/3" …]
```

And obviously confirmed that things like `/dev/fd/../../path/to/printargs.com` do not break it.

jart (approved):
Looks good. Passes fleet. Thanks for using clang-format.

PR #1030, opened 2023-12-18
Fix the build on Apple Silicon

mrdomino:
Build was broken by third_party/bash due to `__floatditf` not being found, even though `rg __floatditf` shows no results other than its definition.

I don't know if __aarch64__ is the right check, but hopefully this commit will be short-lived.

### DIFF
--- a/third_party/bash/builtins_printf.c
+++ b/third_party/bash/builtins_printf.c
@@ -193,6 +193,14 @@ static intmax_t tw;
 static char *conv_buf;
 static size_t conv_bufsize;
 
+#ifdef __aarch64__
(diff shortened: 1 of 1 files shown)


jart (changes requested):
Delete this PR. 

```
master jart@nightmare:/opt/cosmo$ grep -R floatditf
third_party/compiler_rt/floatditf.c://===-- lib/floatditf.c - integer -> quad-precision conversion ----*- C -*-===//
third_party/compiler_rt/floatditf.c:COMPILER_RT_ABI fp_t __floatditf(di_int a) {
```

Add a missing build dependency.

PR #1032, opened 2023-12-18
Fix indentation, make llvm happy

mrdomino:
clang says "label at end of switch statement is a C2x extension."


mrdomino:
Assuming this is a trivial enough change to just merge.

PR #1034, opened 2023-12-18
Fix redbean Fetch redirect with relative URL

pkulchenko:


mrdomino:
I'd like to try and review this while @jart is ~away; however I'm unfamiliar with redbean's code. Could you talk me through what's going on here a bit?

pkulchenko:
`Fetch` allows redbean Lua code to retrieve external resources over HTTP/S. As part of the retrieval it supports redirects with the `Location` header providing a URL for the resource to be retrieved. The current implementation expects that Location URL to be absolute (including the domain name), which is not always the case, so this patch allows the Location value to contain just a path (it can be absolute or relative) and will either use the value from the Location header (if both schema and host are present) or add the domain name used for the previous request (the one that returned the Location header). When the path content is relative, it will merge that path with the path from the previous request, such that `/redir/more` path redirected to `less` becomes `/redir/less`, and `/redir/more` path redirected to `/less` becomes `/less`.

I actually noticed one detail that doesn't follow the specification (https://www.rfc-editor.org/rfc/rfc9110#field.location): the fragment should be preserved from the original URL and applied if the fragment is not provided in the Location field. Let me address this and update the PR.

pkulchenko:
@mrdomino, @jart, this should be good to go. This fixes the issue with Fetch handling of redirects without domain names.

mrdomino:
I'm unfamiliar enough with the code that I'll punt to @jart if she's around; otherwise I can take a look.

mrdomino (approved):
LGTM

PR #1036, opened 2023-12-19
Pass exename as %r8/x4, os as x3

mrdomino:
Makes `cosmo()` and `Launch()` now have the same argument 4 and 5, and aarch64 receives an `os` flag, which `ape-m1.c` now passes. Opens the door for adding support for aarch64 OSes beyond Linux and Mac.

The tradeoff here is whether to make `cosmo()`’s ABI clean, or to make the aarch64/x86_64 registers match. I chose the latter, mostly so that the comment in `launch.S` would be true of both architectures. (N.B. it was not previously true of both architectures; `x3` was not being passed through.)

I initially tried to do a decentralized init of `__program_executable_name` on x86_64, but none of `%r8` through `%r11` was surviving intact up to the point of the initialization, so I gave up and rolled it back.

EDIT: there is no reason to change exename to `%r8` if it is also being clobbered before decentralized init, so probably none of the below applies.

This change will break compatibility for people running prerelease loaders built off of master. Since the os is now getting passed as arg 4 (not 3), the effect of the breakage is relatively small; programs just won't receive `__program_executable_name` from the loader, which will mess up anything that depends on `GetProgramExecutableName` if the binary cannot use `argv[0]` or `_` to locate itself. This will happen if the binary is invoked as e.g. `ape $prog` out of `$PATH`, or `ape - $prog $args` with `argv[0]` not a path to the binary (which means it will happen if `__sys_execve` fails, so on M1 Mac), or `$prog.ape`, again out of `$PATH`. Just running a binary as plain `$prog` out of `$PATH` will still work because the OS edits `argv[0]`.

I suppose it would be possible to reduce this breakage by making `TryPath` smarter (check for shell, use `commandv` if the path doesn't contain a slash.) I don't think it's worth it though, since just replacing the loader with either of `build/bootstrap/ape.elf` or a freshly compiled one will resolve it. ~and anyone subject to this breakage has already done that at least once anyway.~

The tricky case is when someone ran a binary that installed a `.ape-1.9` that is a prerelease loader. I propose to fix this by having `ape/apeinstall.sh` go on a hunt for loaders at the other possible locations, and replace any that it finds.


mrdomino:
Oh, you can probably just keep exename at `%rdx` and start passing os as `%rcx`.

mrdomino:
If I'm understanding correctly, it looks like the clobbering of `%r10` and `%r11` is done by `libc/sysv/systemfive.S`. I suppose it might be possible to rewrite it to preserve `%r11`, but it seems like that'd be pretty involved. Anyway, I don't know how robust `%r11` would be - again if I'm understanding what I've heard correctly, I think NetBSD would clobber it between `Launch` and our own entry point?

It doesn't seem reasonable to expect `%r8` or `%r9` to stay untouched that long. So I think `%rdx` is as good as it gets, and should be kept. I'll close this and may at some point try adding the `os` parameter as a separate PR.

PR #1037, opened 2023-12-19
Remove a redundant load in zipos read/seek

mrdomino:
When `h->pos` has changed to something other than `SIZE_MAX`, we don't need the extra atomic load.

### DIFF
--- a/libc/runtime/zipos-read.c
+++ b/libc/runtime/zipos-read.c
@@ -38,17 +38,15 @@ static ssize_t __zipos_read_impl(struct ZiposHandle *h, const struct iovec *iov,
     return eisdir();
   }
   if (opt_offset == -1) {
-    while (true) {
-      start_pos = atomic_load_explicit(&h->pos, memory_order_relaxed);
+  Restart:
+    start_pos = atomic_load_explicit(&h->pos, memory_order_relaxed);
+    do {
       if (UNLIKELY(start_pos == SIZE_MAX)) {
-        continue;
+        goto Restart;
       }
-      if (LIKELY(atomic_compare_exchange_weak_explicit(
-              &h->pos, &start_pos, SIZE_MAX, memory_order_acquire,
-              memory_order_relaxed))) {
-        break;
-      }
-    }
+    } while (!LIKELY(atomic_compare_exchange_weak_explicit(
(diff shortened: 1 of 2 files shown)


mrdomino:
N.B. were it not for the `SIZE_MAX` locking in `read`, it would be possible to replace the `SEEK_SET` and `SEEK_END` cases with a simple `atomic_store_explicit`. It would almost but not quite be possible to replace the `SEEK_CUR` case with an `atomic_fetch_add`, but I think the error cases prevent it.

I'm indifferent on keeping the `SIZE_MAX` locking vs making `read` happen all in one go. My tendency would be to keep it as-is for now.

pkulchenko:
@mrdomino, I'm most surely missing something here, but when `h->pos` returns something other than `SIZE_MAX`, wouldn't `atomic_compare_exchange_weak_explicit` called as before the change?

mrdomino:
In both cases (before this change and with this change), there is an `atomic_load_explicit`, a comparison against `SIZE_MAX`, and an `atomic_compare_exchange_weak_explicit`. In the usual case as well as the case where the load returns `SIZE_MAX`, the flow is identical before and after the change. (In the usual case, the compare-exchange succeeds and the loop breaks. In case the load returns `SIZE_MAX`, the thread spins until it changes.)

The difference is in the case where `h->pos` changed between the load and the compare-exchange. Before this change, what happened was:

1. initial `atomic_load_explicit`
2. `compare_exchange_weak_explicit` fails, setting `pos` to the value that `h->pos` was changed to.
3. redundant `atomic_load_explicit` also sets `pos` to `h->pos`’s current value.
4. `compare_exchange_weak_explicit` (eventually) succeeds.

After the change:

1. initial `atomic_load_explicit`
2. `compare_exchange_weak_explicit` fails, resetting `pos`
3. `compare_exchange_weak_explicit` reruns against the updated `pos`, (eventually) succeeding.

So it just eliminates one step in the case where there is contention.

pkulchenko:
Got it: before the change `atomic_load_explicit` was executed every time inside the loop, but after the changes it's executed the first time (same as before) and then only if `start_pos == SIZE_MAX`. Thanks!

pkulchenko (approved):
Looks good after the explanation; thanks!

PR #1038, opened 2023-12-23
Reorder Launch arguments, pass aarch64 os

mrdomino:
Third and fourth arguments are now identical between cosmo and Launch.
By passing sp as argument 4, we save a bit of register juggling.

Fourth argument (os) is now always passed by the loader on aarch64. It
is not yet processed by cosmo. Pushing this change separately, as the
cosmo side turns out to be somewhat more involved.


mrdomino:
This can be merged as-is or ignored in favor of #1042.

PR #1041, opened 2023-12-24
aarch64 FreeBSD support (loader side)

mrdomino:
- [x] Adds FreeBSD to the aarch64 loader's support vector.
- [x] Checks in `build/bootstrap/sysv2freebsd.com`, which brands an ELF file as FreeBSD.
- [x] Brands the aarch64 loader as FreeBSD. (N.B. the x86_64 loader is already branded as FreeBSD.)

Tested on aarch64 NixOS and FreeBSD. Loader gets as far as printing usage on both platforms.

Logically depends on #1038. Can be merged prior to the cosmo side of aarch64 os detection (since aarch64 FreeBSD is not yet supported anyway), but probably makes the most sense to wait on that as well.


mrdomino:
Superseded by #1042.

PR #1042, opened 2023-12-25
aarch64 loader passes os

mrdomino:
Both aarch64 loaders (`loader.c` and `ape-m1.c`) now pass the detected OS to cosmopolitan, using the same calling convention as x86_64 for exename and os.


mrdomino:
There is a question of whether to merge this now or wait until FreeBSD/aarch64 cosmopolitan support is ready. There is also a question of whether to do that support in this PR or in a separate one.

My inclination is "wait" and "do it in a separate PR", so this is a draft for now. But I see no reason this couldn't be reviewed and merged as-is, if the loader changes are wanted sooner or something.

jart (approved):
Looks good. Cloned locally and confirmed it passes fleet. Thank you!

PR #1043, opened 2023-12-26
Rename __zipos_free -> __zipos_drop

mrdomino:
Removes the separate decref function, uses keep/drop in the internal API.


pkulchenko (approved):
LGTM. Thanks!

PR #1046, opened 2023-12-31
Re-enable/fix non-argv[0] exe name tests

mrdomino:
These tests should now pass in all cases and if they don't it's a bug.

One buggy case is if they're run against the short-lived loader that used a stubbed `RealPath` on FreeBSD and NetBSD; if they fail on those platforms it's likely for that reason.

Still TODO, preferably in a future PR:

- [ ] Make `GetProgramExecutableName` reliable for assimilated XNU binaries; it seems like this might be doable with what [`dyld::getExecutablePath`][0] (and therefore `_NSGetExecutablePath`) uses.
- [ ] Enable `argv[0]` tests on all platforms, probably by using the to-be-written `cosmo_execve`.

There is probably nothing for assimilated OpenBSD binaries, so the check introduced will probably never completely go away.

[0]: https://opensource.apple.com/source/dyld/dyld-195.6/src/dyld.cpp.auto.html


jart (approved):
Looks good. As of the APE 1.10 deployment, this PR passes fleet.

PR #1047, opened 2023-12-31
Remove COSMOPOLITAN_PROGRAM_EXECUTABLE

mrdomino:


mrdomino:
Will break people who have installed certain versions of the 1.9 loader; if desired, `GetProgramExecutableName` can print a warning to rerun `apeinstall.sh` if it sees that variable set.

Will not break anything for anyone who has not run `apeinstall.sh` or equivalent, now that we have minted a new loader.

jart (approved):
I'm not concerned about breaking code in-between releases. This environment variable was added *after* 3.1.3 was shipped. So there's no harm done.

PR #1048, opened 2024-01-01
Prepend getcwd to exename early in init

### LINKED ISSUE #1039: GetProgramExecutableName is broken with relative paths if called after `chdir`
If `GetProgramExecutableName` is using either `argv[0]` or `__program_executable_name`, and they are relative, and it is called after a `chdir`, then it will return the wrong answer.

The solution is probably going to be some variant of "move the `getcwd` into a constructor."

It sure would be nice if that constructor did not get linked if the program was never going to call `GetProgramExecutableName`.

mrdomino:
Fixes #1039.


jart (approved):
TY

PR #1053, opened 2024-01-04
third_party: Add libunwind

trungnt2910:
Added libunwind from LLVM 17.0.6.
The library includes functions required for C++ exception handling.

Requires the latest cosmo GCC with commit https://github.com/ahgamut/superconfigure/commit/c1e5a6e39a7bd48cd2248057374166faae7727ea for SJLJ exceptions (should have the `__USING_SJLJ_EXCEPTIONS__` macro defined by default).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -328,6 +328,7 @@ include third_party/python/BUILD.mk
 include tool/build/BUILD.mk
 include tool/curl/BUILD.mk
 include third_party/qemu/BUILD.mk
+include third_party/libunwind/BUILD.mk
 include examples/BUILD.mk
 include examples/pyapp/BUILD.mk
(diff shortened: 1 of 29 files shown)


jart (approved):
Looks good. Do you need me to add it to `cosmocc`?

jart:
@trungnt2910 Could you fix the test failures?

trungnt2910:
I can reproduce the build failures using the default `cosmocc` package, but not using the version mentioned above.

Seems like an update to the blobs at `third_party/gcc`. Can you please update those, or do you prefer using the ones already present [in one of my builds](https://github.com/trungnt2910/superconfigure/releases/tag/main)?

PR #1056, opened 2024-01-05
Store StrlenDevFd in n

mrdomino:


jart (approved):

pkulchenko (approved):
LGTM, although I was thinking if there is a way to avoid the duplication in `StrlenDevFd`. Maybe `((DevFd() ? sizeof(DevFd) : 0) -1)`? BTW, wouldn't be better to have `(DevFd() ? sizeof(DevFd)-1 : 0)` or am I reading this incorrectly?

mrdomino:
You can't do `sizeof` of a runtime expression, which `DevFd()` generally is.

mrdomino:
Or well, you technically can, but I think it'll decay to `sizeof(char *)` or something, which is not what is wanted here.

PR #1057, opened 2024-01-05
apeinstall.sh cleanup

### LINKED ISSUE #1050: apeinstall.sh installs x86_64 loader on aarch64 linux
On a fresh clone, `ape/apeinstall.sh` leads to `/usr/bin/ape` being `build/bootstrap/ape.elf`, not `build/bootstrap/ape.aarch64`.

mrdomino:
- Use `build/bootstrap/make.com` if it's available, in case the system make is BSD make.

- Handle aarch64, rather than installing the x86_64 loader everywhere (fixes #1050).

- Detect whether the elf or macho loader is wanted and fetch only that.

- Print an error and exit on unsupported architectures.

- Consistently prepend x to uname tests.

- Use `install` if available. (Cleans up the temporary M1 loader with a trap.)


mrdomino:
Tested on M1 Mac, aarch64 and x86_64 Linux.

jart (approved):

PR #1058, opened 2024-01-05
ape loader follows binfmt_misc P flag

mrdomino:
This allows ape to automatically preserve `argv[0]` [as of Linux kernel 5.12][0] if the [binfmt_misc][1] registration contains the P flag.

This also removes may_path_search, which was identical to the literally flag in usage. As a result, FindCommand is subsumed into Commandv.

[0]: https://patchew.org/QEMU/20210222105004.1642234-1-laurent@vivier.eu/
[1]: https://www.kernel.org/doc/html/latest/admin-guide/binfmt-misc.html

### DIFF
--- a/ape/loader.c
+++ b/ape/loader.c
@@ -113,41 +113,44 @@
 #define EXTERN_C
 #endif
 
-#define O_RDONLY           0
-#define PROT_NONE          0
-#define PROT_READ          1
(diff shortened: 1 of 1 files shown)


jart (approved):
Sweet. Thanks!

PR #1052, opened 2024-01-02
add a byte to loader.c

mrdomino:
makes usage justified for 1.10

PR #1060, opened 2024-01-06
Bugfix: ape --help should exit 0

mrdomino:
Just missing a set of parens.


mrdomino:
Snuck the change from #1052 in here as well as a year update on the usage text.

jart (approved):
TY

PR #1061, opened 2024-01-06
Enable argv[0] tests in more places

mrdomino:


mrdomino:
Tested on:

- Linux x86_64:
   - Non-assimilated:
      - Loader does not preserve `argv[0]`: skips `argv[0]` tests, prints log line
      - Loader preserves `argv[0]`: runs `argv[0]` tests
   - Assimilated: runs `argv[0]` tests
- FreeBSD arm64:
   - Non-assimilated, loader does not preserve `argv[0]`: skips `argv[0]` tests, prints log
   - Assimilated: runs `argv[0]` tests
- Xnu arm64, non-assimilated: runs `argv[0]` tests

Curious if this works everywhere else.

jart (approved):

PR #1062, opened 2024-01-06
apeinstall/uninstall.sh can use doas

mrdomino:
Tested on FreeBSD.


jart (approved):
sudo is tired and doas is wired

PR #1066, opened 2024-01-06
Fix download-cosmocc.sh on XnuSilicon

mrdomino:
shasum behaves enough like sha256sum to work in this case.

N.B. the logic to build sha256sum assumes MODE= which is wrong on aarch64.


jart (approved):
Thanks! Could you send this PR to llamafile too?

PR #1070, opened 2024-01-07
Get execfn the apple way

mrdomino:
https://github.com/opensource-apple/dyld/blob/master/src/dyld.cpp#L4789


mrdomino:
Tested on my laptop.

Unfortunately the `GetProgramExecutableName` equivalent will take a bit of doing, since we polyfill `auxv` on Xnu in `crt.S` currently.

mrdomino:
No, this is wrong; that's `AT_EXECFN` of the ape loader itself.

That approach is what we need for `GetProgramExecutableName` though.

PR #1071, opened 2024-01-07
More reliable $prog.ape

mrdomino:
Now it doesn't matter what argv `$prog.ape` is invoked with. We just get our executable path the apple way.

https://github.com/opensource-apple/dyld/blob/master/src/dyld.cpp#L4789

PR also includes passing `AT_FLAGS` from the M1 loader, which lets the `GetProgramExecutableName` tests always detect whether `argv[0]` should be tested from `auxv`.

Also changes `literally` to match `loader.c` in behavior.

### DIFF
--- a/ape/ape-m1.c
+++ b/ape/ape-m1.c
@@ -997,7 +997,7 @@ int main(int argc, char **argv, char **envp) {
        but it will if you say:
            ln -sf /usr/local/bin/ape /opt/cosmos/bin/bash.ape
        and then use #!/opt/cosmos/bin/bash.ape instead. */
-    M->ps.literally = 0;
+    M->ps.literally = 1;
     argc = sp[0];
     argv = (char **)(sp + 1);
   } else if ((M->ps.literally = argc >= 3 && !StrCmp(argv[1], "-"))) {
--- a/ape/ape-m1.c
+++ b/ape/ape-m1.c
@@ -768,7 +768,7 @@ __attribute__((__noreturn__)) static void Spawn(const char *exe, int fd,
 
 static const char *TryElf(struct ApeLoader *M, union ElfEhdrBuf *ebuf,
                           const char *exe, int fd, long *sp, long *auxv,
-                          char *execfn, char literally) {
+                          char *execfn) {
   long i, rc;
   unsigned size;
   struct ElfEhdr *e;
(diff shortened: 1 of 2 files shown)


mrdomino:
I want to rework AccessCommand a bit now that we no longer need to path search in the indirect case.

mrdomino:
Done.

jart (approved):
Nice! I forgot about Apple's auxv string array. Is there anything else good we can pilfer from there?

mrdomino:
Here's what it gives us. Nothing looks immediately useful to me.

```console
% cc apple.c -o apple
% cat apple.c
#include <stdio.h>

int main(int argc, char *argv[], char *envp[])
{
  char **p;
  for (p = envp; *p; ++p) {
  }
  for (++p; *p; ++p) {
    printf("%s\n", *p);
  }
  return 0;
}
% ./apple
executable_path=./apple



ptr_munge=
main_stack=
executable_file=0x1a0100000d,0x1689fd4
dyld_file=0x1a0100000d,0xfffffff0009a8fa
executable_cdhash=1360c6329d477bc5198ee5997a4f5f5ae8166f9e
executable_boothash=b110e39ca11baa63bdb1da60bdc0c2cc15731989
arm64e_abi=os
th_port=
```

mrdomino:
Notably though, we should grab `executable_path` out of their auxv before we reset it in `crt.S`.

ghaerr:
> Is there anything else good we can pilfer from there?

[Here's](https://github.com/apple-oss-distributions/xnu/blob/a325d9c4a84054e40bbe985afedcb50ab80993ea/bsd/kern/kern_exec.c) the actual Apple exec code that creates the auxv string array. The `exec_apple_add_strings` (line 5530) function shows each of the keys that might be interesting to us. 

The `executable_path=` string is handled by calls to the `exec_save_path` function. @mrdomino, you might find all this interesting. I studied it earlier when I was investigating some issues with @jart's original loader and the ELF AUXV array which isn't used by machO binaries.

BTW, Apple's loader source is [here](https://github.com/apple-oss-distributions/xnu/blob/a325d9c4a84054e40bbe985afedcb50ab80993ea/bsd/kern/mach_loader.c#L844-L862), which I found interesting. The link shows the loader code lines which disallow static binaries on M1, obviating the need for two Apple loaders in Cosmopolitan. Note that development and debug kernels allow M1 static binaries!

PR #1072, opened 2024-01-08
Revert apeinstall.sh binfmt flags

mrdomino:
The P flag breaks backwards compatibility with older binaries. The idea is to revert this commit after that break has been resolved.


mrdomino:
This also reverts the F flag. Still not sure how I feel about that one yet. It makes it so that you have to reset the binfmt registration if you want to update it; the instructions already say to run `apeuninstall.sh` to reset the registration, but maybe we should have the script do it automatically. Anyway; not our problem for now.

jart (approved):

PR #1076, opened 2024-01-08
third_party/libcxxabi: Add test suite

trungnt2910:
Added the `libcxxabi` test suite as found in LLVM 17.0.6.

Some tests that do not apply to the current configuration of comsopolitan are not added. These include:
- `backtrace_test`, `forced_unwind*`: Use unwind function unsupported in SjLj mode.
- `noexception*`: Designed to test `libcxxabi` in no exceptions mode.

Some tests are added but not enabled due to bugs specific to GCC or cosmopolitan. These are clearly indicated in the `BUILD.mk` file.


jart (approved):
Thanks!

PR #1085, opened 2024-01-15
libc: Add macros for madv constants

trungnt2910:
This is helpful for code detecting certain `madvise` constants at compile time.

Such macros have already been defined for other constants like `O_*`.


libc/sysv/consts/madv.h
```diff
@@ -33,5 +33,27 @@ extern const unsigned MADV_COLLAPSE;
 #define MADV_SEQUENTIAL 2
 #define MADV_WILLNEED   3
 
+#define MADV_DODUMP          MADV_DODUMP
```
jart, line 36: I don't want them to be detected easily because they'll only work on Linux.

jart (changes requested):

PR #1088, opened 2024-01-17
libc: Fix __cxa_thread_atexit prototype

trungnt2910:
The first parameter should be a function pointer.

Otherwise, this would conflict with the definition provided by `libcxxabi`.


jart (approved):
Thank you!

PR #1089, opened 2024-01-17
libcxx: Add missing implementation source files

trungnt2910:
Added the implementation for `std::bad_any_cast` from upstream `any.cpp` and `std::bad_variant_access` from upstream `variant.cpp`.

This fixes missing `vtable` and `typeinfo` symbols when trying to link code referencing these exception types.


mrdomino (approved):
LGTM

PR #1091, opened 2024-01-18
tool/cosmocc: Properly handle dependency output

trungnt2910:
a2753de contains some regressions, causing `fixupobj` to be inappropriately suppressed when `-MD` or `-MMD` is passed.

This commit reverts most changes by a2753de, and:
- Treats all invocations of the compiler with `-M` and `-MM` as with the `cpp` intent, since these flags imply `-E`.
- Handles the dependency output path specified by `-MF`.
  + This is trivial for `cosmocross` since the script does not throw objects to and from temporary directories.
  + For `cosmocc`, the file names are calculated based on the `-MF` value provided by the user. If this flag is not specified, the script generates the file name based on the output file using GCC rules. Then, before calling the real compilers, an additional `-MF` flag is passed to override the dependency outputs with mangled file names.

### DIFF
--- a/tool/cosmocc/bin/cosmocc
+++ b/tool/cosmocc/bin/cosmocc
@@ -88,7 +88,9 @@ APELINKFLAGS=
 FLAGS_X86_64=
 FLAGS_AARCH64=
 INPUT_FILE_COUNT=0
-OUTPUT_IS_TEXT=0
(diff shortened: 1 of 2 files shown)


jart (approved):
Thank you!

PR #1096, opened 2024-01-20
Add Curve25519 method in help.txt + add new specs file

solisoft:
Not sure how to run tests locally so I tested it directly in the repl


jart (approved):
Thanks for sending this! You can run the Lua tests locally by running `make -j32 o//test/tool/net` which will glob all the `*_test.lua` files and run them in redbean. I've approved the GitHub Action test runner which will run it too. Once that passes, I need one more thing to merge. Please follow the process in https://github.com/jart/cosmopolitan/blob/master/CONTRIBUTING.md and send the email.

PR #1100, opened 2024-02-05
Fix a couple of minor things needed to compile TeX

dfyz:
We're currently trying to make [busytex](https://github.com/busytex/busytex) Actually Portable. Overall it's going pretty smooth, but there are a couple of changes we had to made that I believe should preferably be upstreamed. Amusingly, they were needed to compile not the TeX project proper but the [Lua modules](https://github.com/TeX-Live/texlive-source/tree/trunk/texk/web2c/luatexdir) they bundle with TeX-Live (so this might be useful even for those who don't care about TeX).

The detailed description of what was changed is in the commit message.

### DIFF
--- a/libc/complex.h
+++ b/libc/complex.h
@@ -1,5 +1,5 @@
-#ifndef _COMPLEX_H
-#define _COMPLEX_H
+#ifndef COSMOPOLITAN_LIBC_COMPLEX_H_
+#define COSMOPOLITAN_LIBC_COMPLEX_H_
 COSMOPOLITAN_C_START_
 #if __STDC_VERSION__ + 0 >= 201112 && !defined(__STDC_NO_COMPLEX__)
 
@@ -116,4 +116,4 @@ complex long double cpowl(complex long double, complex long double) libcesque;
 
 #endif /* C11 */
 COSMOPOLITAN_C_END_
-#endif /* _COMPLEX_H */
+#endif /* COSMOPOLITAN_LIBC_COMPLEX_H_ */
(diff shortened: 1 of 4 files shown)


G4Vi (approved):
LGTM! At some point we should add ip multicast tests, but I don't see any involving `struct ip_mreq` either so this is okay for now. 

Before we can merge, we need you to assign copyright to Justine. Can you please send an email to jtunney@gmail.com saying "I intend to assign Justine Tunney the copyright to the changes I contribute to Cosmopolitan"? It only applies to the changes you choose to contribute. It only needs to happen once. If your employer owns everything you do, then you'll need to clear it with them first. Please do that if you need to and get it on the record. Please link to this PR inside your email and note here once you have done so. Thanks!

dfyz:
If you have an idea how what IPv4/IPv6 multicast tests should look like, I can totally take a stab at implementing those (in a different PR or in this one, it's up to you). I see that you have a pretty comprehensive [test suite](https://github.com/jart/cosmopolitan/tree/master/test/libc/sock) already, but the current tests appear to use local addresses (`127.0.0.1` or `::1`), so I'm not sure this will play nice with multicasting.

Regarding the copyright assignment, I sent a letter to Justine simultaneously with opening this PR, but still haven't received any response. Should I try sending it again?

G4Vi:
No need to send the assignment email again. Once I have confirmation from her, I'll merge.

I don't have anything specific in mind for multicast tests, but any could be helpful. The key thing is just to keep in mind that they should run on all platforms cosmopolitan supports, skipping/bailing out so they pass if the current platform doesn't have the needed functionality. I'm guessing the potential difficulty of implementing them may be related to why they don't exist. A separate PR sounds best.

jart:
Confirmed. Please proceed @G4Vi.

PR #1103, opened 2024-02-06
[Redbean] Feature / OnError(status, message) hook

solisoft:
This PR add a new OnError(status, message) hook.

It will allow developer to catch the error and then do whatever it want with it (e.g. display an error page using a framework, sending error to tier service like sentry or appsignal, etc...)

```lua
function OnHttpRequest()
  assert(false)
  Write("ok")
end

function OnError(status, message)
  Write("error " .. status)
  Write("<br>")
  Write(message)
end
```


Frityet:
> This PR add a new OnError(status, message) hook.
> 
> It will allow developer to catch the error and then do whatever it want with it (e.g. display an error page using a framework, sending error to tier service like sentry or appsignal, etc...)
> 
> ```lua
> function OnHttpRequest()
>   assert(false)
>   Write("ok")
> end
> 
> function OnError(status, message)
>   Write("error " .. status)
>   Write("<br>")
>   Write(message)
> end
> ```

isn't this why `pcall` exists?

```lua
local function on_error(err)
   print("oh noes! "..err)
end

function OnHttpRequest()
  xpcall(function()
      assert(false)
      Write("ok")
  end, on_error)
end
```

solisoft:
Yes you're right ... I just find weird to have my code nested in this xpcall but maybe I'm wrong

solisoft:
But it seems it's not really the same thing
In my case this code works as expected
```lua
function OnError(status, message)
  Params = GetParams()
  Params.action = "index"
  RoutePath("/controllers/welcome_controller.lua")
end
```

but using the xpcall it just display a blank page ...

solisoft:
Also if something go wrong on my OnError function, it will fallback to the default error page ... Using the xpcall it will just display a blank file

pkulchenko (commented):
Is the exclamation point in https://github.com/jart/cosmopolitan/pull/1103/files#diff-a937cc5818818785e079eab14ff76fa5db080fe9bd902d1ab21aa546aa79c7d1R3265 intentional? If not, I'd remove the duplicate error message and just left one after the ` if (hasonerror)` check. Maybe something like this (properly edited):

```
   if (hasonerror) {
      lua_getglobal(L, "OnError");
      lua_pushinteger(L, 500);
      lua_pushstring(L, lua_tostring(L, -1));
      if (LuaCallWithTrace(L, 2, 0, NULL) == LUA_OK) {
        return CommitOutput(GetLuaResponse());
      }
  }
      error = ServeErrorWithDetail(
          500, "Internal Server Error",
          ShouldServeCrashReportDetails() ? lua_tostring(L, -1) : NULL);
      lua_pop(L, 1);  // pop error
      return error;
```

@solisoft, I'd also split the definitions changes into its own PR, as it's not related to these changes, right?

solisoft:
Thanks for the feedback @pkulchenko 
I changed the error method to handle all kind of errors and not only 500 ones !

jart (approved):

PR #1104, opened 2024-02-08
Don't include <crypt.h> from <stdlib.h>

dfyz:
POSIX says that [crypt()](https://pubs.opengroup.org/onlinepubs/9699919799/functions/crypt.html) resides in `<unistd.h>`. As far as I know, `crypt()` doesn't have anything to do with the C standard library, so including it unconditionally from `<stdlib.h>` is somewhat strange and breaks the build of the [Lua MD5 implementation](https://github.com/TeX-Live/texlive-source/blob/tags/texlive-2023.0/texk/web2c/luatexdir/luamd5/md5lib.c#L117) bundled with LuaTeX: 

```
/home/dfyz/busytex/source/texlive/texk/web2c/luatexdir/luamd5/md5lib.c:117:12: error: conflicting types for ‘crypt’; have ‘int(lua_State *)’
  117 | static int crypt (lua_State *L) {
      |            ^~~~~
In file included from /home/dfyz/cosmopolitan/cosmocc/include/stdlib.h:17,
                 from /home/dfyz/busytex/source/texlive/texk/web2c/luatexdir/luamd5/md5lib.c:8:
/home/dfyz/cosmopolitan/cosmocc/include/third_party/musl/crypt.h:12:7: note: previous declaration of ‘crypt’ with type ‘char *(const char *, const char *)’
   12 | char *crypt(const char *, const char *);
      |       ^~~~~
```

A simpler test program is just:
```c
#include <stdlib.h>

void crypt() {
}

int main() {
}
```

It builds successfully with both `gcc` (glibc) and `musl-gcc` (musl), but fails to build with `cosmocc` (with an error very similar to the above).
Given that `crypt()` is a very short word and is likely to clash with other real-world programs in the future, I suggest removing this include.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,31 +1,26 @@
-name: build
-
-on:
-  push:
-    branches:
-      - "master"
-      - "flake"
-  pull_request:
-    branches:
-      - "master"
-
-  # run workflow manually from the Actions tab
-  workflow_dispatch:
+name: build-release
+on: workflow_dispatch
 
 jobs:
-  matrix_on_mode:
-    runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        mode: ["", tiny, rel, tinylinux, optlinux]
+  build-release:
+    runs-on: ubuntu-22.04
     steps:
-      - uses: actions/checkout@v3
+      - name: Disable binfmt_misc shenanigans
+        run:  |
+              sudo wget -O /usr/bin/ape https://cosmo.zip/pub/cosmos/bin/ape-$(uname -m).elf
+              sudo chmod +x /usr/bin/ape
+              sudo sh -c "echo ':APE:M::MZqFpD::/usr/bin/ape:' >/proc/sys/fs/binfmt_misc/register"
+              sudo sh -c "echo ':APE-jart:M::jartsr::/usr/bin/ape:' >/proc/sys/fs/binfmt_misc/register"
 
-      - name: support ape bins 1
-        run: sudo cp build/bootstrap/ape.elf /usr/bin/ape
+      - uses: actions/checkout@v4
 
-      - name: support ape bins 2
-        run: sudo sh -c "echo ':APE:M::MZqFpD::/usr/bin/ape:' >/proc/sys/fs/binfmt_misc/register"
+      - name: Build Cosmo toolchain
+        run:  tool/cosmocc/package.sh
 
-      - name: make matrix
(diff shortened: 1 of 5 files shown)


pkulchenko:
I don't mind removing, but it doesn't fully solve the problem, as there is a POSIX crypt() function with a different signature, so it's likely to run into a conflict sooner or later.  I think the issue is with `md5lib.c` and likely needs to be reporter upstream there (as well).

dfyz:
I agree that in general defining `crypt()` (like I did in my shorter example) is just asking for trouble, but in this case it seems to me that `md5lib.c` did everything right:
  * It only includes `std*.h` headers, avoiding anything POSIX at all, so it should compile even on non-POSIX systems.
  * Their `crypt()` implementation is marked as `static` (they expose it to the outside world via an anonymous pointer), so won't clash with anything when linking.

I also agree that the fix I propose here is very brittle: it is enough for any header from the standard library in Cosmopolitan to include `<crypt.h>` transitively for the conflict to re-appear again.

Ultimately, I guess it's up to you to decide if this is worth merging. If not, I will try to talk upstream into renaming `crypt()` to something else, or add some hack to my build files.

dfyz:
Whoops. I made this PR out of my `master` branch, which was a huge mistake (my `master` now has multiple unrelated commits).

Unfortunately, I can't change the source branch of the PR once it was created. Since this change is a one-liner, it can live in my fork without being upstreamed just fine, so I'm going to just close this PR.

G4Vi:
@dfyz I like this change, if you remake this PR I'll merge it. Ideally, no one else would name their function `crypt`, but I agree `crypt` doesn't belong in `stdlib.h` .

G4Vi:
As just mentioned in your fix xnu binaries PR, our discord is probably the best place: https://discord.gg/FwAVVu7eJ4 , otherwise a GitHub discussion or issue on it may get more eyes on it.

I wonder if it can be tested with virtual network interfaces, maybe a test could be made that only runs on root and Linux that sets up virtual network adapters, tests, and cleans them up. If that goes well, it could be expanded to include virtual network interface setup and tear down for other platforms.

jart:
The tests that get run by `make` need to be local and not require root. If you want to add multicast tests, you could try a few things. For starters, I'm happy to merge a manual test. Just name the file so it doesn't match the `%_test.c` filter I think. Then maybe add a `foo.sh` file that shows how to run it. We have a similar .sh for running pledge tests. Another option is I'm happy to accept a github action runner that does this, so long as it's non-flaky.

PR #1106, opened 2024-02-11
Fix invalid XNU binaries generated by `apelink` in some edge cases

dfyz:
I was experimenting with `aarch64`-only APE binaries that don't contain any embedded loaders. You can generate one with:
```
> cat tmp.c
#include <stdio.h>

int main() {
    puts("wow");
}
> aarch64-unknown-cosmo-cc -O2 -o tmp.aarch64 tmp.c
> apelink -o tmp.com tmp.aarch64
```

When I copied the resulting binary to an M1 machine and tried to run it, I got the following confusing error:
```
> ./tmp.com
./tmp.com: line 19: syntax error near unexpected token `fi'
./tmp.com: line 19: `fi'
```

The details of what's going on and how this it was fixed are in the commit message. The generated binaries still don't work because apparently you can't assimilate on Apple Silicon, but the error message is clearer:
```
> ./tmp.com
./tmp.com: this ape program lacks arm64 support
```

While I'm at it, also fix slightly outdated apelink documentation.

### DIFF
--- a/tool/build/apelink.c
+++ b/tool/build/apelink.c
@@ -102,7 +102,7 @@
   "               - 32: freebsd\n"                             \
   "               - 64: netbsd\n"                              \
   "\n"                                                         \
-  "             for example, `-s 0b1110001` may be used to\n"  \
+  "             for example, `-V 0b1110001` may be used to\n"  \
   "             produce ELF binaries that only support the\n"  \
   "             truly open unix systems. in this case when\n"  \
   "             a single input executable is supplied, the\n"  \
@@ -120,8 +120,8 @@
   "             also pass strings in a variety of intuitive\n" \
   "             supported representations. for example, bsd\n" \
   "             will enable freebsd+netbsd+openbsd and that\n" \
-  "             string too is a legal input. the -s flag is\n" \
-  "             also repeatable, e.g. `-s nt -s xnu` to use\n" \
+  "             string too is a legal input. the -V flag is\n" \
+  "             also repeatable, e.g. `-V nt -V xnu` to use\n" \
   "             the union of the two.\n"                       \
   "\n"                                                         \
   "             since the support vector controls the file\n"  \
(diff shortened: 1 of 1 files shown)


ghaerr:
Looks good!

> There is actually a code path guarded by the `gotsome` variable that
   inserts an extra `true` in this case, but the variable was never
   initialized, so in practice this code path didn't activate in my
   tests. 

I'm a little surprised there was no compiler warning of use of an initialized variable during static flow testing... was that perhaps inadvertently turned off?

dfyz:
> I'm a little surprised there was no compiler warning of use of an initialized variable during static flow analysis... was that perhaps inadvertently turned off?

That's a very good question. I just did some research, and I believe the static analysis is *on*, it's just that `gcc` has [a known issue](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=18501) with this specific code pattern (in particular, [this comment](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=18501#c84) is very similar to what we have with `gotsome`).

If I remove all lines that assign to `gotsome`, I get a compilation error, as expected:
```
> make -j o//tool/build/apelink.com.dbg
build/bootstrap/compile.com -V9 -M2048m -P8192  -AMKDEPS -L320 build/bootstrap/mkdeps.com -o o//depend -s -r o// @o//srcs.txt @o//hdrs.txt @o//incs.txt
    108,453⏰    105,779⏳    14,924k   9,032iop build/bootstrap/mkdeps.com -o o//depend -s -r o// @o//srcs.txt @o//hdrs.txt @o//incs.txt
tool/build/apelink.c: In function ‘main’:
tool/build/apelink.c:2130:10: error: ‘gotsome’ may be used uninitialized [-Werror=maybe-uninitialized]
 2130 |       if (!gotsome) {
      |          ^
cc1: all warnings being treated as errors
```

ghaerr:
Thanks for the research. That's good to know that `-Wuninitialized` doesn't always work, even in some fairly simple cases!

dfyz:
@jart Would you be interested in merging this? I don't think creating pure aarch64 binaries is a very common use case, but the fixes themselves, I think, are relatively uncontroversial (especially the uninitialized variable fix).

PR #1107, opened 2024-02-14
Add LIST to khttpmethod.c

mterron:
Currently Redbean's Fetch() API does not seem to support arbitrary HTTP methods. This change will enable the use of LIST as one of the supported HTTP verbs to be able to interact with the Hashicorp Vault HTTP API.

### DIFF
--- a/net/http/khttpmethod.c
+++ b/net/http/khttpmethod.c
@@ -37,4 +37,5 @@ const char kHttpMethod[18][8] = {
     "PATCH",    //
     "REPORT",   //
     "UNLOCK",   //
+    "LIST",     //
 };
--- a/net/http/khttpmethod.c
+++ b/net/http/khttpmethod.c
@@ -18,7 +18,7 @@
 ╚─────────────────────────────────────────────────────────────────────────────*/
 #include "net/http/http.h"
 
-const char kHttpMethod[18][8] = {
+const char kHttpMethod[19][8] = {
     "WUT",      //
     "GET",      //
     "HEAD",     //


pkulchenko:
@mterron, the PR looks good to me. Can you send the copyright assignment to @jart per [these instructions](https://github.com/jart/cosmopolitan/blob/d3ff48c63f89060844dcfa80f0526b2534dfd56f/CONTRIBUTING.md?plain=1#L17)?

pkulchenko (approved):

mterron:
Copyright email sent

mterron:
@jart , is there a reason for not allowing arbitrary http methods?

jart (approved):
Approval for copyright.

net/http/khttpmethod.c
```diff
@@ -37,4 +37,5 @@ const char kHttpMethod[18][8] = {
     "PATCH",    //
     "REPORT",   //
     "UNLOCK",   //
+    "LIST",     //
```
jart, line 40: Please update these lines to have 19 instead of 18.

```
http.h:190:extern const char kHttpMethod[18][8];
khttpmethod.c:21:const char kHttpMethod[18][8] = {
```

jart (changes requested):

net/http/khttpmethod.c
mterron: Is there a reason to hardcode the methods at all? It might be easier to just remove the whole list and then any custom method will just work?

jart: As you wish skeleton jack. I'll push a change shortly that should allow you to use any method, with two caveats. Maximum method length is 8 chars. Upper casing is canonicalized. Any legal HTTP token is allowed.

jart: What's great is that doing methods the generalized way actually shaves 10 nanoseconds off the HTTP parser. I'm glad you challenged me on this issue. Redbean will be better as a result.

mterron: I'll get myself a T-shirt that reads 'I helped shave 10 nanoseconds off Redbean's HTTP parser'

PR #1109, opened 2024-02-21
`*scanf()` fixes to make TeX work

dfyz:
TeX uses `sscanf()` extensively, and there is some breakage when it is compiled with Cosmpolitan, which I'm trying to fix with this PR.

For clarity, the fixes are split into three more-or-less independent commits. The commit messages describe *what* is being fixed, and below is some context on *why* this is being fixed.

The [first commit](https://github.com/jart/cosmopolitan/commit/a57c6067f0e151b3108437d87bf1e6d205e227f5) doesn't fix any actual breakage in TeX; it's a small fix for an edge case in the current code. It could be squashed with the second commit but I think it's cleaner if this change gets a separate commit.

The [second commit](https://github.com/jart/cosmopolitan/commit/e5d877ba4f46b0767333160e19fe308d88636cd9) fixes the mapfile parsing in TeX. It [uses](https://github.com/TeX-Live/texlive-source/blob/bc6f82f7bc771f34bf507951779251e0ec7300bd/texk/web2c/pdftexdir/mapfile.c#L481) `scanf()` with `%f` directive to check if the current config option contains a valid float.

The [third commit](https://github.com/jart/cosmopolitan/commit/a57c6067f0e151b3108437d87bf1e6d205e227f5) fixes a very weird error that looks like this:
```
! pdfTeX error (\pdfsetmatrix): Unrecognized format..
<argument> ...shipout:D \box_use:N \l_shipout_box
                                                  \__shipout_drop_firstpage_...
```

It turns out that TeX [uses](https://github.com/TeX-Live/texlive-source/blob/trunk/texk/web2c/pdftexdir/utils.c#L1415) `scanf()` with `%lf ... %c` directive, where the trailing `%c` is expected to fail (which would mean the last float actually terminates the string).

### DIFF
--- a/test/libc/stdio/fscanf_test.c
+++ b/test/libc/stdio/fscanf_test.c
@@ -1,7 +1,7 @@
 /*-*- mode:c;indent-tabs-mode:nil;c-basic-offset:2;tab-width:8;coding:utf-8 -*-│
 │ vi: set et ft=c ts=2 sts=2 sw=2 fenc=utf-8                               :vi │
 ╞══════════════════════════════════════════════════════════════════════════════╡
-│ Copyright 2022 Justine Alexandra Roberts Tunney                              │
+│ Copyright 2024 Ivan Komarov                                                  │
 │                                                                              │
 │ Permission to use, copy, modify, and/or distribute this software for         │
 │ any purpose with or without fee is hereby granted, provided that the         │
--- a/test/libc/stdio/fscanf_test.c
+++ b/test/libc/stdio/fscanf_test.c
@@ -1,0 +1,33 @@
+/*-*- mode:c;indent-tabs-mode:nil;c-basic-offset:2;tab-width:8;coding:utf-8 -*-│
+│ vi: set et ft=c ts=2 sts=2 sw=2 fenc=utf-8                               :vi │
+╞══════════════════════════════════════════════════════════════════════════════╡
+│ Copyright 2022 Justine Alexandra Roberts Tunney                              │
+│                                                                              │
+│ Permission to use, copy, modify, and/or distribute this software for         │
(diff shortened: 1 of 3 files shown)


test/libc/stdio/fscanf_test.c
```diff
@@ -0,0 +1,33 @@
+/*-*- mode:c;indent-tabs-mode:nil;c-basic-offset:2;tab-width:8;coding:utf-8 -*-│
+│ vi: set et ft=c ts=2 sts=2 sw=2 fenc=utf-8                               :vi │
+╞══════════════════════════════════════════════════════════════════════════════╡
+│ Copyright 2022 Justine Alexandra Roberts Tunney                              │
```
G4Vi, line 4: You can own the copyright for your own test files, feel free to set the year and your own name when creating a new test file.

G4Vi (changes requested):
Code changes look good. Thanks for adding thorough tests! Just "Requesting Changes" so you can own your own test file.

test/libc/stdio/fscanf_test.c
dfyz: Thanks, I just pushed a fix! I thought of that when I was making this PR but decided it would be weird to claim the copyright for, like, 10 lines of code.

jart: Not weird at all. It leaves a record that you were here and that you helped.

You can also own non-test code if you're contributing a feature. For example, let's say you discovered a faster better way to implement `log10()`. You can rewrite `log10.c` from scratch with your discovery, put your name on the top, and add a `__notice()` directive so that your name will even be embedded inside every binary that uses `log10()`. Stuff like this is obviously an exception to the rule, so just ask me beforehand if you want to do it.

PR #1124, opened 2024-03-12
Update README.md

jperon:
Remove `.com` extension in Readme.


jart:
Thanks for bringing this to my attention. This has been fixed in 9a10adac35.

PR #1129, opened 2024-03-23
change 15.6 to 10.15.6 under macos support

foxt:
resolves ambiguity because there is likely to be a macOS 15.6 some point in the next year (or at least a 15.4), when I assume what was actually meant is 10.15.6 (which actually was released in 2018).


ghaerr:
Well, if you're going to fix things, perhaps the Apple operating system reference ought to be changed to its proper name 'macOS' also?

jart:
We made a similar clarifying change like this in the llamafile readme, where I decided to use the Darwin version number reported by `uname -a`. I'm going to make that change to the readme and close this PR. Thank you for sending it though.

PR #1134, opened 2024-04-03
Remove .com logic from GetProgramExecutableName

mrdomino:
I can't imagine any way that appending `".com"` would still be relevant from `GetProgramExecutableName`’s perspective. However, my lack of imagination may be mere ignorance; in that case, I’d suggest still removing the `com` parameter and having the code always take the `com == 1` path.


G4Vi (approved):
It seems unnecessary to me too. Seeing no objections in the last few weeks, let's merge this.

PR #1135, opened 2024-04-03
Move store after __asan_unpoison in __zipos_alloc

mrdomino:
Previously, the atomic store looked like it was happening while the struct's memory was still poisoned. I was unable to observe any issues with this, but this change seems to make the code more obviously correct (at the cost of a redundant atomic store to zeroed space in case the map needed to be extended.)


G4Vi (approved):
Seems reasonable to me, approving!

PR #1140, opened 2024-04-09
[redbean] Add UuidV4 method 

solisoft:
This PR add the UuidV4 Lua method.


tool/net/lfuncs.c
```diff
@@ -829,6 +829,52 @@ int LuaVisualizeControlCodes(lua_State *L) {
   return LuaCoder(L, VisualizeControlCodes);
 }
 
+int LuaUuidV4(lua_State *L) {
+  char v[] = {'0', '1', '2', '3', '4', '5', '6', '7',
+              '8', '9', 'a', 'b', 'c', 'd', 'e', 'f'};
+  static char uuid_str[37] = {0};
+
+  uuid_str[0] = v[rand() % 16];
```
mrdomino, line 837: You probably want to be using `random()` instead of `rand()`, since the latter gives very low-quality randomness (and if the XNU manpage applies to cosmo's implementation, the low dozen bits will just cycle through a pattern rather than being random.)

I think what is actually wanted here is `arc4random` or one of its variants, but those aren't (yet) in cosmopolitan. I'd suggest using `random` for now.

mrdomino, line 837: You probably want to be using `random()` instead of `rand()`, since the latter gives very low-quality randomness (and if the XNU manpage applies to cosmo's implementation, the low dozen bits will just cycle through a pattern rather than being random.)

I think what is actually wanted here is `arc4random` or one of its variants, but those aren't (yet) in cosmopolitan. I'd suggest using `random() % 16` for now.

tool/net/lfuncs.c
```diff
@@ -829,6 +829,52 @@ int LuaVisualizeControlCodes(lua_State *L) {
   return LuaCoder(L, VisualizeControlCodes);
 }
 
+int LuaUuidV4(lua_State *L) {
```
mrdomino, line 832: As discussed, the explicitly unrolled loop here makes the code quite a bit uglier. I would really question whether we're deriving worthwhile benefit from the performance differences at this level.

tool/net/help.txt
```diff
@@ -1086,7 +1089,7 @@ FUNCTIONS
             - keepalive (default = false): configures each request to keep the
               connection open (unless closed by the server) and reuse for the
               next request to the same host. This option is disabled when SSL
-              connection is used.
+              connection is used.
```
mrdomino, line 1092: Trailing space?

test/tool/net/uuidv4_test.lua
```diff
@@ -0,0 +1,23 @@
+-- Copyright 2022 Justine Alexandra Roberts Tunney
```
mrdomino, line 1: Copyright 2024?

test/tool/net/uuidv4_test.lua
```diff
@@ -0,0 +1,23 @@
...
+-- AUTHOR BE LIABLE FOR ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL
+-- DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR
+-- PROFITS, WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER
+-- TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR
+-- PERFORMANCE OF THIS SOFTWARE.
+
+assert(#UuidV4() == 36)
+assert(string.sub(UuidV4(), 9, 9) == "-")
+assert(string.sub(UuidV4(), 14, 14) == "-")
+assert(string.sub(UuidV4(), 15, 15) == "4")
```
mrdomino, line 19: I don't love having these tests run off of a single call to `UuidV4`. If there were an error, they would become flaky.

Maybe have the tests (at least the ones that check numeric values) run over a few iterations?

mrdomino (changes requested):
Much of this is as discussed in Discord, as well as a couple minor nits.

Looks good overall.

I'm not totally opposed to checking in the unrolled loop, I just want to push back once more, particularly in light of needing to use `random()` (which might change the performance margins already.)

tool/net/lfuncs.c
mrdomino: If `random` winds up being expensive, it would also be possible to extract every bit from it to save on the number of calls, but you'd have to have an iteration bound based on `sizeof(long)`…

mrdomino: Actually strike that, I guess maybe `random` isn't seeded by default.

You could just use `getrandom`; that's what `GetRandomBytes` is using.

test/tool/net/uuidv4_test.lua
solisoft: I just copied an existing one :)

tool/net/help.txt
solisoft: removed

test/tool/net/uuidv4_test.lua
mrdomino: Yeah no problem there, I just usually update the copyright year whenever I'm making a new file

test/tool/net/uuidv4_test.lua
solisoft: I'm running one million tests now

test/tool/net/uuidv4_test.lua
```diff
@@ -12,12 +12,14 @@
...
-
-assert(#UuidV4() == 36)
-assert(string.sub(UuidV4(), 9, 9) == "-")
-assert(string.sub(UuidV4(), 14, 14) == "-")
-assert(string.sub(UuidV4(), 15, 15) == "4")
-assert(string.sub(UuidV4(), 19, 19) == "-")
-y = string.sub(UuidV4(), 20, 20)
-assert(y == "8" or y == "9" or y == "a" or y == "b")
-assert(string.sub(UuidV4(), 24, 24) == "-")
+for i = 1, 1000000 do
```
mrdomino, line 15: haha, I was thinking more like 16 or 32, but I do not object :)

tool/net/lfuncs.c
```diff
@@ -829,6 +829,52 @@ int LuaVisualizeControlCodes(lua_State *L) {
...
+  char v[] = {'0', '1', '2', '3', '4', '5', '6', '7',
+              '8', '9', 'a', 'b', 'c', 'd', 'e', 'f'};
+  static char uuid_str[37] = {0};
```
mrdomino, line 835: Why is this static? Seems like the function should be reentrant.

mrdomino (commented):
Looks good, just update the copyright year and address the `static` comment and I'm happy with this.

test/tool/net/uuidv4_test.lua
solisoft: fixed

mrdomino (approved):
LGTM

Issue #1138, opened 2024-04-04
[windows] fork() ReadFile_SIZE_CHECK() failed with win32 code 0


G4Vi:
`forktest.c`:
```c
#include <unistd.h>
#include <stdio.h>

int main(void) {
    printf("before forking\n");
    pid_t pid = fork();
    if (pid == -1) {
        printf("fork failed\n");
        return 1;
    }
    else if (pid == 0) {
        printf("child\n");
        return 0;
    }
    printf("parent\n");
    return 0;
}
```
`cosmocc/bin/x86_64-unknown-cosmo-cc -o forktest.com forktest.c`

```
C:\Users\username\Downloads\forktest.com
before forking
fork() ReadFile_SIZE_CHECK() failed with win32 error 0
```

Then it appears to hang forever.

G4Vi:
The problem is intermittent, every once and a while it succeeds.

PR #1141, opened 2024-04-11
Fix fork locking on win32

G4Vi:
Fixes https://github.com/jart/cosmopolitan/issues/1138

- `__enable_threads` / set `__threaded` in `__proc_setup` as threads are required for win32 subprocess management
- move mmi/fds locking out of pthread_atfork.c into fork.c so it's done anytime `__threaded` is set instead of being dependent of pthreads
- explicitly yoink `_pthread_onfork_prepare`, `_pthread_onfork_parent`, and `_pthread_onfork_child` in pthread_create.c so they are linked in in-case they are separated from `_pthread_atfork`

Big Thanks to @dfyz for help with locating the issue, testing, and devising a fix!

### DIFF
--- a/libc/intrin/pthreadlock.c
+++ b/libc/intrin/pthreadlock.c
@@ -20,6 +20,10 @@
 
 pthread_spinlock_t _pthread_lock_obj;
 
+void _pthread_init(void) {
+  (void)pthread_spin_init(&_pthread_lock_obj, 0);
+}
+
 void _pthread_lock(void) {
   pthread_spin_lock(&_pthread_lock_obj);
 }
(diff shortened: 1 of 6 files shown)


G4Vi:
Drafting this as it's incomplete, on win32 without pthreads it leaves the child unable to open files.

G4Vi:
Fixed the win32 non-pthread child not initializing the locks. Also decided to with `__threaded` always `_pthread_lock/_pthread_unlock` as they are also used by `libc/calls/sig.c` and `libc/proc/execve-nt.greg.c` regardless of pthreads being linked.

G4Vi:
I'm not really sure other than it's critical to the win32 fork implementation and is used to wait on child processes. `__proc_lock` calls `__proc_setup` which spawns a thread https://github.com/jart/cosmopolitan/blob/master/libc/proc/proc.c#L232-L236

mrdomino (approved):
LGTM

mrdomino:
The conflict came from clang-format line-breaking the short if statement; should be trivial to resolve.

PR #1142, opened 2024-04-11
remove zlib namespacing

ahgamut:
it seems like we can handle the zlib dependency fine both inside and outside the monorepo without having to do any sort of namespacing.

I remember we added the namespacing for the zlib functions because it was clashing with some builds (but I don't remember the exact reason :disappointed: ). Do we still have those clashing builds now? The namespaced symbols are now causing an issue with https://github.com/ahgamut/superconfigure/pull/29


jart (approved):
I say let's give it a try.

PR #1148, opened 2024-04-17
Various paginate improvements

G4Vi:
* Support `C:\Windows\System32\more.com` so paging should work out of the box on win32
* Add `__paginate_file` to paginate an existing file

`paginate.c` was moved from `libc/proc` to `libc/x` mainly to get `utf16to8`. However, it seems to fit as pagination is not standard, but is implemented in a variety of codebases.

This patch will be helpful for getting `perldoc` working again in APPerl. Running a pager via a shell broke when cosmo stopped converting cli arguments to native file paths.

### DIFF
--- a/libc/x/paginate.c
+++ b/libc/x/paginate.c
@@ -70,17 +70,17 @@ void __paginate(int fd, const char *s) {
   char *args[3] = {0};
   char tmppath[] = "/tmp/paginate.XXXXXX";
   char progpath[PATH_MAX];
+  bool done;
   if ((args[0] = get_pagerpath(progpath, sizeof(progpath)))) {
     if ((tfd = mkstemp(tmppath)) != -1) {
-      // defer((void*)unlink, tmppath);
       write(tfd, s, strlen(s));
       close(tfd);
       args[1] = tmppath;
-      if (run_pager(args)) {
-        unlink(tmppath);
+      done = run_pager(args);
+      unlink(tmppath);
+      if (done) {
         return;
       }
-      unlink(tmppath);
     }
   }
   write(fd, s, strlen(s));
--- a/libc/x/paginate.c
+++ b/libc/x/paginate.c
@@ -16,7 +16,9 @@
 │ TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR             │
 │ PERFORMANCE OF THIS SOFTWARE.                                                │
 ╚─────────────────────────────────────────────────────────────────────────────*/
+#include "libc/dce.h"
 #include "libc/calls/calls.h"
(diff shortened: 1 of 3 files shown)


libc/x/paginate.c
```diff
@@ -17,42 +17,93 @@
 │ PERFORMANCE OF THIS SOFTWARE.                                                │
```
mrdomino, line 17: Why move this from `libc/proc` to `libc/x`?

mrdomino (approved):
LGTM, just one question.

libc/x/paginate.c
G4Vi: > paginate.c was moved from libc/proc to libc/x mainly to get utf16to8. However, it seems to fit as pagination is not standard, but is implemented in a variety of codebases.

I think it needs to move to get access to `utf16to8` without `_weaken`. The other justification refers to the description of libc/x in it's `BUILD.MK`.

mrdomino: Per discord, suggesting keeping this in `libc/proc` and using `_weaken`.

G4Vi: I moved it back to `libc/proc` and switched it to use `_weaken` in 38d0728, however the `__static_yoink`  doesn't work from there, in order to get `utf16to8` linked in I had to  `__static_yoink("__utf16to8");` directly from my `__paginate` using program.

Any suggestions of how I could proceed to avoid requiring explicit linking of `__utf16to8` ? I could wrap `tprecode16to8` and use `strlen16` and `malloc` instead, but at that point we're reinventing `utf16to8` inside of `libc/proc`.

mrdomino: ty for the poke - I had missed this.

I do not have a smart answer at the moment. It seems like the three dumb options are:

1. push the dependency off onto the user as you indicate;
2. put the yoink in the file in `libc/proc` thus creating a dependency on that function out of `libc/x` from `libc/proc`; or 
3. reimplement it in `libc/proc`.

My knowledge of what `__static_yoink` is doing has paged out so I'm not sure if there are any problems created by option 2. If you want this merged quickly and are okay with making fixes and refinements later, I would say go with option 1.

ahgamut: I think I messed up by focusing on the move to `libc/x`. 

Looking at the root Makefile, we see that `LIBC_PROC` is declared before `LIBC_X`, and IIRC the parts with fewer dependencies (`LIBC_PROC` in this case) are declared before `LIBC_X`.

So having something in `libc/proc` depend on `libc/x` would introduce some circular issue? Perhaps the easiest way for now is to keep it in `LIBC_X`.

mrdomino (approved):
LGTM

PR #1149, opened 2024-04-22
Update GetRemoteAddr documentation to clarify that it may return `nil`.

pkulchenko:


mrdomino (approved):
LGTM

It is probably worth mentioning that the `nil` case can only happen if the client IP is trusted and contains an erroneous `X-Forwarded-For`.

As a matter of fact, the documentation for this method could probably be improved overall. `IsTrustedIp` seems to encompass more cases than just `IsPrivateIp` or `IsLoopbackIp`, so there's probably a simple sentence or two that describes what `GetRemoteAddr` currently does including the nil case.

pkulchenko:
Right, but it doesn't even need to be erroneous (in the true sense). I think the last case where we ran into this was IPv6 address supplied as part of the `X-Forwarded-For` header.

I wasn't sure how detailed we wanted the documentation to be, but thought it was worth mentioning that `nil` may be returned. Do you want me to add the IPv6 case?

mrdomino:
This is a part of the system that I'm not too familiar with, so I'll defer to you on this one.

Perhaps something like "returns the client IP, unless the client IP is trusted and includes an X-Forwarded-For. In the latter case, if we fail to understand the forwarded IP (e.g. because it is misformatted, or an IPv6 address), this method returns nil."

pkulchenko:
Updated the description; there are more cases covered in IsTrustedIp description, so I just referred to that.

mrdomino:
Great, looks good.

Issue #1130, opened 2024-03-31
Ctrl+C has unusual behaviour while Python waits on `input()`


cloudrac3r:
I ran this file in these builds of Python:

```
input("hello! ")

print("looping")
a = 0
for i in range(200_000_000):
    a += 1

print("goodbye")
```

| Build                                                              | Ctrl+C in loop         | Ctrl+C in input()                                                                               |
|---------------------------------------------------------------------|------------------------|-------------------------------------------------------------------------------------------------|
| Debian Python 3.11.2<br>*Expected behaviour*                           | Print ^C and traceback | Print ^C and traceback                                                                          |
| Cosmopolitan A.P.P. 3.6.14+<br>`make o//third_party/python/python.com` | Print ^C and traceback | Press once: No effect<br>Press twice: Traceback                                                    |
| cosmo.zip Python 3.11.2<br>https://cosmo.zip/pub/cosmos/bin/python     | Print ^C and traceback | Press once: No effect<br>Press repeatedly: No effect<br>Press Ctrl+C and Enter: Print ^C and traceback |

I tested on Linux and macOS and it seems to be the same in both.

I would expect Ctrl+C to traceback straight away regardless of whether the program is running or waiting for input. Is it possible to change this behaviour in Cosmopolitan Python?

cloudrac3r:
I found out how to change this behaviour in A.P.P. Since I am fairly intimidated by the codebase and not sure what to do, I thought I'd document the process I took in this comment to help other newcomers build up enough confidence and learn the skills to make their own changes. If you don't want to read it, skip to the bottom of the comment.

----

Here's the thought process I went through:

Since the [A.P.P. source tree](https://github.com/jart/cosmopolitan/tree/master/third_party/python) is different to the [Cosmos Python source tree](https://github.com/ahgamut/cpython/tree/pypack1), they probably have different code for Ctrl+C handling. My findings for one build will probably not apply to the other build, so I should pick one build to investigate and stick to it. I'll pick A.P.P. so that I only have to investigate code that's inside the monorepo. This means I can search the _entire_ codebase at once with [ag](https://github.com/ggreer/the_silver_searcher), and if ag shows no results, it's not there.

[...]

PR #1152, opened 2024-04-23
Let signals interrupt fgets unless SA_RESTART set

cloudrac3r:
See investigation in #1130.

fgets internally calls readv. readv is a `@restartable` function that understands SA_RESTART. If SA_RESTART is set, readv already handles restarting the system call and eventually the string is transparently returned to the fgets caller.

When SA_RESTART is not set, -1 EINTR should bubble up to the fgets caller. However, until this commit, fgets itself would detect EINTR and keep retrying until it read an entire line.

This commit fixes this behaviour so that fgets understands SA_RESTART.

I hereby waive copyright to this commit and place it in public domain.


cloudrac3r:
I have now done the copyright assignment process, and have checked in my test file with a copyright notice under my own name as encouraged.

I haven't written C since university so I hope my test file is doing the right thing. I will welcome any suggestions.

I did clang-format and valgrind, both tools are happy with my code. I also made sure that the test only passes when fgets_unlocked.c has also been changed. If you undo the change in fgets_unlocked.c, you'll see the test fail.

I was not able to figure out how to run the entire test suite and couldn't find any instructions. The output of `make test` looked like I need to have 7 computers with the correct operating systems already installed, which I'm not able to do. In theory my change should be good across platforms, so fingers crossed.

G4Vi:
You can run the test suite just by making items under `o/$MODE/test` for example `make MODE= o//test/libc/stdio` builds and runs the stdio tests or `make MODE= o//test/libc/stdio/fgets_interrupt_test.runs` builds and runs just your new test.

Unfortunately it looks like your test needs tweaks as it fails on my Linux system and on the CI runs above:
```
sample@frontierjustice:~/repos/cosmopolitan$ make MODE= o//test/libc/stdio
error:test/libc/stdio/fgets_interrupt_test.c:104: assertStringNotEquals() in fgets_eintr_testThatTheSignalInterruptsFgets() on frontierjustice pid 32939 tid 32939
        buf
                need "Hello world!" =
                 got "Hello world!"
        No error information
        o//test/libc/stdio/fgets_interrupt_test
error:test/libc/stdio/fgets_interrupt_test.c:105: fgets_eintr_testThatTheSignalInterruptsFgets() on frontierjustice pid 32939 tid 32939
        EXPECT_EQ(EINTR, errno)
                need 4 (or 0x04 or '\4' or EINTR) =
                 got 0 (or 0x0 or '\0')
        No error information
        o//test/libc/stdio/fgets_interrupt_test @ frontierjustice
2 / 9 tests failed

`make MODE= -j16 o//test/libc/stdio/fgets_interrupt_test.runs` exited with 2:
o//test/libc/stdio/fgets_interrupt_test
consumed 817µs wall time
ballooned to 596kb in size
needed 675us cpu (0% kernel)
caused 82 page faults (100% memcpy)
4 context switches (100% consensual)

make: *** [build/rules.mk:133: o//test/libc/stdio/fgets_interrupt_test.runs] Error 2
```

cloudrac3r:
Looks like the `o//test/libc/stdio/fgets_interrupt_test` executable succeeds on my build server but fails on my laptop. Seems like adding `--strace` on my laptop can sometimes cause it to succeed. I think there might be a race condition in the test. I'll see if I can figure it out.

cloudrac3r:
I couldn't figure out what was causing the race condition. Setting sched_setaffinity to a single core is the only way I could find to fix it reliably. Sorry for resorting to an inelegant hack.

test/libc/stdio/fgets_interrupt_test.c
```diff
@@ -0,0 +1,138 @@
...
+│ PROFITS, WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER               │
+│ TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR             │
+│ PERFORMANCE OF THIS SOFTWARE.                                                │
+╚─────────────────────────────────────────────────────────────────────────────*/
+#include "libc/isystem/errno.h"
+#include "libc/isystem/sched.h"
+#include "libc/isystem/signal.h"
+#include "libc/isystem/stddef.h"
+#include "libc/isystem/unistd.h"
+#include "libc/stdio/stdio.h"
```
G4Vi, line 24:
```suggestion
#include <errno.h>
#include <sched.h>
#include <signal.h>
#include <stddef.h>
#include <stdio.h>
#include <unistd.h>
```

Formatting: I don't see a precedent for including from `isystem` in tests. Using either the direct paths (such as `libc/stdio/stdio.h`) or using the standard include format `<stdio.h>`  is both allowed, I would just stay consistent within the file.

test/libc/stdio/fgets_interrupt_test.c
```diff
@@ -0,0 +1,138 @@
...
+  // Tidy up
+  fclose(stream);
+}
+
+// -=- these set up the tests --------------------------------------------------
+
+void SetUpOnce(void) {
+  cpu_set_t set;
+  CPU_ZERO(&set);
+  CPU_SET(1, &set);
```
G4Vi, line 77:
```suggestion
  CPU_SET(0, &set);
```
I had to set it 0 to run on my single core Windows 10 vm.

test/libc/stdio/fgets_interrupt_test.c
```diff
@@ -0,0 +1,138 @@
...
+    perror("pipe");
+    _exit(1);
+  }
+}
+
+// -=- these are the tests -----------------------------------------------------
+
+TEST(fgets_eintr, testThatFgetsReadsFromPipeNormally) {
+  setup_signal_and_pipe(0);  // 0 = no SA_RESTART
+  ASSERT_NE(-1, (pid = fork()));
```
G4Vi, line 105: Consider rather than using `fork`,  `exit`, etc directly  use `SPAWN`, `PARENT`, and `WAIT` from `#include "libc/testlib/subprocess.h"` as a standard way to `fork`  and wait on children hooked in with the test library. See [`test/libc/sock/recvfrom_test.c` ](https://github.com/jart/cosmopolitan/blob/master/test/libc/sock/recvfrom_test.c) for an example.

On `testThatTheSignalInterruptsFgets` have the child `EXPECT_SYS(0, 0, sigignore(SIGPIPE));` so that it exits successfully.

test/libc/stdio/fgets_interrupt_test.c
```diff
@@ -0,0 +1,138 @@
...
+  if (!pid) {
+    write_pipe(1);  // 1 = signal
+    _exit(0);
+  }
+  read_pipe();
+  EXPECT_STREQ(MY_TEST_STRING_1 MY_TEST_STRING_2, buf);
+  EXPECT_NE(EINTR, errno);
+  EXPECT_EQ(1, got_sigusr1);
+}
```
G4Vi, line 138: Unfortunately neither of these tests run successfully on Windows. However, I think it's unrelated to this work. https://github.com/jart/cosmopolitan/issues/1160

I would `#include "libc/dce.h"` and then at the start of these tests add something like:

```c
  // [Windows] process terminates when receiving signal during readv #1160
  if (IsWindows())
    return;
```

to bail out successfully noting why it's disabled on Windows.

G4Vi (changes requested):
Tests look good overall, thank you! No worries about `sched_setaffinity`, I'm not surprised something like it is needed to make this reproducible.

jart (approved):
Approved for copyright and for the change itself. Thank you!

test/libc/stdio/fgets_interrupt_test.c
jart: Contributors are expected to make their best effort to write code that works across platforms, however they're only required to pass GitHub Actions to get merged. I'm happy to add `if (IsWindows())` statements myself when I run this across my test fleet.

test/libc/stdio/fgets_interrupt_test.c
jart: See `test/posix/...` now that normal includes are possible I'm more encouraging of them for c standard library headers.

PR #1154, opened 2024-04-23
Apply clang-format update to repo

mrdomino:
Commit bc6c1836 introduced a bunch of discrepancies between what files
look like in the repo and what clang-format says they should look like.
However, there were already a few discrepancies prior to that. Most of
these discrepancies seemed to be unintentional, but a few of them were
load-bearing (e.g., a `#include` that violated header ordering needing
something to have been `#define`d by a 'later' `#include`.)

I opted to take what I hope is a relatively smooth-brained approach: I
reverted the .clang-format change, ran clang-format on the whole repo,
reapplied the .clang-format change, reran clang-format again, and then
reverted the commit that contained the first run. Thus the full effect
of this PR should only be to apply the changed formatting rules to the
repo, and from skimming the results, this seems to be the case.

My work can be checked by applying the short, manual commits, and then
rerunning the command listed in the autogenerated commits (those whose
messages I have prefixed `auto:`) and seeing if your results agree.

It might be that the other diffs should be fixed at some point but I'm
leaving that aside for now.


jart (approved):
My name is Justine Tunney and I approve this high impact change.

PR #1163, opened 2024-04-30
add /dev/(u)random on nt

### LINKED ISSUE #1001: is it possible to polyfill /dev/urandom on windows
Some applications **and libraries** (just search /dev/urandom in github) are HARD COEDED to use /dev/urandom for random source, but currently cosmopolitan doesn't provide random device in nt platform

G4Vi:
Resolves #1001

Uses `ProcessPrng` as `getrandom` does.

When feature detection fails software often falls back to `/dev/urandom`, this is a problem as development with cosmo is often done on Linux where tests will pass using `/dev/urandom`, but then won't work on Windows. While it's easy enough to patch and using `getrandom` is better, this makes building correct ports with cosmopolitan/cosmocc easier.

### DIFF
--- a/libc/calls/ioctl.c
+++ b/libc/calls/ioctl.c
@@ -24,7 +24,6 @@
 #include "libc/calls/termios.h"
 #include "libc/dce.h"
 #include "libc/errno.h"
-#include "libc/serialize.h"
 #include "libc/intrin/cmpxchg.h"
 #include "libc/intrin/strace.internal.h"
 #include "libc/intrin/weaken.h"
@@ -42,6 +41,7 @@
 #include "libc/nt/winsock.h"
 #include "libc/runtime/runtime.h"
 #include "libc/runtime/stack.h"
+#include "libc/serialize.h"
 #include "libc/sock/internal.h"
 #include "libc/sock/struct/ifconf.h"
 #include "libc/sock/struct/ifreq.h"
@@ -66,7 +66,7 @@ static struct HostAdapterInfoNode {
   struct sockaddr netmask;
   struct sockaddr broadcast;
   short flags;
-} *__hostInfo;
+} * __hostInfo;
 
 static int ioctl_default(int fd, unsigned long request, void *arg) {
   int rc;
@@ -106,7 +106,8 @@ static int ioctl_fionread(int fd, uint32_t *arg) {
       int bytes = CountConsoleInputBytes();
       *arg = MAX(0, bytes);
       return 0;
-    } else if (g_fds.p[fd].kind == kFdDevNull) {
+    } else if (g_fds.p[fd].kind == kFdDevNull ||
+               g_fds.p[fd].kind == kFdDevRandom) {
       *arg = 1;
       return 0;
     } else if (GetFileType(handle) == kNtFileTypePipe) {
--- a/libc/calls/ioctl.c
+++ b/libc/calls/ioctl.c
@@ -106,10 +106,10 @@ static int ioctl_fionread(int fd, uint32_t *arg) {
       int bytes = CountConsoleInputBytes();
       *arg = MAX(0, bytes);
       return 0;
-    } else if (g_fds.p[fd].kind == kFdDevNull ||
-               g_fds.p[fd].kind == kFdDevRandom) {
-      *arg = 1;
-      return 0;
+    } else if (g_fds.p[fd].kind == kFdDevNull) {
+      return enotty();
+    } else if (g_fds.p[fd].kind == kFdDevRandom) {
+      return einval();
     } else if (GetFileType(handle) == kNtFileTypePipe) {
       uint32_t avail;
       if (PeekNamedPipe(handle, 0, 0, 0, &avail, 0)) {
--- a/libc/calls/write-nt.c
+++ b/libc/calls/write-nt.c
@@ -52,6 +52,11 @@ static textwindows ssize_t sys_write_nt_impl(int fd, void *data, size_t size,
   struct Fd *f = g_fds.p + fd;
   bool isconsole = f->kind == kFdConsole;
 
+  // not implemented
+  if (f->kind == kFdDevRandom) {
+    return enosys();
+  }
+
   // determine win32 handle for writing
   int64_t handle = f->handle;
   if (isconsole && _weaken(GetConsoleOutputHandle)) {
--- a/libc/calls/write-nt.c
+++ b/libc/calls/write-nt.c
@@ -52,9 +52,9 @@ static textwindows ssize_t sys_write_nt_impl(int fd, void *data, size_t size,
   struct Fd *f = g_fds.p + fd;
   bool isconsole = f->kind == kFdConsole;
 
-  // not implemented
+  // not implemented, XNU returns eperm();
   if (f->kind == kFdDevRandom) {
-    return enosys();
+    return eperm();
   }
 
   // determine win32 handle for writing
--- a/test/libc/calls/specialfile_test.c
+++ b/test/libc/calls/specialfile_test.c
@@ -27,9 +27,12 @@
 
 int pipefd[2];
 int stdoutBack;
+int allowMask;
 
 void SetUpOnce(void) {
   testlib_enable_tmp_setup_teardown();
+  // qemu-aarch64 defines o_largefile wrong
+  allowMask = ~(O_LARGEFILE | 0x00008000);
 }
 
 void CaptureStdout(void) {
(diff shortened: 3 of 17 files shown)


ahgamut:
Are there examples of software that have a config check for `/dev/urandom` but not for `getrandom`?

G4Vi:
> Are there examples of software that have a config check for `/dev/urandom` but not for `getrandom`?

It doesn't do a config check, but Perl defaults to `/dev/urandom` and doesn't use `getrandom` at all: https://github.com/Perl/perl5/blob/04d2040ce4136af8389d95e16c75aedcbfa4f9ec/util.c#L4736-L4750

WAMR does a platform config check, rather than one on the symbol:
https://github.com/bytecodealliance/wasm-micro-runtime/blob/4abf288c9431224dfef410bc972feac6dbdb2bc6/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/ssp_config.h#L27-L35
Before I added `defined(__COSMOPOLITAN__)`, it fell back to `/dev/urandom`:  https://github.com/bytecodealliance/wasm-micro-runtime/blob/4abf288c9431224dfef410bc972feac6dbdb2bc6/core/iwasm/libraries/libc-wasi/sandboxed-system-primitives/src/random.c#L69-L80

mrdomino:
Conceptually I like this change, it definitely is the case that build systems will often check for `/dev/urandom` on the host 🙄 rather than the target machine to decide whether to make binaries that try to use it.

I have not looked closely at the PR yet and do not currently have time to; I will try to later on, if you haven't heard back from me by Friday and still need a review please ping me.

One thing is that you may want to use `CryptGenRandom` on Windows - I don't know if `getrandom` uses that or not yet. It probably should.

G4Vi:
> One thing is that you may want to use `CryptGenRandom` on Windows - I don't know if `getrandom` uses that or not yet. It probably should.

Did you mean `BCryptGenRandom`?, `CryptGenRandom` is deprecated. `CryptGenRandom` and `RtlGenRandom` are just wrappers around `ProcessPrng`, and I think `BCryptGenRandom` and `ProcessPrng` share a lot of the same infrastructure.

https://download.microsoft.com/download/1/c/9/1c9813b8-089c-4fef-b2ad-ad80e79403ba/Whitepaper%20-%20The%20Windows%2010%20random%20number%20generation%20infrastructure.pdf

I think `ProcessPrng` is sufficient as:

 `go` upgraded from `CryptGenRandom` to `RtlGenRandom` https://go-review.googlesource.com/c/go/+/210057 and then on to  `ProcessPrng` https://go-review.googlesource.com/c/go/+/536235.

BoringSSL only uses `BCryptGenRandom` on UWP as `ProcessPrng` is unavailable there and it may access resources unavailable in the Chromium sandbox.  https://boringssl.googlesource.com/boringssl/+/master/crypto/rand_extra/windows.c

For `getrandom` on Rust it ends up falling back to `RtlGenRandom`  https://github.com/rust-random/getrandom/blob/master/src/windows.rs as `BCryptGenRandom` can fail: https://github.com/rust-lang/rust/issues/94098 for `hashmap_random_keys` Rust just dumped `ByCryptGenRandom` for `ProcessPrng`: https://github.com/rust-lang/rust/pull/121337/files#diff-4c3f0c2423bd724c1e578979f565930038dc67b341d779dac9fdaab91fd5e77d

mrdomino:
> Did you mean `BCryptGenRandom`?

Yep. I am not a Windows guy, I just remembered that function existing and `CryptGenRandom` was for whatever reason the first search result for whatever I queried.

In any case it sounds like `ProcessPrng` is the thing to use nowadays, so, great.

libc/calls/ioctl.c
```diff
@@ -106,7 +106,8 @@ static int ioctl_fionread(int fd, uint32_t *arg) {
       int bytes = CountConsoleInputBytes();
       *arg = MAX(0, bytes);
       return 0;
-    } else if (g_fds.p[fd].kind == kFdDevNull) {
+    } else if (g_fds.p[fd].kind == kFdDevNull ||
+               g_fds.p[fd].kind == kFdDevRandom) {
       *arg = 1;
```
mrdomino, line 111: Is that true in this case? Presumably a read call will give you some number of bytes without blocking, would it make sense to put some number here? Like I dunno, 128 or 256? (If `ProcessPrng` tops out at some number, that's probably the one to use. Otherwise `getrandom` stops at 256, so maybe that?)

If this is what `urandom` does on other systems I'm fine with keeping it.

libc/calls/write-nt.c
```diff
@@ -52,6 +52,11 @@ static textwindows ssize_t sys_write_nt_impl(int fd, void *data, size_t size,
   struct Fd *f = g_fds.p + fd;
   bool isconsole = f->kind == kFdConsole;
 
+  // not implemented
+  if (f->kind == kFdDevRandom) {
+    return enosys();
```
mrdomino, line 57: I think you can just say `eperm()` here, that's what I get on a XNU system when I try to write to `/dev/urandom`.

test/libc/calls/specialfile_test.c
```diff
@@ -70,6 +71,50 @@ TEST(specialfile, devNullRead) {
   ASSERT_SYS(0, 0, close(3));
 }
 
+TEST(specialfile, devRandomRead) {
+  char buf[8] = {0};
+  ASSERT_SYS(0, 3, open("/dev/random", O_RDONLY));
+  // qemu-aarch64 defines o_largefile wrong
```
mrdomino, line 77: Seeing how many times this has to be written makes me want it to be factored out somehow. I dunno, maybe put the comment at the top of the file instead explaining why the individual tests are knocking out that bit?

Also I wonder if this is still relevant or if something better can be d one here, but I guess that's out of scope for this PR.

test/libc/calls/specialfile_test.c
```diff
@@ -70,6 +71,50 @@ TEST(specialfile, devNullRead) {
...
+  char buf[8] = {0};
+  ASSERT_SYS(0, 3, open("/dev/urandom", O_RDONLY));
+  // qemu-aarch64 defines o_largefile wrong
+  ASSERT_EQ(O_RDONLY, fcntl(3, F_GETFL) & ~(O_LARGEFILE | 0x00008000));
+  ASSERT_SYS(0, 8, read(3, buf, 8));
+  ASSERT_NE(0, memcmp(buf, "       ", 8));
+  ASSERT_SYS(0, 0, close(3));
+}
+
+TEST(specialfile, devRandomWrite_fails_on_nt) {
```
mrdomino, line 94: You can remove the `!IsWindows()` and check for `EPERM` everywhere. (I haven't tested everywhere but I'd be mildly surprised if any OS used a different error code for it. We should run it on the fleet before merging this I guess, so the test failure doesn't show up at some later point.)

mrdomino (changes requested):
Added a few comments. Overall I like it and the implementation seems basically sound to me.

I would want to run it by @jart prior to merging since this is a pretty major API change.

libc/calls/write-nt.c
G4Vi: For me on Linux:
```bash
echo dfdff > /dev/urandom
echo $?
0
```

`man 4 random` on my Linux system says:

> Writing to /dev/random or /dev/urandom will  update  the  entropy  pool
       with  the  data  written,  but this will not result in a higher entropy
       count.  This means that it will impact  the  contents  read  from  both
       files, but it will not make reads from /dev/random faster.

Does it work with root?

I figured `enosys()` was fitting as updating the entropy pool is not possible or at least not implemented, but if there's already a convention on returning `eperm()` we could use it instead.

G4Vi:
> Oh actually on closer inspection it looks like maybe you should use `__getrandom` out of `libc/calls/syscall_support-sysv.internal.h` instead of directly calling `ProcessPrng`. (The former is just a simple wrapper around the latter on Windows, so shouldn't have any impact.)

This was not done, following @jart 's comments here: 
https://github.com/jart/cosmopolitan/issues/1001#issuecomment-1853425711
> The only real concern to doing this is that `getrandom()` is a pretty heavy dependency for our `read()` function to pull in. We'd also have to implement this carefully, since `getrandom()` depends on `read()`

> Actually, now that I think about it, we're only talking about Windows code, so `sys_read_nt()` could implement support in a very lightweight way, since it'd only need to depend on `RtlGenRandom()`. Accepted.


If we need /dev/(u)random emulation on other plaforms we may have to.

libc/calls/write-nt.c
mrdomino: Wow, wacky. I dunno if you should be allowed to drive after smoking whatever those guys are on.

No it doesn't work with root on XNU.

But this unfortunately makes the tests more annoying to write.

I guess I'd say use `eperm()` and keep the write tests `IsWindows()`-gated.

mrdomino:
Hmm, okay, yeah. Using `ProcessPrng` directly seems fine.

Sorry, I hadn't noticed that issue link previously - in that case I'm fine with this and see no need to bug her again.

libc/calls/ioctl.c
G4Vi: I tried it out, it wasn't true even for `/dev/null`. Updated to Linux behavior: return `enotty()`  on `/dev/null` and `einval()` for `/dev/(u)random`.

libc/calls/ioctl.c
```diff
@@ -107,8 +107,9 @@ static int ioctl_fionread(int fd, uint32_t *arg) {
       *arg = MAX(0, bytes);
       return 0;
     } else if (g_fds.p[fd].kind == kFdDevNull) {
-      *arg = 1;
-      return 0;
+      return enotty();
```
jart, line 110: Nice

jart (approved):
Looks good to me.

libc/calls/ioctl.c
jart: Thank you for digging into that.

mrdomino (commented):
(I approve as well.)

PR #1164, opened 2024-05-03
fix nt accept/connect not initializing with SO_UPDATE_*_CONTEXT

G4Vi:
Fixes calling `getpeername`, `getsockname`, `getsockopt`, `setsockopt` after `accept`
and `getpeername`, `getsockname`, `getsockopt`, `setsockopt`, `shutdown` after `connect`.

### DIFF
--- a/libc/sock/accept-nt.c
+++ b/libc/sock/accept-nt.c
@@ -121,7 +121,8 @@ textwindows int sys_accept_nt(struct Fd *f, struct sockaddr_storage *addr,
 
   // inherit properties of listening socket
   // errors ignored as if f->handle was created before forking
-  // this fails with WSAENOTSOCK
+  // this fails with WSAENOTSOCK, see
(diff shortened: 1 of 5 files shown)


jart:
Is there any reason why this is a work in progress? Could you explain more about what exactly is happening here?

G4Vi:
This change alone is an improvement and can be merged now if you like!

Before `__imp_setsockopt(args->listensock, SOL_SOCKET,
                               kNtSoUpdateAcceptContext, &handle,
                               sizeof(handle))` was never (at least observed) being called as I think nonblocking io is always used to make the `AcceptEx` call. This change makes it always be called after a successful `accept`.

This was marked WIP to see if anything can be done about noticing that the `__imp_setsockopt` call fails with `WSAENOTSOCK` if `f->handle` was created before forking.

G4Vi:
I just added fixes for the same issue with `connect`.

jart (approved):

PR #1167, opened 2024-05-06
Update MODE=tiny time zone list

mrdomino:
Winnows down Europe to one zone per different set of current offsets / DST rules in each region that contains a Google office (so excepting Iceland, Kaliningrad, and Armenia / Georgia / Samara, which all do not observe DST and as such can use a `Etc/` zone without issue.)

Adds some outliers, in particular:

- Nova Scotia (as America/Halifax) (AST / ADT)
- Newfoundland (as America/St_Johns) (DST, offset by a half hour interval from UTC)
- Chatham Islands (offset by a 45 minutes interval from UTC)

Total net space in `o/tiny/examples/localtime` is about -5kb.


jart (changes requested):
Anything you removed, could you remove it from the list in `third_party/tz/windows.py` too? Then run `python third_party/tz/windows.py` to regenerate some of the content for `kWindowsZones` below.

mrdomino:
Done although I now wonder if we should include e.g. Budapest so that we get "Central Europe Standard Time", which, Brussels is mapped to "Romance Standard Time" (which is the same time zone) for some reason.

mrdomino:
Also do you want the newly added ones added to that list?

mrdomino:
I looked at it a bit more and I can't make heads or tails of the list of Windows timezone names.

I'm fine with merging this as-is, the supported list does not include anything that is not available anymore.

jart (approved):

PR #1168, opened 2024-05-06
loader.c uses AT_FLAGS to communicate literally

mrdomino:


ghaerr:
I hadn't noticed the previous AT_FLAGS changes, so I might be a little behind. Will this method work with macOS, given that it doesn't receive an AUX vector?

mrdomino:
This is a pretty niche piece of functionality — basically just a way for the loader to tell the binary that its argv[0] has been preserved. I'm sort of okay with it not working on x86_64 XNU; `GetProgramExecutableName()` (whose unit tests are this feature's only current user as far as I'm aware) already doesn't work reliably on x86_64 XNU.

ghaerr:
Agreed, it seems a number of (somewhat fancy?) niche modifications to the loader have been made, to solve the old problem of "who am i and where did I come from" for the loaded binary. I just wonder what their utility is if they come to be depended on and don't work on all mainline supported Cosmo host systems.

mrdomino:
If there is loader code that should be changed or removed, please just point at what it is in particular. Or better still, suggest a better design.

The current and planned work is to try to get a reasonable compromise on two desiderata that are in tension:

1. Binaries would like to be able to use argv[0] as a communication channel, e.g. for login shells.
2. Binaries need to be able to locate themselves for ZipOS to work.

We need 2 to work under all circumstances to the best of our ability. It would be nice for 1 to work where practicable.

The old school approach was to just sacrifice 1 entirely for the sake of 2. The new school approach is to try to get both more of the time.

mrdomino:
The only cases where point 2 currently does *not* work reliably are OpenBSD and x86_64 XNU assimilated binaries. We can fix x86_64 XNU trivially by reading out the `executable_path=` pseudo-auxv value instead of just zeroing it in `libc/crt/crt.S` on that platform. I don't think there's anything we can do about OpenBSD.

Point 1 doesn't work a lot more of the time, but is also already depended on by a bunch of stuff, so it's not like we're creating that problem by having incrementally better support for it.

ghaerr:
Hello @mrdomino,

I am sorry, my previous comments were not intended as a complaint in any way. I'm still developing mostly on Intel macOS, with my point of view mostly coming from that perspective.

> The new school approach is to try to get both more of the time.

Sounds good!

> We can fix x86_64 XNU trivially by reading out the executable_path= pseudo-auxv value 

That also sounds like potentially a good idea. I had thought of the idea of fully polyfilling aux for macOS x86_64, but wasn't sure most of the useful auxv entries could be usefully emulated, nor whether more dependencies on upstairs auxv[] access is a good idea within Cosmopolitan either.

> instead of just zeroing it in libc/crt/crt.S on that platform.

I think auxv is also zeroed out in ape/loader.c when os is XNU.

Thank you!

mrdomino:
> I am sorry, my previous comments were not intended as a complaint in any way. I'm still developing mostly on Intel macOS, with my point of view mostly coming from that perspective.

No worries and that makes perfect sense then; yeah, x86_64 XNU is at risk of becoming a dinosaur, so it's good to be vocal about stuff that doesn't work right on it.

> That also sounds like potentially a good idea.

If you are interested, I have just filed #1169 for this and would be happy to answer any questions or chat about it further on redbean discord if you wanted to tackle it. Otherwise it is on my radar but currently “best effort” (meaning I may get around to it one day.)

> I think auxv is also zeroed out in ape/loader.c when os is XNU.

You're right, it is, but we don't care about it in this case since `executable_path=` is going to point to the loader and not the binary.

mrdomino:
I guess if we are continuing down the road of using `AT_FLAGS` to signal that argv[0] is preserved, and if we care about XNU, then another thing that might be done is to make that snippet in `crt.S` not run if we notice that we've been loaded.

Strawman logic might be:

```
if __program_executable_name is set
  then do nothing, the loader has already set auxv up for us
else
  if it is good to do so (i.e. we're assimilated, not running under an older loader)
    then save executable_path= to __program_executable_name
    and zero out auxv
```

ghaerr:
> if we are continuing down the road of using AT_FLAGS to signal that argv[0] is preserved, and if we care about XNU

You're into this deeper than I am, but it seems to me if both the above are true, then perhaps auxv should always be emulated by the loader, as well as saving executable_path= on macOS beforehand. I realize this may be a mess for assimilated binaries.

Are there other higher-level cases where auxv[] contents are used (e.g. AT_PAGESZ)? I don't know how much auxv[] is used within Cosmopolitan, nor whether Cosmo applications can be assumed to reliability access various otherwise oft-used AT_xxx parameters.

ghaerr:
> Strawman logic might be:

I see now, yes. Since auxv can't be emulated for assimilated binaries, this might be the best we can do, using the new school approach of trying to get the most that's possible. If the loader runs, it could polyfill an ELF auxv, for possible higher level use. If not, then the assimilated binary saves executable_path= and creates a zeroed auxv.

In the assimilated case above, there would have to be internal Cosmo workarounds (defaults) for any code that otherwise uses a host-OS supplied or loader-emulated ELF auxv, which I am assuming is already the case.

mrdomino:
> In the assimilated case above, there would have to be internal Cosmo workarounds (defaults) for any code that otherwise uses a host-OS supplied or loader-emulated ELF auxv, which I am assuming is already the case.

Yep. It looks like the only aux value that is read with any regularity is `AT_PAGESZ`, which is already polyfilled.

jart:
@ghaerr Nothing is niche if one of us needs it. Especially when that person is volunteer to be the one to write it. I believe Cosmopolitan should be inclusive of your needs, @mrdomino's needs, and others too. I believe we can keep the code simple and elegant in the process if we devote ourselves to having a culture that values this. However our needs and curiosities should come first, since there's no shame in refactoring things later.

mrdomino:
(Sorry, not ready yet. I got mixed up about which PR this was.)

mrdomino:
I'm gonna go ahead and say this is ready. Should work everywhere except x86_64 XNU, which has an issue filed. I will leave it to be tackled in a future PR, possibly by some intrepid young coder, who might even be me.

mrdomino:
(comments marked off-topic, including mine, are fine and good but not relevant to this PR; discussion should probably happen at #1169 or in new issues instead)

G4Vi (commented):
The code looks good. Before I approve, I'm curious how this is intended to be used? Should programs dependent on preserved argv[0]/non path argv[0] check the flag? I'd imagine that would be error prone as programs could be assimilated and I'd think it wouldn't be set then? It's okay if this is purely advisory information too. Feel free to just link me is this is explained elsewhere, I haven't followed the argv[0] preserving improvements super closely.

mrdomino:
The more I think about this, the less I like it.

That flag has particular semantics in the context of binfmt on Linux. It's the kernel's way of telling a binfmt interpreter, "I have incremented argc, and your argv[1] is the original argv[0] of the script you are interpreting."

The ape binary, when run as a binfmt interpreter, now knows how to act on this flag in order to present the correct arguments to the binary it loads.

But this flag currently loses its meaning in the context of the loaded binary. The ape interpreter has left it set as a side effect; this side effect can be observed, which will tell the binary that its interpreter has preserved argv[0]. But suppose you want to use an assimilated ape binary as an interpreter; now, it needs to know whether that flag came from the kernel or from the ape loader, and it has two different meanings in those two cases.

Overall I do not like this anymore. I think if anything, the ape loader should actually *unset* this flag if it is set. The test cases should do something else to decide whether to test argv[0].

A bit late in the game, but it also occurs to me that you could use this flag in the original sense; increment argc and pass an extra argv with the original argv[0] in it, instead of doing the register trick. At this point just more arguments for unsetting the flag and not abusing it in current binaries...

PR #1171, opened 2024-05-07
Reapply apeinstall.sh binfmt flags

mrdomino:
7d31fc3 made it safe to register ape with the binfmt P flag. Older files
will still get their paths passed via argv[0] and everything should just
work. This also restores the F flag that was rolled back alongside the P
flag; this seems like a good idea to me now. It makes it so /usr/bin/ape
is kind of part of the kernel; simply replacing the file will not change
the loader used. The binfmt will have to be reregistered as well.

ape/apeinstall.sh will already nag you if there's an existing binfmt for
ape. It might be nice to make it actually replace the registration.

This reverts commit df648fb17457a3f9f770ddd9898271fa3f4fbf1c.


mrdomino:
Based off of #1170.

G4Vi (approved):
I'm glad preserving argv[0] is now possible in more scenarios. Approved!

PR #1179, opened 2024-05-18
Fix #920 (sleb64 is certainly broken)

### LINKED ISSUE #920: sleb64 is certainly broken
At the moment it takes every 7 bits to make a char and corrupts 7th bit by bit-or with 64.

It should either take 6 bit, or do bit-or with 128 (as in uleb64 and zleb64).

(sorry for absence of formatting, I wrote it using phone).

mrdomino:


mrdomino:
Does this look right to you @jart? I think this is correct based on [wiki/LEB128](https://en.wikipedia.org/wiki/LEB128) but I'm a little iffy on the specifics.

jart (approved):

PR #1185, opened 2024-05-20
Run all BLAKE2B256 test vectors

mrdomino:


mrdomino:
This may be a bit heavy-handed, I just followed the example for hyperion.txt. I wound up having to move the vectors into libc/testlib, since I couldn't get it to work in the old location. If there is a better way to do this I would love to learn it.

jart (approved):
I'll relocate it when I have time. We need a more intuitive and simple way to configure the build for these sorts of things.

Issue #1188, opened 2024-05-25
[Windows] execve regression


G4Vi:
Before cf70a4475651d9864f6a16b88ebc6930659d0898 the following program paginates a file: on Windows.

```c
#include <unistd.h>

int main(int argc, char *argv[]) {
    pid_t pid = fork();
    if (pid == 0) {
        char *newargv[] = {"/C/Windows/system32/more.com", "C:\\Users\\username\\AppData\\Local\\Temp\\8z8K3x9WA2", NULL};
        char *noenv[] = {NULL};
        execve("/C/Windows/system32/more.com", newargv, noenv);
    } else {
        wait(NULL);
    }
}
```

`CreateProcess` succeeds in both places, but immediately terminates since cf70a4475651d9864f6a16b88ebc6930659d0898.

The `--strace` before and after looks the same. No changes appear to be made to `lpApplicationName` and `lpCommandLine`.

PR #1190, opened 2024-05-27
ntspawn: fix initializing NtStartupInfoEx

G4Vi:
Fixes https://github.com/jart/cosmopolitan/issues/1188

Initializing `info.StartupInfo.cb = sizeof(info)` inside of the initializer statement was clearing the rest of `info.StartupInfo`.


jart (approved):
Ah, makes sense. Thanks!

PR #1197, opened 2024-06-01
Run clang-format

mrdomino:
Scanning over this, most of it looks fine - the only place where we might want a `clang-format off` would be `libc/str/iswpunct.c`.


jart (approved):

jart:
```
In file included from tool/decode/lib/disassemblehex.c:19:
./tool/decode/lib/disassemblehex.h:8:49: error: unknown type name ‘FILE’
    8 | void disassemblehex(uint8_t *data, size_t size, FILE *f);
      |                                                 ^~~~
```

Merge this yourself when the CI is green. Thanks!

PR #1198, opened 2024-06-01
Add .git-blame-ignore-revs

mrdomino:
If you follow the directions in that file then git blame will ignore the listed commits. A commit should only go in that file if its only changes were to formatting, particularly on a large part of the codebase (like a change to .clang-format getting applied to the repo.)

Cribbed from here:

https://www.stefanjudis.com/today-i-learned/how-to-exclude-commits-from-git-blame/

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,23 @@
+# Run this command to always ignore formatting commits in git blame
+# git config blame.ignoreRevsFile .git-blame-ignore-revs
+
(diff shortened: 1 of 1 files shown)


mofosyne (approved):

mofosyne:
due diligence check that all commit corresponds to the title in the comments - passed

mrdomino:
So generally speaking we do "squash and merge" on commits here, so that the repo has a single linear commit history.

PR #1199, opened 2024-06-06
wip ctl::string small-string optimization

mrdomino:
A small-string optimization is a way of reusing inline storage space for sufficiently small strings, rather than allocating them on the heap. The current approach takes after an old Facebook string class: it reuses the highest-order byte for flags and small-string size, in such a way that a maximally-sized small string will have its last byte zeroed, making it a null terminator for the C string.

The only flag we have is in the highest-order bit, that says whether the string is big (set) or small (cleared.) Most of the logic switches based on the value of this bit; e.g. data() returns big()->p if it's set, else small()->buf if it's cleared. For a small string, the capacity is always fixed at sizeof(string) - 1 bytes; we store the length in the last byte, but we store it as the number of remaining bytes of capacity, so that at max size, the last byte will read zero and serve as our null terminator.

Morally speaking, our class's storage is a union over two POD C structs. For now I gravitated towards a slightly more obtuse approach: the string class itself contains a blob of the right size, and we alias that blob's pointer for the two structs, taking some care not to run afoul of object lifetime rules in C++. If anyone wants to improve on this, contributions are welcome.

This commit also introduces the `ctl::__` namespace. It can't be legally spelled by library users, and serves as our version of boost's "detail".

We introduced a string::swap function, and we now use that in operator=. operator= now takes its argument by value, so we never need to check for the case where the pointers are equal and can just swap the entire store of the argument with our own, leaving the C++ destructor to free our old storage afterwards.

There are probably still a few places where our capacity is slightly off and we grow too fast, although there don't appear to be any where we are too slow. I will leave these to be fixed in future changes.

TODO:

- [x] tests are currently segfaulting
- [x] think about operator string_view (thought about it - it's still only 1.3ns to make one)
- ~maybe migrate to POD anonymous union~ (i like this way better)
- [x] benchmark and see if this is even worth it
- [x] __ namespace is documented, at least here
- [ ] we are probably incorrectly setting size in a few places
- [x] explain why assign-by-value and "swapperator", at least here

### DIFF
--- a/ctl/string.cc
+++ b/ctl/string.cc
@@ -38,7 +38,7 @@ big_string::~big_string() /* noexcept */
     free(p);
 }
 
-}   // namespace __
+} // namespace __
 
 string::~string() /* noexcept */
 {
@@ -97,11 +97,11 @@ string::reserve(size_t c2) noexcept
     if (c2 <= __::sso_max)
         return;
     if (!isbig()) {
-        if (!(p2 = (char *)malloc(c2)))
+        if (!(p2 = (char*)malloc(c2)))
             __builtin_trap();
         __builtin_memcpy(p2, data(), size() + 1);
     } else {
-        if (!(p2 = (char *)realloc(big()->p, c2)))
+        if (!(p2 = (char*)realloc(big()->p, c2)))
             __builtin_trap();
     }
     std::atomic_signal_fence(std::memory_order_seq_cst);
--- a/ctl/string.cc
+++ b/ctl/string.cc
@@ -47,27 +47,27 @@ string::~string() /* noexcept */
     }
 }
 
-string::string(const char* s) noexcept
+string::string(const char* s) noexcept : string()
 {
     append(s, strlen(s));
 }
 
-string::string(const string& s) noexcept
+string::string(const string& s) noexcept : string()
 {
     append(s.data(), s.size());
 }
 
-string::string(const string_view s) noexcept
+string::string(const string_view s) noexcept : string()
 {
     append(s.p, s.n);
 }
 
-string::string(size_t size, char ch) noexcept
+string::string(size_t size, char ch) noexcept : string()
 {
     resize(size, ch);
 }
 
-string::string(const char* s, size_t size) noexcept
+string::string(const char* s, size_t size) noexcept : string()
 {
     append(s, size);
 }
@@ -75,6 +75,8 @@ string::string(const char* s, size_t size) noexcept
 const char*
 string::c_str() const noexcept
 {
+    if (!size())
+        return "";
     if (size() >= capacity())
         __builtin_trap();
     if (data()[size()])
@@ -86,8 +88,9 @@ void
 string::reserve(size_t c2) noexcept
 {
     char* p2;
-    if (c2 < size())
-        c2 = size();
+    size_t n = size();
+    if (c2 < n)
+        c2 = n;
     if (ckd_add(&c2, c2, 15))
         __builtin_trap();
     c2 &= -16;
@@ -101,7 +104,6 @@ string::reserve(size_t c2) noexcept
         if (!(p2 = (char *)realloc(big()->p, c2)))
             __builtin_trap();
     }
-    size_t n = size();
     std::atomic_signal_fence(std::memory_order_seq_cst);
     set_big_capacity(c2);
     big()->n = n;
--- a/ctl/string.cc
+++ b/ctl/string.cc
@@ -23,7 +23,9 @@
 
 namespace ctl {
 
-string::~string() noexcept
+namespace __ {
+
+big_string::~big_string() /* noexcept */
 {
     if (n) {
         if (n >= c)
@@ -36,6 +38,15 @@ string::~string() noexcept
     free(p);
 }
 
+}   // namespace __
+
+string::~string() /* noexcept */
+{
+    if (isbig()) {
+        big()->~big_string();
+    }
+}
+
 string::string(const char* s) noexcept
 {
     append(s, strlen(s));
@@ -43,7 +54,7 @@ string::string(const char* s) noexcept
 
 string::string(const string& s) noexcept
 {
-    append(s.p, s.n);
+    append(s.data(), s.size());
 }
 
 string::string(const string_view s) noexcept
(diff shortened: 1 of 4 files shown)


ctl/string.cc
```diff
@@ -64,29 +75,37 @@ string::string(const char* s, size_t size) noexcept
...
         __builtin_trap();
     c2 &= -16;
-    if (!(p2 = (char*)realloc(p, c2)))
-        __builtin_trap();
+    if (c2 <= __::sso_max)
+        return;
+    if (!isbig()) {
+        if (!(p2 = (char *)malloc(c2)))
+            __builtin_trap();
+        __builtin_memcpy(p2, data(), size() + 1);
```
jart, line 99: `__builtin_memcpy()` would make sense for a lightweight header, but in the `.cc` file you can do anything.

ctl/string.cc
```diff
@@ -96,41 +115,54 @@ string::resize(size_t n2, char ch) noexcept
...
+        __builtin_trap();
+    if (n2 > capacity()) {
+        size_t c2 = capacity() + 2;
+        if (ckd_add(&c2, c2, c2 >> 1))
+            __builtin_trap();
         reserve(c2);
     }
-    p[n++] = ch;
-    p[n] = 0;
+    // XXX do we care to fence this?
```
jart, line 140: The fence is used because if a trap or asynchronous signal is raised, and I want to `pthread_exit()` the thread from the signal handler as a result of that, I'd like to be able to have my pthread cleanup callbacks free these strings, or alternatively, re-use the objects for the next request. So for example, don't say this:

```
if (!(this->p = realloc(this->p, c2)))
  __builtin_trap();
```

Because then the signal handler can't free the original allocation. The object needs to be in a reasonable state when trap is called, or an exception is thrown, otherwise I want to know if that's not possible. By reasonable I mean good enough to ~destroy or clear() and go at it again.

ctl/string.h
```diff
@@ -6,29 +6,59 @@
...
 
 string
 strcat(const string_view, const string_view) noexcept __wur;
 
-struct string
+namespace __ {
+
+constexpr size_t string_size = 3 * sizeof(size_t);
+constexpr size_t sso_max = string_size - 1;
+constexpr size_t big_mask = ~(1ull << (8ull * sizeof(size_t) - 1ull));
```
jart, line 18: String needs to be 0x7fffffffffffffff characters long before it stops using sso?

jart:
I'm liking this so far. I'm glad to see Facebook's alpha coming to the Cosmopolitan codebase. Since we're replacing simple code with intelligent code, I'd like to see a benchmark too that demonstrates the advantage. Please don't use my legacy benchmark macros. Something simple and less frameworky like this should do:

```c
#define BENCH(ITERATIONS, WORK_PER_RUN, CODE) \
    do { \
        struct timespec start = timespec_real(); \
        for (int i = 0; i < ITERATIONS; ++i) { \
            asm volatile("" ::: "memory"); \
            CODE; \
        } \
        long long work = WORK_PER_RUN * ITERATIONS; \
        double nanos = (timespec_tonanos(timespec_sub(timespec_real(), start)) + work - 1) / (double)work; \
        printf("%10g ns %2dx %s\n", nanos, ITERATIONS, #CODE); \
    } while (0)
```

ctl/string.h
```diff
@@ -220,8 +285,69 @@ struct string
     {
         return compare(s) >= 0;
     }
+
+  private:
+    inline bool isbig() const noexcept
+    {
+        return *(__builtin_launder(blob) + __::sso_max) & 0x80;
```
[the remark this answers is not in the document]
mrdomino: move and forward are both trivial plain C++ functions; there is no reason we couldn't write a brief `ctl::move` and `ctl::forward` of our own.

E.g. capnproto's libkj does this:

https://github.com/capnproto/capnproto/blob/5faf3144b5e1b0fbbe80992f7e3a890a472c82a3/c%2B%2B/src/kj/common.h#L678-L687

`std::launder` is different. It is an escape hatch from the compiler's ability to make crazy assumptions based on the object lifetime rules of C++ - I have not carefully studied this in a bit but my understanding is, for example, if you have a byte that is part of a union where one arm is `const` and the other isn't, and you change it in the non-`const` arm, then unless you call `launder` when referencing the `const` arm, the compiler can just assume that the value did not change and not bother to look.

Since you obviously can't implement this in terms of the actual C++ language itself, there is going to be a `__builtin_launder` or something that your STL provider is going to collaborate with your compiler author to use.

ctl/string.cc
ghaerr: This is pretty neat stuff you guys are doing.

I'm assuming the issue here is not to rewrite `this->p` before knowing `realloc(this->p, ...)` succeeds. Would just using `reallocf` solve the problem? I am also unsure as to whether the standard specifies that the original memory allocation (ie. first argument to realloc) is guaranteed to remain valid when realloc fails.

mrdomino:
master:

```
   17.9248 ns 10000000x { ctl::string s; s.append("hello "); s.append("world"); }
   4.10612 ns 1000000x { ctl::string s; for (int i = 0; i < 8; ++i) { s.append('a'); } }
   3.94862 ns 1000000x { ctl::string s; for (int i = 0; i < 16; ++i) { s.append('a'); } }
   3.72796 ns 1000000x { ctl::string s; for (int i = 0; i < 23; ++i) { s.append('a'); } }
   3.83362 ns 1000000x { ctl::string s; for (int i = 0; i < 32; ++i) { s.append('a'); } }
    13.607 ns 1000000x { ctl::string s("hello world"); }
    12.453 ns 1000000x { ctl::string s2(s); }
     17.64 ns 1000000x { ctl::string s("hello world"); ctl::string s2(std::move(s)); }
    24.125 ns 1000000x { ctl::string s("hello world"); ctl::string s2(s); }
    12.818 ns 1000000x { ctl::string s(23, 'a'); }
     13.22 ns 1000000x { ctl::string s(24, 'a'); }
```

ctl-sso:

```
   10.4207 ns 10000000x { ctl::string s; s.append("hello "); s.append("world"); }
     3.938 ns 1000000x { ctl::string s; for (int i = 0; i < 8; ++i) { s.append('a'); } }
   4.50244 ns 1000000x { ctl::string s; for (int i = 0; i < 16; ++i) { s.append('a'); } }
   4.83043 ns 1000000x { ctl::string s; for (int i = 0; i < 23; ++i) { s.append('a'); } }
     4.833 ns 1000000x { ctl::string s; for (int i = 0; i < 32; ++i) { s.append('a'); } }
     7.472 ns 1000000x { ctl::string s("hello world"); }
     6.844 ns 1000000x { ctl::string s2(s); }
      8.18 ns 1000000x { ctl::string s("hello world"); ctl::string s2(std::move(s)); }
    13.086 ns 1000000x { ctl::string s("hello world"); ctl::string s2(s); }
      6.89 ns 1000000x { ctl::string s(23, 'a'); }
    18.017 ns 1000000x { ctl::string s(24, 'a'); }
```

ctl/string.cc
jart: Yes. It's an ironclad guarantee that when `realloc()` fails, the original memory is left unharmed. I haven't heard of `reallocf()` until now. It sounds similar to `realloc_in_place()` which is likely its modern replacement, which has good semantics. Really happy to see you here btw!

ctl/string.h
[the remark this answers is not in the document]
jart: Thank you. This information has much more clarity than Google results.

jart (approved):

ctl/string.cc
mrdomino: Indeed, good to see you @ghaerr!

In any case, resolving this conversation as for this PR's purposes the answer to the question is simply "no."

PR #1204, opened 2024-06-08
Make big_string pod

alkis:
`big_string` is not pod which means it needs to be properly constructed and destroyed. Instead make it POD and destroy it manually in `string` destructor.

### DIFF
--- a/ctl/string.h
+++ b/ctl/string.h
@@ -38,8 +38,6 @@ struct big_string
     size_t c : sizeof(size_t) * 8 - 1;
     size_t big : 1 /* = 1 */;
 #endif
-
-    ~big_string() /* noexcept */;
 };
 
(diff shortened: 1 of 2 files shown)


ctl/string.h
```diff
@@ -51,7 +49,7 @@ class string
     using const_iterator = const char*;
     static constexpr size_t npos = -1;
 
-    ~string() /* noexcept */;
+    ~string() noexcept;
```
mrdomino, line 52: Destructors are noexcept by default, I would rather leave this commented.

mrdomino (approved):
LGTM

Thanks!

ctl/string.h
alkis: Done.

PR #1211, opened 2024-06-10
Decouple swap from std

mrdomino:
This allows you to implement your own swap function without it having to be part of the std namespace. std::swap is still used if it's available.


jart (approved):
nice trick

PR #1212, opened 2024-06-11
Causes cosmoc++ to segfault

mrdomino:
cosmoc++ produces "internal compiler error: SIGSEGV" on this code if run via:

    mmake ctl/unique_ptr.o


ahgamut:
Seems like debian's `g++-10` also segfaults for this, as does cosmo's `g++-12` without the `-fportcosmo` flag.

mrdomino:
Apparently it’s a gcc bug, fixed by 14.1.

ahgamut:
https://godbolt.org/z/PG3rGa6v1

PR #1213, opened 2024-06-11
[Redbean] Add UuidV7 method

mterron:
To Complete #1140 add UUID version 7 to Redbean


tool/net/lfuncs.c
```diff
@@ -854,6 +858,22 @@ int LuaUuidV4(lua_State *L) {
   return 1;
 }
 
+int LuaUuidV7(lua_State *L) {
+  char uuid_str[37] = {0};
+  struct timespec now = timespec_real();
+  uint64_t time_now = timespec_tonanos(now);
```
jart, line 864: This function isn't really intended to be called on an absolute timestamp. Could you use the `tv_sec` and `tv_nsec` fields instead?

jart (changes requested):

mrdomino:
On looking at [the spec](https://datatracker.ietf.org/doc/html/rfc9562#name-uuid-version-7), it says that the first 48 bits of a UUIDv7 should be the timestamp in milliseconds since the UNIX epoch. The rest is allowed to be random data, but is also allowed to be a sub-millisecond timestamp or monotonic counter (to ensure sortability at sub-millisecond resolution.)

I'm a bit confused about all the conversions vs bit arithmetic at the moment, but it doesn't look like that's what the current implementation is doing - is it?

I would suggest instead taking, as your 64-bit value, the following:

```c
uint64_t now_ms = now.tv_sec * 1000 + now.tv_nsec / 1000000;
```

Put that in the first 48 bits. Then you have a choice; there are about (just shy of) 20 bits of further entropy in the tv_nsec field, and at least on my system the first (just shy of) 10 bits are supposed to be accurate for time ordering. So if you wanted, you could have the next 10–20 bits of the UUID come from the lower-order bits of that field.

But you don't have to; you could just fill the rest of the UUID after the first 48 bits with random data.

(I have questions about rounding with regard to the nanosecond field too but I don't think those are as important right now.)

mterron:
> On looking at [the spec](https://datatracker.ietf.org/doc/html/rfc9562#name-uuid-version-7), it says that the first 48 bits of a UUIDv7 should be the timestamp in milliseconds since the UNIX epoch. The rest is allowed to be random data, but is also allowed to be a sub-millisecond timestamp or monotonic counter (to ensure sortability at sub-millisecond resolution.)
> 
> I'm a bit confused about all the conversions vs bit arithmetic at the moment, but it doesn't look like that's what the current implementation is doing - is it?
> 
> I would suggest instead taking, as your 64-bit value, the following:
> 
> ```c
> uint64_t now_ms = now.tv_sec * 1000 + now.tv_nsec / 1000000;
> ```
> 
> Put (the lower 48 bits of) that in the first 48 bits. Then you have a choice; there are about (just shy of) 20 bits of further entropy in the tv_nsec field, and at least on my system the first (just shy of) 10 bits are supposed to be accurate for time ordering. So if you wanted, you could have the next 10–20 bits of the UUID come from the lower-order bits of that field.
> 
> But you don't have to; you could just fill the rest of the UUID after the first 48 bits with random data.
> 
> (I have questions about rounding with regard to the nanosecond field in my proposed implementation but I don't think those are as important right now.)

This is exactly what the implementation does actually. I have talked to Justine and have a less "magical" (or more explicit) implementation almost ready to push.

Thanks for the feedback! I won't try to be clever on my next commit.

mrdomino:
Oh awesome, very good then. I dunno, cleverness is fine if it’s well documented, or if it’s clear enough that it doesn’t need to be documented.

tool/net/lfuncs.c
```diff
@@ -858,6 +858,73 @@ int LuaUuidV4(lua_State *L) {
   return 1;
 }
 
+int LuaUuidV7(lua_State *L) {
```
jart, line 861: I like how this code very clearly communicates what's happening. I bet GCC turns it into marvelously small and fast assembly too. You should try comparing it to your previous benchmarks. Do me one favor, and add a comment citing the RFC(s) you're implementing? Also @mrdomino could you give this another glance before we merge it?

tool/net/help.txt
```diff
@@ -1065,6 +1065,9 @@ FUNCTIONS
   UuidV4() -> str
           Returns an uuid v4 string.
 
+  UuidV7() -> str
+          Returns an uuid v7 string.
```
jart, line 1069: s/an/a/

test/tool/net/uuidv7_test.lua
```diff
@@ -0,0 +1,25 @@
+-- Copyright 2024 Justine Alexandra Roberts Tunney
```
jart, line 1: Assuming you want to get your name out there, you are encouraged to own the copyright on your test code, so long as you release it under the same ISC license (see CONTRIBUTING.md). But it's optional. You're perfectly welcome to put my name here too, since I approve of the work you're doing. Thank you for it!

jart (approved):
A few comments and requested changes, but overall LGTM.

test/tool/net/uuidv7_test.lua
mterron: That's ok, I copied the uuidv4 one verbatim. I don't need any recognition for that 😆

tool/net/lfuncs.c
mterron: It performs about the same,  ~half as fast as uuidv4

tool/net/lfuncs.c
```diff
@@ -858,6 +858,74 @@ int LuaUuidV4(lua_State *L) {
...
+  uuid_str[11] = "0123456789abcdef"[(bin[5] & 0xf0) >>4];
+  uuid_str[12] = "0123456789abcdef"[(bin[5] & 0x0f)];
+  uuid_str[13] = '-';
+  uuid_str[14] = '7';
+  uuid_str[15] = "0123456789abcdef"[(bin[6] & 0xf0) >>4];
+  uuid_str[16] = "0123456789abcdef"[(bin[6] & 0x0f)];
+  uuid_str[17] = "0123456789abcdef"[(bin[7] & 0xf0) >>4];
+  uuid_str[18] = '-';
+  uuid_str[19] = "0123456789abcdef"[(0x8 | ((bin[7] & 0x0f) >>2))];
+  uuid_str[20] = "0123456789abcdef"[(bin[7] & 0x03) | (bin[8] & 0xf0) >>6]; //See https://www.rfc-editor.org/rfc/rfc9562.html#version_field
```
mrdomino, line 907: might wind up needing to move this comment to the its own line above

mrdomino (approved):
LGTM

Remember to run clang-format.

mterron:
> LGTM
> 
> Remember to run clang-format.

I can do that, it makes it really ugly and hard to follow IMO though

mrdomino:
If it's giving you trouble, then my opinion would be maybe don't worry about it for now and it can get cleaned up later.

mterron:
> If it's giving you trouble, then my opinion would be maybe don't worry about it for now and it can get cleaned up later.

Let's keep it this way and merge it then

jart (approved):
LGTM. Thank you!

PR #1215, opened 2024-06-15
ctl::string cleanup

mrdomino:
We now fully initialize a ctl::string’s memory, so that it is always set to a well-defined value, thus making it always safe to memcpy out of it. This incidentally makes our string::swap function legal, which it wasn’t before. This also saves us a store in string::reserve.

Now that we have made both big_string and smalL_string POD, I believe it is safe to elide the launder calls, and have done so, thus cleaning up a lot of the blob-related code.

I also got rid of set_big_capacity and replaced it with a set_big_string that leaves us in a well-defined state afterwards. This function also is able to be somewhat simpler; rather than delicate bit-twiddling, it just reaches straight into blob and rewrites it wholesale.

Overall, this shaves about 1–2ns off of most benchmarks, and adds 1ns to only one of them - creating a string from a char *.

### DIFF
--- a/ctl/string.cc
+++ b/ctl/string.cc
@@ -91,16 +91,13 @@ string::reserve(size_t c2) noexcept
     if (!isbig()) {
         if (!(p2 = (char*)malloc(c2)))
             __builtin_trap();
-        memcpy(p2, data(), size());
-        p2[size()] = 0;
+        memcpy(p2, data(), size() + 1);
     } else {
         if (!(p2 = (char*)realloc(big()->p, c2)))
             __builtin_trap();
     }
     std::atomic_signal_fence(std::memory_order_seq_cst);
-    set_big_capacity(c2);
-    big()->n = n;
-    big()->p = p2;
+    set_big_string(p2, n, c2);
 }
 
 void
--- a/ctl/string.cc
+++ b/ctl/string.cc
@@ -91,7 +91,7 @@ string::reserve(size_t c2) noexcept
     if (!isbig()) {
         if (!(p2 = (char*)malloc(c2)))
             __builtin_trap();
-        memcpy(p2, data(), size() + 1);
+        memcpy(p2, data(), __::string_size);
     } else {
         if (!(p2 = (char*)realloc(big()->p, c2)))
             __builtin_trap();
--- a/ctl/string.h
+++ b/ctl/string.h
@@ -81,28 +81,23 @@ class string
 
     string() noexcept
     {
-        set_small_size(0);
-#if 0
-        small()->buf[0] = 0;
-#endif
+        __builtin_memset(blob, 0, sizeof(size_t) * 2);
+        // equivalent to set_small_size(0) but also zeroes memory
+        *(((size_t *)blob) + 2) = __::sso_max << (sizeof(size_t) - 1) * 8;
     }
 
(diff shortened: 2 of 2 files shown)


mrdomino:
master:

```
   10.1243 ns 10000000x { ctl::string s; s.append("hello "); s.append("world"); }
   3.79312 ns 1000000x { ctl::string s; for (int i = 0; i < 8; ++i) { s.append('a'); } }
   4.20162 ns 1000000x { ctl::string s; for (int i = 0; i < 16; ++i) { s.append('a'); } }
   4.03643 ns 1000000x { ctl::string s; for (int i = 0; i < 23; ++i) { s.append('a'); } }
   4.89956 ns 1000000x { ctl::string s; for (int i = 0; i < 32; ++i) { s.append('a'); } }
     7.245 ns 1000000x { ctl::string s("hello world"); }
     6.634 ns 1000000x { ctl::string s2(s); }
     8.184 ns 1000000x { ctl::string s("hello world"); ctl::string s2(std::move(s)); }
    13.006 ns 1000000x { ctl::string s("hello world"); ctl::string s2(s); }
     6.717 ns 1000000x { ctl::string s(23, 'a'); }
    16.947 ns 1000000x { ctl::string s(24, 'a'); }
     1.312 ns 1000000x { ctl::string_view s2(s); }
```

sso-cleanup:
```
    8.2504 ns 10000000x { ctl::string s; s.append("hello "); s.append("world"); }
   3.65112 ns 1000000x { ctl::string s; for (int i = 0; i < 8; ++i) { s.append('a'); } }
    3.6805 ns 1000000x { ctl::string s; for (int i = 0; i < 16; ++i) { s.append('a'); } }
   3.76074 ns 1000000x { ctl::string s; for (int i = 0; i < 23; ++i) { s.append('a'); } }
   3.64453 ns 1000000x { ctl::string s; for (int i = 0; i < 32; ++i) { s.append('a'); } }
     8.163 ns 1000000x { ctl::string s("hello world"); }
      5.39 ns 1000000x { ctl::string s2(s); }
     6.936 ns 1000000x { ctl::string s("hello world"); ctl::string s2(std::move(s)); }
    10.395 ns 1000000x { ctl::string s("hello world"); ctl::string s2(s); }
     5.687 ns 1000000x { ctl::string s(23, 'a'); }
    15.712 ns 1000000x { ctl::string s(24, 'a'); }
     1.312 ns 1000000x { ctl::string_view s2(s); }
```

ctl/string.cc
```diff
@@ -91,16 +91,13 @@ string::reserve(size_t c2) noexcept
     if (!isbig()) {
         if (!(p2 = (char*)malloc(c2)))
             __builtin_trap();
-        memcpy(p2, data(), size());
-        p2[size()] = 0;
+        memcpy(p2, data(), size() + 1);
```
jart, line 94: wouldn't it be faster to just memcpy by a constant the whole thing?

jart (approved):

ctl/string.cc
mrdomino: oh yeah good point

mrdomino: shaves 1ns off the append benchmark that does the initial jump to big string so yeah good catch.

(some others slightly increased but that's gotta be just noise)

PR #1218, opened 2024-06-15
ctl string const/value tweaks

mrdomino:
The mangled name of a C++ function will typically not vary by const-ness of a by-value parameter; in other words, there is no meaning to a const- qualified by-value parameter in a function prototype. However, the const keyword \__does_\_ matter at function \__definition_\_ time, like it does with a variable declared in the body. So for prototypes, we strip out const for by-value parameters; but for definitions, we leave them alone.

At function definition (as opposed to prototype), we add const to values in parameters by default, unless we’re going to mutate them.

This commit also changes a couple of `const string_view&` to be simply by- value `string_view`. A `string_view` is only two words; it rarely ever makes sense to pass one by reference if it’s not going to be mutated.

### DIFF
--- a/ctl/string.cc
+++ b/ctl/string.cc
@@ -54,12 +54,12 @@ string::string(const string_view s) noexcept : string()
     append(s.p, s.n);
 }
 
-string::string(size_t size, char ch) noexcept : string()
+string::string(const size_t size, const char ch) noexcept : string()
 {
     resize(size, ch);
 }
 
-string::string(const char* s, size_t size) noexcept : string()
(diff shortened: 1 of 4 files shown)


mrdomino:
I split out the commit that const-qualifies method definitions, since that might be more contentious - I don't care strongly either way on it, but I've known people who've insisted on it in the past, and it's closer to the thing that you get out of e.g. Rust's defaults (of `let` being const and `let mut` being non-const.)

I think the first commit is probably unambiguously desirable.

jart (approved):

Issue #1219, opened 2024-06-16
testing github autolabeler label is too noisy


mrdomino:
Right now, the testing label gets applied every time a commit touches anything under `test/`. An incrementally better heuristic might be “this touched `test/`, and didn’t touch anything outside of `test/`.”

PR #1220, opened 2024-06-16
Update labeler.yml to remove testing label from the autolabeler 

mofosyne:
https://github.com/jart/cosmopolitan/issues/1219 had an issue with noisy testing label. Investigated if there is a syntax to make it more exclusive, but turns out there isn't one yet. So let's remove an manually add it in as needed.


mrdomino (approved):
LGTM

PR #1222, opened 2024-06-16
Use std::swap in string

mrdomino:
The STL spec says that swap is located in the string_view header anyway. Performance-wise this is a noop, but it’s slightly cleaner.


mrdomino:
Superseded.

PR #1227, opened 2024-06-16
CTL: utility.h, use ctl::swap in string

mrdomino:
```
* Add ctl utility.h

Implements forward, move, swap, and declval. This commit also adds a def
for nullptr_t to cxx.inc. We need it now because the CTL headers stopped
including anything from libc++, so we no longer get their basic types.

* Use ctl::swap in string

The STL spec says that swap is located in the string_view header anyawy.
Performance-wise this is a noop, but it’s slightly cleaner.
```

Supersedes #1222.


jart (approved):
hooray

PR #1231, opened 2024-06-18
Fix c.inc _Atomic define for C++

mrdomino:


mrdomino:
#### PR description (can be drag-select copied and pasted in github's UI)

c.inc (AFAICT erroneously) defined `_Atomic(t)` as \``volatile t *`\`, when it should have just said \``volatile t`\`, when `__STDC_VERSION__` was too small. This happens when we’re compiling C++, but in C++11, `_Atomic` is a define supplied by the STL rather than a keyword supplied by the compiler. Wait though, it gets better: in C++11, `_Atomic` hooks you into the morass that is stdatomic.h, and ultimately refers everything back to `std::atomic<T>`.

The gory, horrifying details are in libcxx's __atomic/cxx_atomic_impl.h. The tldr is that for our purposes it’s fine to just say volatile and use the normal libc/intrin/atomic.h functions.

mrdomino:
#1229 compiles with this change, and not without it.

jart (approved):

mrdomino:
From [cppreference on std::atomic][0]:

> Implementations are recommended to ensure that the representation of `_Atomic(T)` in C is same as that of `std::atomic<T>` in C++ for every possible type `T`. The mechanisms used to ensure atomicity and memory ordering should be compatible.

This gives us hope that it may be legitimate to go down this road…

[0]:  https://en.cppreference.com/w/cpp/atomic/atomic

PR #1237, opened 2024-06-29
wip ctl type traits

mrdomino:
I am a bit ambivalent about this. <type_traits> is actually not that bad by STL standards, “only” 148 transitive includes for the whole file. The split out variants under __type_traits are also mostly reasonable - e.g. __type_traits/is_same.h only includes 4 other STL files. Many others ask for more, but nothing is all that egregious.

I do think we can probably provide an 80–95% solution with a fraction of the footprint in CTL by for example eliding direct struct definitions in favor of the _t or _v variants when the latter are simpler to write, and omitting the more obscure tests in favor of a common core. I don’t think it’s worth splitting things out by file but it might be worth it to e.g. have a type_traits.h and a type_traits_advanced.h or something.

### DIFF
--- a/ctl/type_traits.cc
+++ b/ctl/type_traits.cc
@@ -1,0 +1,71 @@
+// -*- mode:c++; indent-tabs-mode:nil; c-basic-offset:4; coding:utf-8 -*-
+// vi: set et ft=cpp ts=4 sts=4 sw=4 fenc=utf-8 :vi
+//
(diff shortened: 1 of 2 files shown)


mrdomino:
Probably no longer makes sense to merge since we've gone in a different direction but sharing this in case some of the minor ideas (e.g. the directly-cribbed-from-cppreference is_convertible) are useful.

PR #1255, opened 2024-08-10
one-click-Win11-ps1-install

thedavidyoungblood:
with meRead(tm) documentation - by LouminAI Labs

### DIFF
--- a/one-click-cosmo.ps1
+++ b/one-click-cosmo.ps1
@@ -1,0 +1,171 @@
+# § - "meRead" - as a 'mini-Readme,' in the script itself - §
+
+# Header Metadata
+# **Script Name**: Cosmopolitan Libc Setup Script  
+# **Author**: LouminAI Labs - louminai.com  
+# **Creation Date**: 2024-08-09  
+# **Last Modified Date**: 2024-08-09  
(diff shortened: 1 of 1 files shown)


G4Vi:
Nice, looks like a convenient way to get the toolchain setup in Windows WSL.

The best place for this looks like it might be `tool/cosmocc/`, can you move it there and add a note to the WSL platform notes (https://github.com/jart/cosmopolitan?tab=readme-ov-file#wsl) that it's available to install?

Also have you sent the copyright assignment email to Justine? https://github.com/jart/cosmopolitan/blob/master/CONTRIBUTING.md#copyright-assignment

jart:
What is aka.ms?

G4Vi:
Microsoft seems to use them for shortened urls to their product downloads. I'm not sure if anyone else can create them. The urls in the script appear to be legit: The vs build tools link is referenced here https://learn.microsoft.com/en-us/visualstudio/install/build-tools-container?view=vs-2022 and the terminal link  here https://github.com/microsoft/terminal?tab=readme-ov-file#microsoft-store-recommended

thedavidyoungblood (commented):
Looks good. Thanks! 👍

PR #1266, opened 2024-08-28
Minor cleanup/improvements in unique_ptr_test

mrdomino:
I'd previously introduced a bunch of small wrappers around the class and functions under test to avoid excessive cpp use, but we can achieve this more expediently with simple using-declarations. This also cuts out some over-specified tests (e.g. there's no reason a stateful deleter wouldn't compile.)


jart (approved):
Thank you!

PR #1267, opened 2024-08-28
Basic CTL shared_ptr implementation

mrdomino:


mrdomino:
Supersedes #1229.

ctl/shared_ptr.h
```diff
@@ -0,0 +1,454 @@
...
+};
+
+static inline __attribute__((always_inline)) void
+incref(size_t* r) noexcept
+{
+#ifdef NDEBUG
+    __atomic_fetch_add(r, 1, __ATOMIC_RELAXED);
+#else
+    size_t refs = __atomic_fetch_add(r, 1, __ATOMIC_RELAXED);
+    if (refs > ((size_t)-1) >> 1)
```
jart, line 47: Is there any particular reason why you wouldn't just use `long` and then check if it's negative?

jart (approved):
Looks great. We know this algorithm is correct, but if you want to make extra sure, you could add a `pthread_create` torture test. I have lots of them in the codebase.

ctl/shared_ptr.h
mrdomino: You’d need `ckd_add` or something in that case - I believe signed integer overflow is UB

mrdomino: Oh wait, no, I see what you mean. Yeah, that could work.

PR #1271, opened 2024-08-31
Fix %[hl]s field length/precision on printf funcs

GabrielRavier:
Cosmopolitan currently has a completely wrong understanding (according to the C Standard) of how to handle precision and field length for wide strings, which have to be handled not by using the monospace display width but instead by using the outputted byte width (i.e. the UTF-8 length of the wide string).

This patch corrects this, corrects comments documenting the previous behavior, adds tests for the behavior and refactors tests that tested for the previous behavior to correctly use the new correct behavior.

### DIFF
--- a/libc/stdio/fmt.c
+++ b/libc/stdio/fmt.c
@@ -365,19 +365,24 @@ static int __fmt_stoa_byte(out_f out, void *a, uint64_t c) {
   return out(buf, a, 1);
 }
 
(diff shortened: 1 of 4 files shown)


GabrielRavier:
Discussion on Discord has yielded a rejection of the patch for implementing stupid behavior (it may be standard, but jart considers that "while I believe strongly in standard conformance, this is the one deviation from the standard I feel most strongly about, because not having this makes printf precision and width modifiers worthless"), so closing this now.

PR #1274, opened 2024-09-01
Fix shared_ptr::owner_before

mrdomino:
This method is supposed to give equivalence iff two shared pointers both own the same object, even if they point to different addresses. We can't control the exact order of the control blocks in memory, so the test can only check that this equivalence/non-equivalence relationship holds, and this is in fact all that it should check.


jart (approved):
Sweet. Thanks!

PR #1275, opened 2024-09-01
Fix printf funcs on memory pressure with floats

GabrielRavier:
Cosmopolitan's printf-family functions currently crash if one tries to
format a floating point number with a large precision (large enough
that gdtoa tries to allocate memory to format the number) when under
memory pressure (i.e. when malloc fails), because gdtoa fails to check
if malloc fails.

The added tests (which would previously crash under cosmopolitan
without this patch) show how to reproduce the issue.

This patch fixes this, and adds the aforementioned tests.

### DIFF
--- a/test/libc/stdio/snprintf_enomem_test.c
+++ b/test/libc/stdio/snprintf_enomem_test.c
@@ -1,0 +1,61 @@
+/*-*- mode:c;indent-tabs-mode:nil;c-basic-offset:2;tab-width:8;coding:utf-8 -*-│
+│ vi: set et ft=c ts=2 sts=2 sw=2 fenc=utf-8                               :vi │
+╞══════════════════════════════════════════════════════════════════════════════╡
+│ Copyright 2024 Gabriel Ravier                                                │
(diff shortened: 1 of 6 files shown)


test/libc/stdio/snprintf_enomem_test.c
```diff
@@ -0,0 +1,61 @@
...
+#include "libc/testlib/testlib.h"
+
+static void limit_memory_to_1mb() {
+  struct rlimit limit = {};
+  ASSERT_GE(getrlimit(RLIMIT_AS, &limit), 0);
+
+  if (limit.rlim_max > 1000000 || limit.rlim_max == RLIM_INFINITY) {
+    limit.rlim_max = 1000000;
+    limit.rlim_cur = limit.rlim_max;
+    ASSERT_GE(setrlimit(RLIMIT_AS, &limit), 0);
```
jart, line 33: You truly understand Cosmopolitan Libc if you're using this function.

jart (approved):
Nice! Thank you!

PR #1279, opened 2024-09-04
redbean: add tls socket lua binding

chamot1111:
Add TLS module for Lua using mbedTLS

- Implements TLS socket creation and management
- Supports connection, read, write, and close ops
- Handles memory safely with proper cleanup
- Provides error handling and status reporting
- Configurable SSL verification and timeouts

### DIFF
--- a/tool/net/definitions.lua
+++ b/tool/net/definitions.lua
@@ -4982,6 +4982,48 @@ unix = {
     X_OK = nil
 }
 
+---@class TlsContext
+---@field connect fun(self: TlsContext, server_name: string, server_port: string): boolean, string?
+---@field write fun(self: TlsContext, data: string): integer, string?
+---@field read fun(self: TlsContext, bufsiz?: integer): string?, string?
+---@field close fun(self: TlsContext)
+
+---@class tls
+local tls = {}
+
+--- Creates a new TLS socket.
+---@param verify? boolean Whether to verify the server's certificate (default: true)
+---@param timeout? integer Read timeout in milliseconds (default: 0, no timeout)
+---@return TlsContext|nil context
+---@return string? error
+function tls.socket(verify, timeout) end
+
+--- Connects to a server using TLS.
+---@param context TlsContext
+---@param server_name string
+---@param server_port string
+---@return boolean success
+---@return string? error
+function tls:connect(server_name, server_port) end
+
+--- Writes data to the TLS connection.
+---@param context TlsContext
+---@param data string
+---@return integer bytes_written
+---@return string? error
+function tls:write(data) end
+
+--- Reads data from the TLS connection.
+---@param context TlsContext
+---@param bufsiz? integer Maximum number of bytes to read (default: BUFSIZ)
+---@return string? data
+---@return string? error
+function tls:read(bufsiz) end
+
+--- Closes the TLS connection.
+---@param context TlsContext
+function tls:close() end
+
 --- Opens file.
 ---
 --- Returns a file descriptor integer that needs to be closed, e.g.
--- a/tool/net/ltls.inc
+++ b/tool/net/ltls.inc
@@ -1,0 +1,293 @@
+#include "libc/intrin/kprintf.h"
+static const char *const tls_meta = ":mbedtls";
+
+typedef enum {
+  TLS_STATE_INIT,
+  TLS_STATE_CONNECTED,
+  TLS_STATE_CLOSED
+} TlsConnectionState;
+
+typedef struct {
+  mbedtls_entropy_context entropy;
+  mbedtls_ctr_drbg_context ctr_drbg;
+  mbedtls_ssl_context ssl;
+  mbedtls_ssl_config conf;
+  mbedtls_net_context server_fd;
+  int ref;  // Reference to self in the Lua registry
+  TlsConnectionState connection_state;
+  char *read_buffer;
+  size_t read_buffer_size;
+} TlsContext;
+
+static TlsContext **checktls(lua_State *L) {
+  TlsContext **tls = (TlsContext **)luaL_checkudata(L, 1, tls_meta);
+  if (tls == NULL || *tls == NULL)
+    luaL_typeerror(L, 1, tls_meta);
+  return tls;
+}
+
+static int tls_gc(lua_State *L) {
+  TlsContext **tlsp = checktls(L);
+  TlsContext *tls = *tlsp;
+
+  if (tls) {
+    if (tls->connection_state != TLS_STATE_CLOSED) {
+      mbedtls_net_free(&tls->server_fd);
+      mbedtls_ssl_free(&tls->ssl);
+      mbedtls_ssl_config_free(&tls->conf);
+      mbedtls_ctr_drbg_free(&tls->ctr_drbg);
+      mbedtls_entropy_free(&tls->entropy);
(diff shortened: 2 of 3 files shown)


tool/net/ltls.inc
```diff
@@ -0,0 +1,293 @@
+#include "libc/intrin/kprintf.h"
```
jart, line 1: You can probably remove this line.

I'm also reasonably certain this doesn't need to be a .inc file. LuaFetch() had to be one because it's dug in like a tick with all the global variables defined in redbean.c. But I see that here you're going for a more conventional library API. For example, rather than checking the `sslclientverify` you're just accepting that as a parameter. That is probably the right thing to do, honestly.

So could you, if possible, make this ltls.c and update tool/net/BUILD.mk appropriately? Be sure to put the standard copyright header at the top.

tool/net/ltls.inc
```diff
@@ -0,0 +1,293 @@
...
+  if (ret > 0) {
+    lua_pushlstring(L, tls->read_buffer, ret);
+    return 1;
+  } else if (ret == 0) {
+    // End of file
+    lua_pushnil(L);
+    return 1;
+  } else {
+    lua_pushnil(L);
+    if (ret == MBEDTLS_ERR_SSL_WANT_READ || ret == MBEDTLS_ERR_SSL_WANT_WRITE) {
```
jart, line 217: Do you really support non-blocking i/o?

tool/net/ltls.inc
```diff
@@ -0,0 +1,293 @@
...
+      break;
+    default:
+      state_str = "unknown";
+  }
+
+  lua_pushfstring(L, "TLS connection (%p): %s", tls, state_str);
+  return 1;
+}
+
+static const struct luaL_Reg tls_methods[] = {{"connect", tls_connect},
```
jart, line 266: Please put the elements on their own lines.

tool/net/ltls.inc
```diff
@@ -0,0 +1,293 @@
...
+      state_str = "connected";
+      break;
+    case TLS_STATE_CLOSED:
+      state_str = "closed";
+      break;
+    default:
+      state_str = "unknown";
+  }
+
+  lua_pushfstring(L, "TLS connection (%p): %s", tls, state_str);
```
jart, line 262: With redbean what we like to do is have a `__repr` method so when an object is printed on the redbean shell, it'll print the code you'd type to create that object. Then the `__tostring` method will usually be the same thing as `__repr`.  I don't think it's worth showing state here, because the state you're showing is just the internal class state and might not reflect the current status in the OS of the socket.

tool/net/definitions.lua
```diff
@@ -4982,6 +4982,48 @@ unix = {
     X_OK = nil
 }
 
+---@class TlsContext
+---@field connect fun(self: TlsContext, server_name: string, server_port: string): boolean, string?
```
jart, line 4986: In terms of API design, have you considered this?

```lua
tls = require "tls"
conn = tls.TlsClient(ResolveIp("google.com"), 80) -- returns object of TlsClient class
conn.write("GET / HTTP/1.0\r\n\r\n")
print(conn.read())
conn.close()
```

That would be more similar and composable with other redbean APIs. For example, we like to pass IPs as uint32. You wouldn't need to maintain a state machine with this design. You could have functions or variables associated with the `tls` module for doing context-wide configuration, like whether or not SSL client verification should be enabled. You could also have that default to the redbean settings.

Another thing you might do that's even better is:

```lua
tls = require "tls"
unix = require "unix"

fd = assert(unix.socket(unix.AF_INET, unix.SOCK_STREAM, unix.IPPROTO_IP))
assert(unix.connect(fd, ResolveIp("google.com"), 80))
conn = assert(tls.TlsClient(fd)) -- returns object of TlsSocket class
assert(conn.write("GET / HTTP/1.0\r\n\r\n"))
response = assert(conn.read())
print(response)
assert(unix.close(fd))
```

I noticed you're drawing a lot of influence from mbedTLS's net_sockets.c API. I don't like that API. I don't think it's very good. If you use that abstraction, then you lose the ability to compose with redbean APIs like ResolveIp(), unix.poll(), etc. For examples of how I've made mbedTLS work with raw file descriptors, see `tool/curl/curl.c` and `tool/build/lib/eztls.c`.

jart (changes requested):

jeromew:
I add a reference to this fullmoon request - https://github.com/pkulchenko/fullmoon/issues/29#issuecomment-1826528990 that seems related to the exposure of more mbedtls endpoints (I don't know if this should belong to the tls module or somewhere else)

@pkulchenko listed the openssl APIs that we are currently missing in order to add automatic letsencrypt DNS-01 challenge like the one existing in the caddy web server.

it would be a great new redbean feature if the exposure of more mbedtls endpoints could make it possible to have automatic ssl certification embedded.

I copy them here for reference

> capturex('openssl', ('genrsa', '-out', $_, KEY_SIZE));
> capturex('openssl', ('rsa', '-text', '-in', $self->{domain}{account}, '-noout', '-modulus'));
> capturex('openssl', ('rsa', '-in', $self->{domain}{account}, '-pubout')));
> capturex('openssl', ('dgst', '-sha256', '-binary', '-sign', $self->{domain}{account}, $stf->filename))
> capturex('openssl', ('req', '-new', '-outform', 'DER', '-key', $self->{domain}{key}, '-config', $oct->filename, '-out', $self->{req}{csr}));

tool/net/definitions.lua
chamot1111: This API indeed seems more natural. I will make the modifications to get as close as possible to this example. Thank you

tool/net/ltls.inc
chamot1111: No I don't, i remove it

tool/net/ltls.inc
chamot1111: Ok. I will use:

``` C
static int tls_tostring(lua_State *L) {
  TlsContext **tlsp = checktls(L);
  TlsContext *tls = *tlsp;

  lua_pushfstring(L, "tls.TlsClient(fd=%d)", tls->fd);
  return 1;
}

static const struct luaL_Reg tls_methods[] = {
    {"write", tls_write},
    {"read", tls_read},
    {"close", tls_close},
    {"__gc", tls_gc},
    {"__tostring", tls_tostring},
    {"__repr", tls_tostring},
    {NULL, NULL}
};
```

tool/net/definitions.lua
chamot1111: use:

``` lua

---@class tls
local tls = {}

--- Creates a new TLS client.
---@param fd integer File descriptor of the socket
---@param verify? boolean Whether to verify the server's certificate (default: true)
---@param timeout? integer Read timeout in milliseconds (default: 0, no timeout)
---@return TlsContext|nil context
---@return string? error
function tls.TlsClient(fd, verify, timeout) end

--- Writes data to the TLS connection.
---@param context TlsContext
---@param data string
---@return integer bytes_written
---@return string? error
function tls:write(data) end

--- Reads data from the TLS connection.
---@param context TlsContext
---@param bufsiz? integer Maximum number of bytes to read (default: BUFSIZ)
---@return string? data
---@return string? error
function tls:read(bufsiz) end
```

PR #1282, opened 2024-09-07
Fix ecvt/fcvt issues w.r.t. value==0 and ndigit==0

GabrielRavier:
Before this commit, cosmopolitan had some issues with handling arguments
of 0 and signs, such as returning an incorrect sign when the input value
== -0.0, and incorrectly handling ndigit == 0 on fcvt (ndigit determines
the amount of digits *after* the radix character on fcvt, thus the parts
before it still must be outputted before fcvt's job is completely done).

This patch fixes these issues, and adds tests with corresponding inputs.

### DIFF
--- a/libc/stdio/ecvt.c
+++ b/libc/stdio/ecvt.c
@@ -22,6 +22,7 @@
 │ Materiel Command, USAF, under agreement number F39502-99-1-0512.             │
 │ SUCH DAMAGE.                                                                 │
 ╚─────────────────────────────────────────────────────────────────────────────*/
+#include "libc/math.h"
 #include "libc/mem/mem.h"
 #include "libc/runtime/runtime.h"
 #include "libc/str/str.h"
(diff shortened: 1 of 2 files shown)


jart (approved):
These functions aren't part of the C standard anymore. I agree it's worth it to have them be more correct and consistent with glibc. What you discovered is also a bug in OpenbSD. I recommend sending them a patch, so they'll know Cosmo is helping them. It's also a bug in Apple Libc. OTOH FreeBSD has removed these APIs.

GabrielRavier:
w.r.t. glibc consistency I was actually only really trying to make changes that I could back up with a reasonable interpretation of POSIX-2004 (i.e. discarding cases that are either deliberately implementation defined (e.g. `*decpt` can be 0 or 1 when `value == 0`) or UB (e.g. `ndigit < 0` seems to be UB)) but if you think it'd be better to try to make it behave exactly like glibc I could also do that (well, *exactly* like glibc would be going clearly too far though, given glibc's implementation is actually buggy in some respects: https://sourceware.org/bugzilla/show_bug.cgi?id=29556)

GabrielRavier:
I'll also try to make a bug report to OpenBSD (though that'll take a bit given I'll want to setup an OpenBSD VM to do it "properly") but do you know how to report bugs to Apple for libc at all ? (if there's a way to do that, especially given I don't have any Apple computer in the first place...)

PR #1283, opened 2024-09-07
Fix printf-family long double prec/rounding issues

GabrielRavier:
Currently, in cosmopolitan, there is no handling of the current rounding
mode for long double conversions, such that round-to-nearest gets always
used, regardless of the current rounding mode. %Le also improperly calls
gdtoa with a too small precision (which led to relatively similar bugs).

This patch fixes these issues, in particular by modifying the FPI object
passed to gdtoa such that it is modifiable (so that __fmt can adjust its
rounding field to correspond to FLT_ROUNDS (note that this is not needed
for dtoa, which checks FLT_ROUNDS directly)) and ors STRTOG_Neg into the
kind field in both of the __fmt_dfpbits and __fmt_ldfpbits functions, as
the gdtoa function also depends on it to be able to accurately round any
negative arguments. The change to kind also requires a few other changes
to make sure kind's upper bits (which include STRTOG_Neg) are masked off
when attempting to only examine the lower bits' value. Furthermore, this
patch also makes exactly one change in gdtoa, which appears to be needed
to fix rounding issues with FE_TOWARDZERO (this seems like a gdtoa bug).

The patch also adds a few tests for these issues, along with also taking
the opportunity to clean up some of the previous tests to do the asserts
in the right order (i.e. with the first argument as the expected result,
and the second one being used as the value that it is compared against).

### DIFF
--- a/libc/stdio/fmt.c
+++ b/libc/stdio/fmt.c
@@ -53,6 +53,7 @@
 #include "libc/math.h"
 #include "libc/mem/mem.h"
 #include "libc/mem/reverse.internal.h"
+#include "libc/runtime/fenv.h"
 #include "libc/runtime/internal.h"
 #include "libc/serialize.h"
 #include "libc/str/str.h"
@@ -90,7 +91,7 @@
 
 struct FPBits {
   uint32_t bits[4];
-  const FPI *fpi;
+  FPI fpi;
   int sign;
   int ex;  // exponent
   int kind;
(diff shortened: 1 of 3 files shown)


jart (approved):
Heck yes! We pay a heavy cost for generalized dtoa. It's nice to know we're getting the benefits of full generality. One of these days I'm hoping I'll discover a use case for these other rounding modes. When that day comes, it makes me happy I can count on cosmo more to work, thanks to this change.

PR #1284, opened 2024-09-08
Fix rare corner case in ntspawn.c

jeromew:
There could be cases when libc/proc/describefds.c calls calloc(handlecount, ..) with handlecount = 0
to initialize  what will later become opt_lpExplicitHandleList. But calloc(0, ..) does not always return null
according to https://learn.microsoft.com/ka-ge/cpp/c-runtime-library/reference/calloc?view=msvc-140

> In the Microsoft implementation, if number or size is zero, calloc returns a pointer to an allocated block
> of non-zero size. An attempt to read or write through the returned pointer leads to undefined behavior.

If the chosen conditional is a->opt_lpExplicitHandleList then UpdateProcThreadAttribute is called and
produces a Windows error. The modification avoids this problem.

### DIFF
--- a/libc/calls/ntspawn.c
+++ b/libc/calls/ntspawn.c
@@ -166,7 +166,7 @@ static textwindows int ntspawn2(struct NtSpawnArgs *a, struct SpawnBlock *sb) {
         alist, 0, kNtProcThreadAttributeParentProcess, &a->opt_hParentProcess,
         sizeof(a->opt_hParentProcess), 0, 0);
   }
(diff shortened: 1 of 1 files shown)


jart (approved):
Good eye. LGTM. Please confirm the edit I made to your PR is correct, then I'll merge.

jeromew:
LGTM. I should have spotted this. Thanks for the review

jart:
Commit message edited to follow project style.

- first line should be 50 characters or fewer
- paragraphs must be justified to exactly 72 characters without inserting extra spaces

PR #1285, opened 2024-09-08
Add POSIX's apostrophe flag for printf-based funcs

GabrielRavier:
POSIX specifies the <apostrophe> flag character for printf as formatting
decimal conversions with the thousands' grouping characters specified by
the current locale. Given that cosmopolitan currently has no support for
obtaining the locale's grouping character, all that is required (when in
the C/POSIX locale) for supporting this flag is ignoring it, and as it's
already used to indicate quoting (for non-decimal conversions), all that
has to be done is to avoid having it be an alias for the <space> flag so
that decimal conversions don't accidentally behave as though the <space>
flag has also been specified whenever the <apostrophe> flag is utilized.

This patch adds this flag, as described above, along with a test for it.

### DIFF
--- a/libc/stdio/fmt.c
+++ b/libc/stdio/fmt.c
@@ -78,7 +78,7 @@
 #define FLAGS_ISSIGNED  0x40
 #define FLAGS_NOQUOTE   0x80
 #define FLAGS_REPR      0x100
-#define FLAGS_QUOTE     FLAGS_SPACE
+#define FLAGS_QUOTE     0x200
 #define FLAGS_GROUPING  FLAGS_NOQUOTE
 
 #define __FMT_PUT(C)              \
--- a/test/libc/stdio/snprintf_test.c
+++ b/test/libc/stdio/snprintf_test.c
@@ -250,3 +250,10 @@ TEST(snprintf, testAConversionSpecifier) {
   ASSERT_STREQ("0x1.8p+4", buf);
 }
 */
+
+TEST(snprintf, apostropheFlag) {
+  char buf[20];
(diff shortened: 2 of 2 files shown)


jart (changes requested):
We do support grouping. You can say `%,d` for `1,000,000`. Could you please change `%'d` so it formats `1'000'000`? Then afterwards we should probably look into importing Musl's localeconv() function so we can access its grouping field. I want the default behavior for Cosmopolitan Libc to be using that apostrophe since it's culture agnostic and I have a disability that makes me allergic to calling setlocale() and unable to read large numbers without grouping. Obviously though if the locale actually specifies a grouping (e.g. English, German) then we should use what it specifies. If you decide to do this (I will if you won't) then please take care to ensure that obtaining the current locale grouping doesn't link any locale code. It should only need to access a global variable through TLS I think.

GabrielRavier:
Well, given that in the C/POSIX locale (the mandatory default before `setlocale` is ever called) the thousands grouping character is nothing (as specified in the *LC_NUMERIC in the POSIX Locale* section of the POSIX standard), I'm pretty sure using `'` as the separator without being in the context of an appropriate locale is a completely and utterly non-standard implementation of this flag. I guess I'll implement it like that if you want (when I get the time, so perhaps not today) but I would just like to say that it is 100% a POSIX violation.

GabrielRavier:
I've added a commit to make this change.

jart:
I can see that you view this as something important and the standard is on your side. In that case, I'm convinced. Could you please update this PR to go back to the conformant behavior you originally chose?

GabrielRavier:
I have just done so.

jart (approved):

PR #1286, opened 2024-09-10
Partially fix printf hex float numbers/%a rounding

GabrielRavier:
Hexadecimal printing of floating-point numbers on cosmopolitan (that is,
using the a conversion specifier) currently does not correctly round the
result in any rounding mode except for the default one (i.e. FE_NEAREST)

This commit fixes this, and adds tests for the change (note that there's
still some rounding issues with the a conversion specifier in general in
relatively rare cases (that is, without non-default rounding modes), but
this author gave up on fixing them after they tried to read the floating
point formatting code, and wounded up almost more confused than before).

### DIFF
--- a/libc/stdio/fmt.c
+++ b/libc/stdio/fmt.c
@@ -690,26 +690,45 @@ static int __fmt_fpiprec(struct FPBits *b) {
 // prec1 = incoming precision (after ".")
 static int __fmt_bround(struct FPBits *b, int prec, int prec1) {
   uint32_t *bits, t;
-  int i, inc, j, k, m, n;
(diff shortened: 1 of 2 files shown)


jart (approved):
Thank you for doing this. That floating point code was inherited from Lucent. I wouldn't have been able to implement `__fmt()` without their help, but their code has always been ugly, and it used to fail a lot more things. Since adopting it, I've made it even uglier by smishing and smashing it into passing more tests. My dream is one day I'll get good enough at floating point code to rewrite it all. But that day is not today.

jart:
Commit message edited to communicate the improvement rather than the problem.

PR #1287, opened 2024-09-13
Hopefully completely fix printf-family %a rounding

GabrielRavier:
The a conversion specifier to printf had some issues w.r.t. rounding, in
particular in edge cases w.r.t. "to nearest, ties to even" rounding (for
instance, "%.1a" with 0x1.78p+4 outputted 0x1.7p+4 instead of 0x1.8p+4).

This patch fixes this and adds several tests w.r.t ties to even rounding

### DIFF
--- a/libc/stdio/fmt.c
+++ b/libc/stdio/fmt.c
@@ -565,12 +565,12 @@ static int __fmt_stoa(int out(const char *, void *, size_t), void *arg,
 static void __fmt_dfpbits(union U *u, struct FPBits *b) {
   int ex, i;
   b->fpi = kFpiDbl;
-  // Uncomment this if needed in the future - we currently do not need it, as
-  // the only reason we need it in __fmt_ldfpbits is because gdtoa reads
-  // fpi.rounding to determine rounding (which dtoa does not need as it directly
-  // reads FLT_ROUNDS)
-  // if (FLT_ROUNDS != -1)
-  //   b->fpi.rounding = FLT_ROUNDS;
+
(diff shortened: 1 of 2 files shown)


libc/stdio/fmt.c
```diff
@@ -701,22 +704,23 @@ static int __fmt_bround(struct FPBits *b, int prec, int prec1) {
...
   // doesn't need to be set as we have already done so above)
-  // The last one handles rounding to nearest
-  current_rounding_mode = fegetround();
-  if (current_rounding_mode == FE_TOWARDZERO ||
-      (current_rounding_mode == FE_UPWARD && b->sign) ||
-      (current_rounding_mode == FE_DOWNWARD && !b->sign))
+  // They use the FLT_ROUNDS value - i.e. 0 -> toward zero, 1 -> to nearest,
+  // ties to even, 2 -> toward positive infinity, 3 -> towards negative infinity
+  // The last if handles rounding to nearest
+  if (b->fpi.rounding == 0 || (b->fpi.rounding == 2 && b->sign) ||
```
jart, line 710: Wouldn't it be better to say `FPI_Round_zero` and `FPI_Round_up` here?

jart (changes requested):
This is impressive. Could you avoid the magic numbers?

libc/stdio/fmt.c
GabrielRavier: This commit now uses the `FPI_Round_*` values.

GabrielRavier:
The commit now uses the `FPI_Round_*` enum values as appropriate.

jart (approved):

PR #1289, opened 2024-09-13
Fix occasional crash in test/libc/intrin/mmap_test

GabrielRavier:
This test would sometimes crash due to the EZBENCH2() macro occasionally
running the first benchmark (BenchMmapPrivate()) less times than it does
the second benchmark (BenchUnmap()) - this would then lead to a crash in
BenchUnmap() because BenchUnmap() expects that BenchMmapPrivate() has to
previously have been called at least as many times as it has itself such
that a region of memory has been mapped, for BenchUnmap() to then unmap.

This commit fixes this by utilizing the newer BENCHMARK() macro (instead
of the EZBENCH2() macro) which runs the benchmark using an count of runs
specified directly by the benchmark itself, which allows us to make sure
that the two benchmark functions get ran the exact same amount of times.

### DIFF
--- a/test/libc/intrin/mmap_test.c
+++ b/test/libc/intrin/mmap_test.c
@@ -32,7 +32,7 @@
 #include "libc/sysv/consts/msync.h"
 #include "libc/sysv/consts/o.h"
 #include "libc/sysv/consts/prot.h"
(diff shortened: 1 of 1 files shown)


jart (approved):
Thank you.

PR #1290, opened 2024-09-13
Make the pledge sandbox .so object work with UBSAN

GabrielRavier:
Currently, cosmopolitan's pledge sandbox .so shared object wrongly tries
to use a bunch of UBSAN symbols, which are not defined when outside of a
cosmopolitan-based context (save if the sandboxed binary also happens to
be itself using UBSAN, but that's obviously very commonly not the case).

Fix this by making it such that the sandbox .so shared object traps when
UBSAN is triggered, avoiding any attempt to call into the UBSAN runtime.


jart (approved):
That's a cool flag I hadn't heard of. LGTM.

PR #1291, opened 2024-09-13
Fix statx not being allowed on rpath/wpath pledges

GabrielRavier:
While always blocking statx did not lead to particularly bad results for
most cases (most code that uses statx appears to utilize a fallback when
statx is unavailable), it does lead to using usually far less used (thus
far less well tested) code: for example, musl's current fstatat fallback
for statx fails to set any values for stx_rdev_major and stx_rdev_minor,
which the raw syscall wouldn't (I've have sent a patch to musl for this,
but this won't fix older versions of musl and binaries/OSes using them).
Along with the fact that statx extends stat in several useful ways, this
seems to indicate it is far better to simply allow statx whenever pledge
also allows stat-family syscalls, i.e. for both rpath and wpath pledges.

### DIFF
--- a/libc/calls/pledge-linux.c
+++ b/libc/calls/pledge-linux.c
@@ -712,6 +712,7 @@ static const uint16_t kPledgeRpath[] = {
 #endif                             //
     __NR_linux_fstat,              //
     __NR_linux_fstatat,            //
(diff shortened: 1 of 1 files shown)


jart (approved):
LGTM

PR #1294, opened 2024-09-15
Fix unicode look-alike in define

mierenhoop:


jart (approved):
Thanks. My fingers sometimes slip and do that.

PR #1296, opened 2024-09-15
Fix printing decimal-point character on printf %#a

GabrielRavier:
The C standard indicates that when processing the a conversion specifier
"if the precision is zero *and* the # flag is not specified, no decimal-
point character appears.". This means that __fmt needs to ensure that it
prints the decimal-point character not only when the precision is non-0,
but also when the # flag is specified - cosmopolitan currently does not.

This patch fixes this, along with adding a few tests for this behaviour.


jart (approved):
LGTM. Thank you.

PR #1297, opened 2024-09-16
Fix 0 before decimal-point in hex float printf fns

GabrielRavier:
The C standard specifies that, upon handling the a conversion specifier,
the argument is converted to a string in which "there is one hexadecimal
digit (which is nonzero [...]) before the decimal-point character", this
being a requirement which cosmopolitan does not currently always handle,
sometimes printing numbers like "0x0.1p+5", where a correct output would
have been e.g. "0x1.0p+1" (despite both representing the same value, the
first one illegally has a '0' digit before the decimal-point character).

This patch fixes this and adds tests which would previously have failed.


jart (approved):
Thank you!

PR #1298, opened 2024-09-16
Make printf %La test-case work properly on AArch64

GabrielRavier:
As the size of long double changes between x86 and AArch64, this results
in one of the printf a conversion specifier test-cases getting different
output between the two. Note that both outputs (and a few more) are 100%
standards-conforming, but the testcase currently only expects a specific
one - the one that had been seen on x86 when initially writing the test.

This patch fixes the testcase so it accepts either of those two outputs.


GabrielRavier:
Ping ?

jart (approved):
Sorry about the delay. LGTM

PR #1299, opened 2024-09-17
Add weak self make_shared variant

mrdomino:
This extends the CTL version of make_shared with functionality not found
in the STL, with inspiration taken from Rust's Rc class.


jart (approved):
Looks cool. Thank you!

PR #1300, opened 2024-09-22
Add the uppercase B conversion specifier to printf

GabrielRavier:
The (uppercase) B conversion specifier is specified by the C standard to
have the same behavior as the (lowercase) b conversion specifier, except
that whenever the # flag is used, the (uppercase) B conversion specifier
alters a nonzero result by prefixing it with "0B", instead of with "0b".

This commit adds this conversion specifier alongside a few tests for it.


jart (approved):
Thank you!

PR #1318, opened 2024-10-26
Add several NT functions

bullno1:
With these addtions, I could build and run a [sokol](https://github.com/floooh/sokol) application (using OpenGL) on both Linux and Windows.


jart (approved):
Thank you!

jart:
Could you please see CONTRIBUTING.md and send the email? Then I can merge. Thanks!

PR #1319, opened 2024-10-29
Share that APE files are also zip archives and how to use them!

cdrubin:
Added some notes about how APE files are zips and can contain application support files


jart (approved):

jart:
Thank you!

PR #1324, opened 2024-11-18
feat: [redbean] add details to OnError Hook

solisoft:
The details of the error was missing, this PR add details to the OnError hook so we can now get why the error occurs


jart (approved):
Thank you!

PR #1331, opened 2024-11-29
Fix atomic_fetch_sub on workers

mrdomino:
clangd was showing a diagnostic for this line.


mrdomino:
(I'm kinda confused about what LockInc / LockDec are doing, and why they're doing a type cast - but this change follows suit with them.)

jart (approved):
TY

PR #1332, opened 2024-12-02
redbean: concurrency improvements

mrdomino:
In the course of playing with redbean I was confused about how the state
was behaving and then noticed that some stuff is maybe getting edited by
multiple processes. I tried to improve things by changing the definition
of the counter variables to be explicitly atomic. Claude assures me that
most modern Unixes support cross-process atomics, so I just went with it
on that front.

I also added some mutexes to the shared state to try to synchronize some
other things that might get written or read from workers but couldn't be
made atomic, mainly the rusage and time values. I could've probably been
less granular and just had a global shared-state lock, but I opted to be
fairly granular as a starting point.

This also reorders the resetting of the lastmeltdown timespec before the
SIGUSR2 signal is sent; hopefully this is okay.

### DIFF
--- a/tool/net/redbean.c
+++ b/tool/net/redbean.c
@@ -182,11 +182,9 @@ __static_yoink("blink_xnu_aarch64");    // is apple silicon
 #define HeaderEqualCase(H, S) \
   SlicesEqualCase(S, strlen(S), HeaderData(H), HeaderLength(H))
 #define LockInc(P)                                            \
-  atomic_fetch_add_explicit((_Atomic(typeof(*(P))) *)(P), +1, \
-                            memory_order_relaxed)
+  atomic_fetch_add_explicit(P, +1, memory_order_relaxed)
 #define LockDec(P)                                            \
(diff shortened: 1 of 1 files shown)


tool/net/redbean.c
```diff
@@ -7346,6 +7376,14 @@ void RedBean(int argc, char *argv[]) {
            (shared = mmap(NULL, ROUNDUP(sizeof(struct Shared), getgransize()),
                           PROT_READ | PROT_WRITE, MAP_SHARED | MAP_ANONYMOUS,
                           -1, 0)));
+  pthread_mutexattr_t attr;
+  unassert(!pthread_mutexattr_init(&attr));
+  unassert(!pthread_mutexattr_setpshared(&attr, PTHREAD_PROCESS_SHARED));
```
jart, line 7381: Nice. The time for this has come, now that `PTHREAD_PROCESS_SHARED` is pretty much finally fully supported.

PR #1344, opened 2024-12-27
replace strings in ELF .rodata sections

ahgamut:
With `tool/build/renamestr` (or an equivalent feature), I hope to avoid the `/zip` folder when building binaries with https://github.com/ahgamut/superconfigure. Most of the folder locations involving `/zip` are stored in the `.rodata` section of the ELF binary, and `/zip` is more useful for the APE than the debug ELF binaries.

 The current build process on `superconfigure` involves:

- create a `/zip` folder using sudo
- any config data that a binary requires (`terminfo` entries, SSL certs) is copied to `/zip`
- debug binaries use the actual `/zip` folder (or crash)

With this PR, the build process hopefully becomes:

- no need to create a `/zip` folder with sudo
- install config data to the usual location( `$LOCATION/superconfigure/cosmos/$ARCH/$SUBFOLDER`)
- run `renamestr` on the ELF binary to replace `$LOCATION/superconfigure/cosmos/$ARCH/` to `/zip`, before running `apelink`
- now debug binaries will work like regular ELF binaries built on a Linux system

Perhaps this feature can be part of `apelink` itself (the structure and some code of `renamestr` is copied from `apelink`), but I wrote a separate tool in a draft PR, for an initial exploration.


ahgamut:
Have confirmed that `renamestr` avoids the use of `/zip` for building `bash` and `berry` in superconfigure:

image

(There is a build error relating to the installation of `terminfo` entries, but I assume that is due my system interfering with the `ncurses` build process).

tool/build/renamestr.c
```diff
@@ -0,0 +1,469 @@
...
+      params.n++;
+    }
+  }
+  if (params.n == 0) {
+    Die(prog, "no replacements provided");
+  }
+  if (partial) {
+    Die(prog, "partial replacement provided");
+  }
+  for (int i = 0; i < params.n; ++i) {  // no-op
```
jart, line 185: If this code does nothing then delete it. The check has already been performed.

tool/build/renamestr.c
```diff
@@ -0,0 +1,469 @@
...
+  if (optind == argc) {
+    Die(prog, "missing input argument");
+  }
+  if (optind != argc - 1) {
+    Die(prog, "too many args");
+  }
+  exepath = argv[optind];
+}
+
+static void ValidateElfImage(Elf64_Ehdr *e, Elf64_Off esize,  //
```
jart, line 202: Don't bother validating. Your program doesn't even use program headers. You only care about sections. Some of these restrictions, like that it has to be a non-dynamic executable, don't make sense. Since your program could generalize to object files too.

tool/build/renamestr.c
```diff
@@ -0,0 +1,469 @@
...
+  }
+}
+
+struct Param {
+  const char *from_string;
+  size_t from_len;
+  const char *to_string;
+  size_t to_len;
+  int count;
+  char *roloc;
```
jart, line 120: It's a little weird to have multiple purposes in this same data structure. Normally I'd assume a data structure for storing input parameters would be immutable. But this struct is almost acting like a job object.

tool/build/renamestr.c
```diff
@@ -0,0 +1,469 @@
...
+  if (!IsElf64Binary(input.elf, input.size))
+    Die(path, "not an elf64 binary");
+  exefd = fd;
+}
+
+static void ReplaceString(struct Param *param) {
+  Elf64_Xword len = strnlen(param->roloc, roend - param->roloc);
+  char *targstr = Malloc(len + 1);
+  for (Elf64_Xword i = 0; i < len + 1; ++i)
+    targstr[i] = 0;
```
jart, line 397: This could write out of bounds if the last string in a section isn't nul-terminated.

tool/build/renamestr.c
```diff
@@ -0,0 +1,469 @@
...
+  strcpy(targstr, param->to_string);
```
jart, line 398: Why not just say:

```c
if (param->to_len)
  memcpy(param->roloc, param->to_string, param->to_len);
memmove(param->roloc + param->to_len, param->roloc + param->from_len, len + 1 - param->from_len);
```

Then you don't need malloc.

tool/build/renamestr.c
```diff
@@ -0,0 +1,469 @@
...
+
+  if (!prog)
+    prog = "renamestr";
+
+  GetOpts(argc, argv);
+  OpenInput(exepath);
+  ValidateElfImage(input.elf, input.size, input.path);
+  rodata = FindElfSectionByName(
+      input.elf, input.size,
+      GetElfSectionNameStringTable(input.elf, input.size), ".rodata");
```
jart, line 423: Run `o//tool/decode/elf o//tool/build/apelink.o | less` and notice how object files have sections like:

```
.Lsh10: .long   101                     // ".rodata.str1.1" == GetElfString(elf, st->st_size, shstrtab, shdr->sh_name)
        .long   SHT_PROGBITS            // shdr->sh_type
        .long   SHF_ALLOC|SHF_MERGE|SHF_STRINGS // shdr->sh_flags
        .quad   0                       // shdr->sh_addr
        .quad   0x41a0                  // shdr->sh_offset
        .quad   0x282                   // shdr->sh_size
        .long   0                       // shdr->sh_link
        .long   0                       // shdr->sh_info
        .quad   0x1                     // shdr->sh_addralign
        .quad   0x1                     // shdr->sh_entsize
```

Those `SHF_STRINGS` sections would be the safest place to do this transform. But you're replacing strings on the .rodata section of a fully linked binary. In that case, it's probably not even worth bothering with ELF at all. You would probably be just as well off treating the file as an opaque binary object and then just using your replace string algo on that.

ahgamut: I'm replacing the strings only after linking the `dbg` binary, running it a few times, right before running `apelink` to create the APE. 

When building projects likes `emacs` or `python` in [`superconfigure`](https://github.com/ahgamut/superconfigure), the installation steps involve running the built binary and checking folder locations. I run `renamestr` after the installation is complete, so that `/zip` is used only for the APE and not the debug binary.

ahgamut: > You would probably be just as well off treating the file as an opaque binary object and then just using your replace string algo on that.
Hmm, the first idea for this was a perl/sed script that would patch binary files in-place. If the ELF checks are not needed, perhaps we can use some other external tool? 

Alternatively, perhaps `renamestr` can be a flag/function added to `apelink`?

tool/build/renamestr.c
ahgamut: Hmm, each from/to combination needs to go over the whole file -- so we have two strings, the `roloc` refers to current location within the file, and `count` provides number of replacements performed.

Perhaps the `roloc` and `count` parameters can be in a separate struct?

ahgamut: Or the `roloc` can be a global, reset for each iteration.

tool/build/renamestr.c
ahgamut: I used `targstr` more for debug printing, perhaps it can be removed now.

tool/build/renamestr.c
```diff
@@ -0,0 +1,469 @@
...
+  const char *p, *e;
+  for (p = data, e = p + size; p < e; p += (size_t)rc, offset += (size_t)rc) {
+    if ((rc = pwrite(exefd, p, e - p, offset)) == -1) {
+      DieSys(exepath);
+    }
+  }
+}
+
+struct Param {
+  const char *from_string;
```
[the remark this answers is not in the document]
ahgamut: I can re-compute the `len` fields with strlen at each call of `ReplaceString`, but thought it was better to calculate once, at initialization.

tool/build/renamestr.c
ahgamut: Can this still happen with the `memmove` usage?

jart (approved):

Issue #1346, opened 2025-01-03
Readme need be updated to using make instead of build/bootstrap/make


mingodad:
After doing a `git pull` and trying to build following the instructions on `README` I found that the `build/bootstrap/make` doesn't exists anymore and have been moved to the root folder but the `README` still points to `build/bootstrap/make`.

mingodad:
Also when making on Ubuntu-18.04 I'm getting this error:
```
make
.cosmocc/3.9.2/bin/x86_64-linux-cosmo-ld.bfd: o//libc/intrin/intrin.a(sigblock.o):cosmopolitan/libc/intrin/sigblock.c:33: multiple definition of `__sig'; o//libc/intrin/intrin.a(sigvar.o):cosmopolitan/libc/intrin/sigvar.c:21: first defined here

`make MODE= -j8 o//dsp/prog/recvaudio.dbg` exited with 1:
[...]

PR #1353, opened 2025-01-06
Standard make path

mrdomino:
Modifies download-cosmocc.sh to maintain a .cosmocc/current symlink that
always points to the most recently downloaded version of cosmocc. We can
use this to point at a canonical make for a bootstrapped repository. For
first-time builds, we suggest: https://cosmo.zip/pub/cosmos/bin/make and
have updated the docs in a few places to mention this.

Fixes the other part of #1346.


jart (approved):
Looks good.

PR #1358, opened 2025-01-24
Adds some NT functions

A2va:
Adds some NT functions for [xmake](https://github.com/xmake-io/xmake/)
I have a few questions because I'm not sure if what I've done is right:

- Where do I put `cpinfoex.h`?
- Are the constants in `cpinfoex.h` in the right place?
- A previous PR #1318 did not include a prototype for all the functions that were added, was this an oversight?
    If not, I'm a bit confused because all the other NT functions I've looked at have a prototype somewhere.


libc/nt/struct/cpinfoex.h
```diff
@@ -0,0 +1,17 @@
+#ifndef COSMOPOLITAN_LIBC_NT_STRUCT_CPINFOEX_H_
+#define COSMOPOLITAN_LIBC_NT_STRUCT_CPINFOEX_H_
+
+#define MAX_LEADBYTES    12
```
jart, line 4: I recommend just inlining these values and not defining constants for them. Definitions are only useful if they can be changed. But these numbers are a permanent part of this ABI. Do you actually need them in your source code? If so, possibly put them in windowsesque.h? Please note `PATH_MAX` in cosmo is defined as 1024 and our `__mkntpath()` function will actually insert `\\?\` when absolutely necessary to overcome the 260 character limit.

libc/isystem/windowsesque.h
```diff
@@ -1420,6 +1420,15 @@
 #define HKEY_CURRENT_CONFIG kNtHkeyCurrentConfig
 #define HKEY_DYN_DATA kNtHkeyDynData
 #define HKEY_CURRENT_USER_LOCAL_SETTINGS kNtHkeyCurrentUserLocalSettings
+#define KEY_QUERY_VALUE kNtKeyQueryValue
```
jart, line 1423: Don't you want to add `#include`s for your new headers to the top of this file?

jart (changes requested):
Looks good. A few suggestions.

jart:
Also take a look at the test failures.

libc/nt/struct/cpinfoex.h
A2va: I use them in my source code but I finally choose to inline them because I noticed that `MAX_PATH` was inlined in `processentry32.h` so for the sake of consistency I did the same.

jart (approved):
TY

PR #1365, opened 2025-02-08
cosmocc: build toolchain on MacOS

bjia56:
This updates the cosmocc toolchain packaging script to work on MacOS. Tested on GitHub Actions macos-13 (x86_64) and macos-14 (arm64) runners and verified to still work on Ubuntu (GitHub Actions runners ubuntu-24.04 and ubuntu-24.04-arm). End goal is to bring cosmocc to MacPorts by running the packaging script.

`gmake` is used instead of `make` because both Homebrew and MacPorts install newer `make`'s with the `g` prefix.

Additionally, the APE loader from the bootstrapper toolchain is used instead of a system APE, which may not be available.

### DIFF
--- a/tool/cosmocc/package.sh
+++ b/tool/cosmocc/package.sh
@@ -54,10 +54,12 @@ ARM64=${3:-aarch64}
 NPROC=$(($(_nproc)/2))
 GCCVER=14.1.0
 MAKE=$(which_make)
-APE=$(which_ape)
 
 $MAKE -j$NPROC m= \
   $APELINK
+
+APE=$(which_ape)
+$APE
 
 $MAKE -j$NPROC m=$AMD64 \
   o/cosmocc.h.txt \
--- a/tool/cosmocc/package.sh
+++ b/tool/cosmocc/package.sh
@@ -31,22 +31,6 @@ which_make() {
 
 TMPDIR=${TMPDIR:-/tmp}
 
-which_ape() {
-  case $(uname -s) in
-    Darwin)
-      case $(mode) in
-        aarch64)
-          cc -O -o "$TMPDIR/ape.$$" .cosmocc/current/bin/ape-m1.c || exit
-          trap 'rm "$TMPDIR/ape.$$"' EXIT
-          echo $TMPDIR/ape.$$
-        ;;
-        *) echo .cosmocc/current/bin/ape-x86_64.macho ;;
-      esac
-      ;;
-    *) echo .cosmocc/current/bin/ape-$(uname -m).elf ;;
-  esac
-}
-
 OUTDIR=${1:-cosmocc}
 APELINK=o/$(mode)/tool/build/apelink
 AMD64=${2:-x86_64}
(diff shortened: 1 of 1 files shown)


tool/cosmocc/package.sh
```diff
@@ -15,17 +15,49 @@ mode() {
...
+NPROC=$(($(_nproc)/2))
 GCCVER=14.1.0
+MAKE=$(which_make)
 
-make -j$NPROC m= \
+$MAKE -j$NPROC m= \
   $APELINK
 
-make -j$NPROC m=$AMD64 \
+case $(uname -s) in
```
jart, line 45: Could you check if it's on the path first?

```sh
if ! APE=$(command -v ape); then
  # do your thing
fi
```

tool/cosmocc/package.sh
```diff
@@ -15,17 +15,49 @@ mode() {
...
 }
 
+_nproc() {
+  case $(uname -s) in
+    Darwin) sysctl -n hw.logicalcpu ;;
+    *)      nproc                   ;;
+  esac
+}
+
+which_make() {
```
jart, line 25:
```sh
if ! MAKE=$(command -v gmake); then
  if ! MAKE=$(command -v make); then
    echo please install gnu make >&2
    exit 1
  fi
fi
```

On second thought, couldn't we just say

```sh
MAKE="$APE .cosmocc/current/bin/make"
```

jart (changes requested):

tool/cosmocc/package.sh
bjia56: I think I would prefer the first option, per reasoning below.

As things are built now, there's a chicken-and-egg problem when starting off of a fresh repo, such as building from a source release tarball. There is no guarantee of APE on the system, and a valid MAKE is required to launch the root Makefile to download the bootstrap cosmocc which has a valid APE and MAKE.

Calling download-cosmocc.sh from package.sh directly, before any MAKE invocations, could work, however as-is, there would be a duplication of specifying which bootstrap toolchain to use. We could update download-cosmocc.sh to be the authoritative source of which version to download, and modify the root Makefile to not provide any arguments, though I'm not sure how "generic" that script should be or if it's ok to impose a set of conventions. There would also need to be a way for the root Makefile to extract out of the script which version directory it is creating.

bjia56: Ok, I see there is a `build/bootstrap` directory which contains APE launchers. Could we add a blessed MAKE and ape-m1.c to that as well? In that case, the package.sh would first find/build APE (if one doesn't exist in PATH already), then use the MAKE inside `build/bootstrap`.

I see from git history that a MAKE was part of that folder at one point.

jart: I say just go with the first one. It's not too burdensome to ask for gmake if they're building a cosmocc toolchain.

bjia56: Got it, in that case I think I applied the requested changes and this is ready for another review.

jart (approved):

jart:
Thank you! Commit message edited to follow 72 column rule and use imperative tense in first line.

PR #1367, opened 2025-02-09
Check downloaded gcc/clang checksums

bjia56:
Check sha256 checksums of the downloaded gcc/clang toolchains. This allows us to extend trust to the external toolchains when building from source.


jart (approved):
Thanks! You could also probably use `.cosmocc/3.9.2/bin/sha256sum.ape` but this is fine.

jart:
Commit message modified to follow 72-column rule. See https://youtu.be/Y65FRxE7uMc for further details.

PR #1370, opened 2025-02-18
Fix: Potential Vulnerability in bzip2 Library

tabudz:
This PR fixes a security vulnerability in inflate() that was cloned from zlib but did not receive the security patch applied in zlib. The original issue was reported and fixed under https://gitlab.com/federicomenaquintero/bzip2/-/commit/74de1e2e6ffc9d51ef9824db71a8ffee5962cdbc.

This PR applies the same patch as the one in zlib to eliminate the vulnerability.

**References**
https://nvd.nist.gov/vuln/detail/CVE-2019-12900
https://gitlab.com/federicomenaquintero/bzip2/-/commit/74de1e2e6ffc9d51ef9824db71a8ffee5962cdbc

### DIFF
--- a/third_party/bzip2/README.cosmo
+++ b/third_party/bzip2/README.cosmo
@@ -9,3 +9,6 @@ ORIGIN
   Author: Mark Wielaard <mark@klomp.org>
   Date:   Sat Jul 13 17:17:58 2019 +0200
   Prepare for 1.0.8 release.
+
+local vulnerability fix:
+- Fix CVE-2019-12900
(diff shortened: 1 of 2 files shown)


third_party/bzip2/README.cosmo
```diff
@@ -9,3 +9,6 @@ ORIGIN
   Author: Mark Wielaard <mark@klomp.org>
   Date:   Sat Jul 13 17:17:58 2019 +0200
   Prepare for 1.0.8 release.
+
+local vulnerability fix:
+- Fix CVE-2019-12900
```
ket395, line 14: GH shows something is missing in this file.

nobkd: That's an End Of File (EOF) **without** newline marker, to be able to tell it apart from EOF with newline. Nothing's missing.

PR #1375, opened 2025-03-02
Add mkfifo()

danila-zol:
**STANDARDS**
[POSIX.1-2024](https://pubs.opengroup.org/onlinepubs/9799919799/)


danila-zol:
[Kakoune](https://github.com/mawww/kakoune) uses this function in a few places and wouldn't compile without this patch

A2va:
I just realized I need this function for one of my project. Did you send the email like it is mentioned in contributing ?
https://github.com/jart/cosmopolitan/blob/master/CONTRIBUTING.md#copyright-assignment

jart:
We've considered support mkfifo() before, but there's no way to make it portable to other OSes. Having it be defined at all when we can't support it as intended is problematic, because `./configure` scripts will usually check for mkfifo() and then produce programs that don't work on non-Linux OSes when it thinks this function is available.

PR #1381, opened 2025-03-06
Support additional architectures in apelink

bjia56:
This updates apelink to support machine architectures not in the source program input list by adding additional loaders, extracting the correct one that matches the host uname machine. With this change, blink can be supplied as the additional loader to run the program in x86_64 VMs. The change has been verified against blink 1.0, powerpc64le and mips64el in Docker using QEMU.


jart (approved):
Good idea. Thank you.

PR #1383, opened 2025-03-12
Add optional uname kernel check for non-ape loader

bjia56:
ELF executables of the same architecture lack mechanisms to specify the host operating system they are built for. As an example, SerenityOS and Linux both specify SYSV in the OSABI field, and while Linux specifies a .note.ABI-tag, SerenityOS lacks one altogether. Therefore, we do not do autodetection of the target kernel from provided loaders, and defer the matching of loaders to uname -s to the user.

### DIFF
--- a/tool/build/apelink.c
+++ b/tool/build/apelink.c
@@ -85,6 +85,13 @@
   "             executable will self-modify its header on\n"   \
   "             the first run, to use the platform format\n"   \
   "\n"                                                         \
+  "  -k KERNEL  test for maching kernel name [repeatable]\n"   \
+  "             when set, the shell script for subsequent\n"   \
+  "             loader executables will check if uname -s\n"   \
+  "             output matches the kernel string, only if\n"   \
+  "             the loader executable architecture is not\n"   \
+  "             an architecture in the input binary list\n"    \
+  "\n"                                                         \
   "  -M PATH    bundle ape loader source code file for m1\n"   \
   "             processors running the xnu kernel so that\n"   \
   "             it can be compiled on the fly by xcode\n"      \
@@ -213,6 +220,7 @@ struct Loader {
   char *ddarg_size1;
   char *ddarg_skip2;
   char *ddarg_size2;
+  char kernel[64];
 };
 
 struct Loaders {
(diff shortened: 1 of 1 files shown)


tool/build/apelink.c
```diff
@@ -244,6 +252,7 @@ static struct Inputs inputs;
 static char ape_heredoc[15];
 static enum Strategy strategy;
 static struct Loaders loaders;
+static char loader_kernel[64] = {'\0'};
```
jart, line 255: BSS memory is initialized to zero by default.

tool/build/apelink.c
```diff
@@ -979,13 +988,24 @@ static void AddLoader(const char *path) {
...
+  if (loader_kernel[0] != '\0') {
+    strncpy(loader->kernel, loader_kernel, sizeof(loader->kernel));
+  } else {
+    loader->kernel[0] = '\0';
+  }
+}
+
+static void SetLoaderKernel(const char *kernel) {
+  strncpy(loader_kernel, kernel, sizeof(loader_kernel));
+  loader_kernel[sizeof(loader_kernel) - 1] = 0;
```
jart, line 1002: Just use strlcpy(). In fact, technically speaking, it would be valid to have the static field be a `const char *` pointer, since the memory pointed to by `optarg` doesn't change. Then the code above could be updated to simply copy the pointer over. Then later, when you do your strcmp() it'll actually go very faster since the pointers will be equal in practice.

https://github.com/jart/cosmopolitan/blob/b235492e715df777bd14424dc1b7f9cd0605b379/libc/intrin/strcmp.c#L31-L36

tool/build/apelink.c
```diff
@@ -2206,6 +2240,20 @@ int main(int argc, char *argv[]) {
     }
 
     // extract the ape loader for non-input architectures
+    // if the user requested a host kernel check, get the host kernel
+    if (loader_kernel[0] != '\0') {
```
jart, line 2244: [optional; nit] My personal preference is to avoid redundancy and embrace the type messiness.

jart (changes requested):
Thanks for another change!

jart (approved):
Thank you!

jart:
If you look at blink's configure script, we determined years ago that `uname -s` isn't enough for certain systems like Cygwin, which need to be detected as `[ $(uname -o) = Cygwin ]`. This might be worth taking into consideration for a future change. Anyway I hope you don't mind me rewriting your commit message for clarity. I noticed you took the time and effort to align to column 71 but we've been pretty disciplined in recent months about always going for column 72.

PR #1384, opened 2025-03-13
Add NetBSD evbarm and fix segfault

bjia56:
This change fixes a segmentation fault when comparing loaders that don't have a target kernel set. Additionally, adds evbarm, which is the output of uname -m on NetBSD on aarch64.


bjia56:
I've elected to only set evbarm on the generic loader (blink) code path, since the APE loader doesn't run on NetBSD aarch64 for some reason. If support is added, then we can also move this condition up to the other script section.

Blink on NetBSD aarch64 actually fails on a mprotect call, which I am looking into. Strangely enough, it doesn't crash when run through truss, but crashes when run normally.

pkulchenko:
The change looks good to me, but I'd slightly tweak the first error message to distinguish it from the second one (to assist in troubleshooting). Maybe add "with no kernel set"?

jart (approved):

jart:
Thank you!

PR #1390, opened 2025-03-25
Fix shared_ptr<T>::owner_before

mrdomino:
`!(a < b)` is not the same as `b < a`.

I think I originally wrote it this way to avoid making weak_ptr a friend of shared_ptr, but weak_ptr already is a friend.


jart (approved):
Thanks for spotting this and sending the fix. Please merge when it's green.

PR #1400, opened 2025-04-17
Cache .cosmocc and o for github workflows

mrdomino:
Uses GitHub’s actions/cache@v4 to store the cosmocc distribution and the output directory between runs of the build workflow, with the version of cosmocc as the cache key.

Upgrades to actions/checkout@v4.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -45,7 +45,7 @@ jobs:
             o-
 
       - name: support ape bins 1
-        run: sudo cp build/bootstrap/ape.elf /usr/bin/ape
+        run: sudo cp -a build/bootstrap/ape.elf /usr/bin/ape
 
       - name: support ape bins 2
         run: sudo sh -c "echo ':APE:M::MZqFpD::/usr/bin/ape:' >/proc/sys/fs/binfmt_misc/register"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,5 +1,8 @@
 name: build
 
+env:
+  COSMOCC_VERSION: 3.9.2
+
 on:
   push:
     branches:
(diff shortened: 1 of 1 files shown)


mrdomino:
I’d be pretty fine with removing .cosmocc from the cache; it’s just downloading from cosmo.zip instead of GitHub.

jart (approved):
TY

mrdomino:
@jart PTAL

I'm not sure at this point why this is redoing work, but we're down from around 8 minutes to around 2 minutes.

My best guess as to why stuff is still getting recomputed is that the output cache is not having timestamps preserved, so anything that depends on things in the output cache gets recomputed.

jart (approved):
Two minutes? That's insane. I never could have imagined this would be possible.

jart:
It would potentially be fun to write a simple CLI program for caching mtime. I wonder how much time the current tool spends in it.

mrdomino:
The current tool claims to back it out of the git reflog, so mtime is commit time of last commit that changed file.

The Ruby tool I found apparently just built a big database of `(mtime, md5sum, path)` and applied the mtimes to the paths unless the md5 changed from what was in the database.

.github/workflows/build.yml
```diff
@@ -1,5 +1,8 @@
 name: build
 
+env:
+  COSMOCC_VERSION: 3.9.2
```
mrdomino, line 4: N.B. this creates another place that has to be updated when a new cosmocc dist is rolled.

Another option would be to have a `.cosmoversion` or something that has the version number in it and read that for the Makefile as well as this action.

.github/workflows/build.yml
```diff
@@ -19,10 +22,30 @@ jobs:
...
+      - uses: chetan/git-restore-mtime-action@v2
+
+      - uses: actions/cache@v4
+        with:
+          path: .cosmocc/${{ env.COSMOCC_VERSION }}
+          key: cosmocc-${{ env.COSMOCC_VERSION }}
+
+      - uses: actions/cache@v4
+        with:
+          path: o
```
mrdomino, line 40: It might be fun to see if `ln -sfn $(mktemp -d /dev/shm/o.XXXXXXXX) o` speeds up the linux builds further.

PR #1401, opened 2025-04-18
More shared_ptr fixes

mrdomino:
* Make refcount reads explicitly atomic
* Consistently put `const` in the same place
* Write the general `operator=` on `weak_ptr`


jart (approved):

jart:
Thank you! Merged.

PR #1404, opened 2025-04-22
Prioritize extra loader and OSNAME targets

bjia56:
This change modifies the order of non-input platform loaders and generic multi-OS loaders to prioritize non-input loaders, which allows us to add loaders which match APE-supported architectures (arm64/amd64) but aren't one of Linux/FreeBSD/NetBSD/OpenBSD. For example, this change enables us to add a Blink binary compiled for SunOS, which only supports amd64. The existing APE loaders passed as input will be matched last using existing script logic if no specific OSNAME target is set.


bjia56:
An alternative use case would be to embed a Blink loader for modern OpenBSD, which though emulated, should technically bypass the syscall pinning, though I haven't tried it.

954681:
AI

Issue #1409, opened 2025-05-01
Feature Request: Use different name for x86_64 and arm64 version of .ape-1.10 on Darwin


BenceSzalai:
### Prerequisites

- [x] I am running the latest code. Mention the version if possible as well.
- [x] I carefully followed the [README.md](https://github.com/jart/cosmopolitan/blob/master/README.md).
- [x] I searched using keywords relevant to my issue to make sure that I am creating a new issue that is not already open (or closed).
- [x] I reviewed the [Discussions](https://github.com/jart/cosmopolitan/discussions), and have a new and useful enhancement to share.

### Feature Description

When a cosmo app is started it creates a `/private/var/folders/bq/<long_string_of_characters_and_numbers>/T/.ape-1.10` file, which will be executed for the application runtime.

This file should be created using a name, that includes architecture, e.g.:
`/private/var/folders/bq/<long_string_of_characters_and_numbers>/T/.ape-1.10-x86_64`
`/private/var/folders/bq/<long_string_of_characters_and_numbers>/T/.ape-1.10-arm64`

### Motivation

I've encountered an issue which took quite some time to overcome, you can see the details in https://github.com/Mozilla-Ocho/llamafile/issues/750

Long story short, the problem is that on Darwin systems with Rosetta it is possible to launch a cosmo application both in ARM and Intel mode. Cosmo apparently decides which version of `.ape` it creates only on first launch, and reuses the already present version on any subsequent launches.

Which means that if someone kicks of a cosmo app in `x86_64` mode, they will end up with:
```
> file .ape-1.10
.ape-1.10: Mach-O 64-bit executable x86_64
```

Whereas if they start a cosmo app in ARM mode, they will end up with:
```
> file .ape-1.10
.ape-1.10: Mach-O 64-bit executable arm64
```

Problem arises when someone is trying to switch between Rosetta and native or if they use multiple cosmo apps, some with Rosetta and some in native ARM mode. Based on which mode they started first, all other cosmo app will also run either as `x86_64` or `arm`. This can give the very puzzling experience that even `arch -arm64 ./someapp` results in an `x86_64` process, or `arch -x86_64 ./someapp` results in an `arm64 process`.

I guess a naive solution could be to always rewrite `.ape` with the correct version, but that would still prevent someone to run cosmo apps concurrently with different architecture.

If the `.ape` file name included the architecture, this would be just fine.

PR #1410, opened 2025-05-01
Feature: Do not reuse different architecture version of temporary APE Loader on Darwin

BenceSzalai:
Addresses #1409 

Disclaimer: It has been more than a decade I last worked with C/C++ and I have no clue what am I doing, so please scrutinize everything I do and say.

So I thought I'll attempt to address the issue which caused me large headache few days ago.

As far as I can see the solution in this PR addresses the issue: Now on Apple Silicon Mac the APE loader is created as 
`${TMPDIR:-${HOME:-.}}/.ape-1.10-x86_64`
`${TMPDIR:-${HOME:-.}}/.ape-1.10-arm64`
`${TMPDIR:-${HOME:-.}}/.ape-1.10-arm64.c`

This way one can use Cosmo apps with both Rosetta and native ARM mode, even at the same time.

What I actually don't understand is why it appears that the same logic is implemented twice:
The code from [`ape/ape.S:613`](https://github.com/jart/cosmopolitan/blob/master/ape/ape.S#L613) and [`tool/build/apelink.c:2019`](https://github.com/jart/cosmopolitan/blob/master/tool/build/apelink.c#L2019) seems to do the same thing.

Ihave attempted to make sure they behave the same in my PR as well, but actually Idon't know what `ape/ape.S` does or when it is used, so I could not test if they indeed are fully compatible after my changes.

### DIFF
--- a/tool/build/pledge.c
+++ b/tool/build/pledge.c
@@ -555,10 +555,14 @@ void ApplyFilesystemPolicy(unsigned long ipromises) {
       if ((p = getenv("TMPDIR"))) {
         UnveilIfExists(
             __join_paths(buf, sizeof(buf), p, ".ape-" APE_VERSION_STR), "rx");
+        UnveilIfExists(
+            __join_paths(buf, sizeof(buf), p, ".ape-" APE_VERSION_STR GetXnuSuffix()), "rx");
       }
       if ((p = getenv("HOME"))) {
         UnveilIfExists(
             __join_paths(buf, sizeof(buf), p, ".ape-" APE_VERSION_STR), "rx");
+        UnveilIfExists(
+            __join_paths(buf, sizeof(buf), p, ".ape-" APE_VERSION_STR GetXnuSuffix()), "rx");
       }
     }
   }
--- a/ape/ape.S
+++ b/ape/ape.S
@@ -613,6 +613,11 @@ apesh:	.ascii	"\n@\n#'\"\n"			// sixth edition shebang
 	.ascii	    "t=\"${TMPDIR:-${HOME:-.}}/.ape-"
 	.ascii	     APE_VERSION_STR
 	.ascii	     "\"\n"
+#if SupportsXnu()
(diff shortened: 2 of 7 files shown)


pkulchenko:
Bence, thank you for your contribution! It's an interesting approach that I think may work (although I'm also not sure if it breaks anything for others). As a first time contributor, could you check instructions in CONTRIBUTING.md?

BenceSzalai:
Well, sure, I've started, and then got distracted after the send email part. Will move on now.

tool/build/pledge.c
```diff
@@ -555,10 +555,16 @@ void ApplyFilesystemPolicy(unsigned long ipromises) {
       if ((p = getenv("TMPDIR"))) {
         UnveilIfExists(
             __join_paths(buf, sizeof(buf), p, ".ape-" APE_VERSION_STR), "rx");
+        UnveilIfExists(__join_paths(buf, sizeof(buf), p,
```
BenceSzalai, line 558: Probably these could be wrapped in a condition as well to only run on xnu, but I don't know what / if there is an appropriate check available in this context. Same for bellow.

tool/build/pledge.c
```diff
@@ -555,10 +555,16 @@ void ApplyFilesystemPolicy(unsigned long ipromises) {
...
+                                    ".ape-" APE_VERSION_STR GetXnuSuffix()),
+                       "rx");
       }
       if ((p = getenv("HOME"))) {
         UnveilIfExists(
             __join_paths(buf, sizeof(buf), p, ".ape-" APE_VERSION_STR), "rx");
+        UnveilIfExists(__join_paths(buf, sizeof(buf), p,
+                                    ".ape-" APE_VERSION_STR GetXnuSuffix()),
+                       "rx");
       }
```
BenceSzalai, line 568: Probably these could be wrapped in a condition as well to only run on xnu, but I don't know what / if there is an appropriate check available in this context. Same for bellow.

PR #1415, opened 2025-05-11
Ip6 getifaddrs

izissise:
Github discussion: https://github.com/jart/cosmopolitan/discussions/1412

This is Linux only, I don't known how to implement this for other platforms, any help is welcome.

Reading the `/proc/net/if_inet6` pseudo-file allows parsing IPv6 addresses,
which are now included in the list returned by `getifaddrs()`.

This patch:
- Adds support for AF_INET6 entries by reading from the procfs
- Parses scope, prefix length, and flags from the if_inet6 format
- Constructs `sockaddr_in6` and attaches them to `ifaddrs` structs
- Maintains compatibility with existing AF_INET handling
- Adds `IfAddr6` and supporting structures to represent IPv6 info

### DIFF
--- a/libc/sock/ifaddrs.c
+++ b/libc/sock/ifaddrs.c
@@ -18,13 +18,19 @@
 ╚─────────────────────────────────────────────────────────────────────────────*/
 #include "libc/sock/ifaddrs.h"
 #include "libc/calls/calls.h"
+#include "libc/calls/syscall-sysv.internal.h"
+#include "libc/dce.h"
+#include "libc/limits.h"
 #include "libc/mem/mem.h"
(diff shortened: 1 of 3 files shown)


jart (changes requested):
Looks good. Could you sync to head for a cachestat CI fix so we can verify tests work? Also could you send the email mentioned in CONTRIBUTING.md? Thanks!

izissise:
Thanks for considering this change, I already did the CCA in 2022 see https://github.com/jart/cosmopolitan/pull/641#pullrequestreview-1122882959

jart (approved):
Right you are. Welcome back :) LGTM

PR #1421, opened 2025-05-19
Better mtimes for github workflow build cache

mrdomino:
Saves and restores mtimes to a file, also covering the `o/` directory to hopefully preserve make dependency information better.


jart (approved):
Looks good. Thanks for doing this!

Issue #913, opened 2023-10-16
ls


Lattay:

mrdomino:
cmder.app looks to be based on something called [ConEmu](https://github.com/Maximus5/ConEmu). My first guess is that the explanation is going to be somewhere in here:

https://github.com/Maximus5/ConEmu/blob/master/src/ConEmu/CreateProcess.cpp

But I'm not even really all that sure of that; it's a huge project and I'm not a Windows guy.

If some dedicated cmder / conemu user wants to dig in, that would probably be where to start looking...

N.B. here's ours:

https://github.com/jart/cosmopolitan/blob/master/libc/proc/execve-nt.greg.c

wdscxsj:
It should work fine if "Inject ConEmuHk.dll into processes started in ConEmu tabs" is disabled in ConEmu.

PR #1431, opened 2025-07-19
Fix two problems regarding fork() and async IO for Windows

hsfzxjy:
**1. `DeduplicateStdioHandles()` should mark the new handles inheritable**

[sys_fork_nt_parent()](https://github.com/hsfzxjy/cosmopolitan/blob/e7e5856b8312c9d34de67b40cd6b8522ae33ced9/libc/proc/fork-nt.c#L191) is directly passing std handles to child process. It would be a problem if those handles are non-inheritable.

**2. Async IO on Windows might be synchronous**

Even if the `OVERLAPPED` parameter supplied, IO functions like `ReadFile`/`WriteFile` may still complete their jobs before they return. In that case, the `numberOfBytes` result from the IO functions should be respected, and `GetOverlappedResult` should not be called.

[WriteFile](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-writefile#synchronization-and-file-position)
[Asynchronous disk I/O appears as synchronous on Windows](https://learn.microsoft.com/en-us/previous-versions/troubleshoot/windows/win32/asynchronous-disk-io-synchronous)

### DIFF
--- a/libc/calls/ntspawn.c
+++ b/libc/calls/ntspawn.c
@@ -68,7 +68,7 @@ static textwindows ssize_t ntspawn_read(intptr_t fh, char *buf, size_t len) {
   uint32_t got;
   struct NtOverlapped overlap = {.hEvent = CreateEvent(0, 0, 0, 0)};
   ok = overlap.hEvent &&
-       (ReadFile(fh, buf, len, 0, &overlap) ||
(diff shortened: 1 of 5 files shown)


hsfzxjy:
p.s. This PR fixes https://github.com/jart/cosmopolitan/issues/913

PR #1434, opened 2025-08-10
Make tools/getifaddrs output closer to ifconfig

izissise:
- Add bswap_128()
- Make tools/getifaddrs output closer to ifconfig
This is easier to read with a lots of addresses and give more infos


izissise:
On Windows the tool is called `ipconfig` and look a bit different, wikipedia has a screeshot https://en.wikipedia.org/wiki/Ipconfig

PR #1436, opened 2025-08-13
Add IPv6 support to getifaddrs() on Bsd

izissise:
- Add bswap_128()
- Add IPv6 support to getifaddrs() on Bsd
Tested on x86_64 MacOs OpenBsd FreeBsd and NetBsd.

There is extra ioctl calls on the ipv6 bsd path
because SIOCGIFCONF doesn't returns the addr flag
or the netmask.

Luckily all bsd systems share the same values for
SIOCGIFAFLAG_IN6 and SIOCGIFNETMASK_IN6.

### DIFF
--- a/libc/calls/ioctl.c
+++ b/libc/calls/ioctl.c
@@ -44,6 +44,7 @@
 #include "libc/runtime/runtime.h"
 #include "libc/runtime/stack.h"
 #include "libc/serialize.h"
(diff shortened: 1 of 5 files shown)


jart (changes requested):
Wow this is great. I wish Linux had this! Thank you for doing this.

izissise:
I made the requested change

PR #1442, opened 2025-09-12
update onError in definition.lua

solisoft:
The onError definition was wrong in the definition.lua file. The details params was missing.


jart (approved):
Thanks!

PR #1444, opened 2025-09-26
Fix bare metal boot loop

randomtwdude:
This PR fixes the boot looping on bare metal.
- One of the init stubs `_init_metalfile` calls `memmove()` before the jump table is set up
- The global variable `__pid` is never set on metal. Causing lock acquisitions by `mmap()` to always fail.


jart (approved):
Awesome thank you for figuring this out. This will be helpful when the release goes out. Could you send the email mentioned in https://github.com/jart/cosmopolitan/blob/master/CONTRIBUTING.md#copyright-assignment so I can merge this? Thanks!

PR #1500, opened 2026-05-25
various pledge improvements

amrfti:
I ran into several issues trying to pledge [monerod](https://github.com/monero-project/monero).

This MR contains the patches I had to write in order to get it to work, which should be generally useful.

### DIFF
--- a/libc/calls/pledge-linux.c
+++ b/libc/calls/pledge-linux.c
@@ -1543,6 +1543,7 @@ __privileged static void AllowIoctlTty(struct Filter *f) {
 //   - SO_SNDTIMEO          (0x15)
 //   - IP_RECVTTL           (0x0c)
 //   - IP_RECVERR           (0x0b)
+//   - IP_MTU_DISCOVER      (0x0e)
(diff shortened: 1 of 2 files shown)


couleurm:
unrelated- could you share the pledge arguments you ended up using to wrap monerod?

amrfti:
> could you share the pledge arguments you ended up using to wrap monerod?
```
exec pledge \
	-M 4T \
	-F 1T \
	-O 4096 \
	-v /etc/monero \
	-v /data/monerod \
	-v /var/log/monero \
	-p 'stdio rpath wpath cpath flock proc prot_exec inet dns' \
	/usr/bin/monerod "$@"
```
If you're running it in the foreground you probably want the `tty` pledge--else the console will look weird--I log to a file so I don't need it. Limits are chosen mostly arbitrarily, monerod is massive so it's kinda pointless to limit the memory/file size limits. If pledge had the functionality I would probably just disable the limits.

couleurm:
looking at the blog post it seems you can disable the limit by providing a negative value

PR #1507, opened 2026-07-15
fix: cleanup remnants of kErrnoNames

G4Vi:
https://github.com/jart/cosmopolitan/commit/5ddb5c2adad79407fb800fce47f389611f90a511 removed kErrnoNames , but left it in the header.

### DIFF
--- a/test/libc/fmt/strerror_r_test.c
+++ b/test/libc/fmt/strerror_r_test.c
@@ -23,8 +23,7 @@
 
 /*
  * If these tests break, it's probably because
- * libc/sysv/consts.sh changed and
- * libc/sysv/kErrnoNames.S needs updating.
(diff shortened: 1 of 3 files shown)


test/libc/fmt/strerror_r_test.c
```diff
@@ -24,7 +24,6 @@
 /*
  * If these tests break, it's probably because
  * libc/sysv/consts.sh changed and
- * libc/sysv/kErrnoNames.S needs updating.
```
ahgamut, line 27: minimal grammar fix in the comment? perhaps we can just close the line at `libc/sysv/consts/ changed.`

G4Vi: Good catch, updated.

PR #1509, opened 2026-07-19
feat(build): automate Fat APE generation to resolve x86_64/arm64 fric…

### LINKED ISSUE #1493: error: this ape binary only supports x86_64
I've built cosmopolitan on Linux x86_64 and copy redbean to an osx arm64 and I'm getting this message:
```
error: this ape binary only supports x86_64
```
If I understood correctly it should work without any further processing but it doesn't, if I download redbean from https://redbean.dev/redbean-demo-3.0.0.com it does work.

What I'm missing ?

DjibyDiop:
Fixes #1493 

The current workflow for building native-execution APE binaries across both Apple Silicon (aarch64) and x86_64 requires manual dual-compilation and sequential `apelink` merging. As noted in #1493, this creates friction and undocumented edge cases for users expecting out-of-the-box portability.

This PR introduces a dedicated Makefile target (`fat-redbean`) to fully automate this pipeline. 

Highlights:
* Compiles both `default` (x86_64) and `m=aarch64` targets sequentially.
* Ensures `o//tool/build/apelink` is built before merging.
* Outputs a unified Fat APE binary ready for native multi-arch execution.
* Pure Makefile logic, zero external shell scripts added.

Tested locally on my WSL environment. Let me know if you'd prefer this integrated differently within the `BUILD.mk` hierarchy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,6 +122,16 @@ all the unit tests in the `TEST_POSIX` package, you could say:
 .cosmocc/current/bin/make o//test/posix
 ```
 
+### Fat APE Binaries
+
+To compile a native-execution Fat APE binary across both Apple Silicon (aarch64) and x86_64, use the provided helper script. For example, to build a fat binary for redbean:
+
+```sh
+build/fat.sh tool/net/redbean
+```
+
+This will sequentially build the x86_64 and aarch64 targets and merge them using apelink into a unified `o/fat/tool/net/redbean.com` binary ready for multi-arch execution.
+
 Cosmopolitan provides a variety of build modes. For example, if you want
 really tiny binaries (as small as 12kb in size) then you'd say:
 
--- a/build/fat.sh
+++ b/build/fat.sh
@@ -1,0 +1,70 @@
+#!/bin/sh
+# cosmopolitan fat ape builder
+#
+# Builds a fat APE binary containing both x86_64 and aarch64 native code.
+#
+# Usage:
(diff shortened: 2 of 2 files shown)


ahgamut:
This PR is a nice idea, but I feel a shell script is a simpler answer here (similar to `tool/cosmocc/package.sh`), or maybe just an extended explanation in the documentation. So we can call fat binary builds explicitly and outside of the makefile dependency structure.

IIUC the makefile target added in this PR will build incorrectly if we're building on a host `linux-aarch64` system, because `o//` and `o/aarch64` will both be building aarch64 binaries.

For now I would also prefer to add explicit targets in the makefile if possible, and avoid having the makefile call itself.

DjibyDiop:
Thanks for the detailed review, @ahgamut!

You are completely right about the linux-aarch64 host edge case—I missed that o// would default to the host architecture in that scenario, breaking the multi-arch logic. I also totally agree on avoiding recursive $(MAKE) to keep the dependency graph clean.

Would you prefer I update this PR to introduce a standalone shell script (something like tool/build/make-fat-ape.sh similar to package.sh), or should we just close this one and I'll submit a small PR to update the documentation with the explicit steps instead?

Happy to pivot to whichever approach keeps the repo cleaner!

ahgamut:
Let's close this one, and then have a PR just to update the documentation, showing what the shell script would look like. We can add a shell script later if it's necessary.

PR #1519, opened 2026-08-28
fix: include __program_executable_name in stack mapping even if argc == 0

### LINKED ISSUE #1520: __program_executable_name is not always mapped when argc == 0
When running an APE with the APE loader without any arguments, for example:  `./ape.elf - /path/to/ape-program`, the program executable path is not always included in cosmopolitan's memory maps (`kisdangerous(__program_executable_name)` sometimes returns `true`). This causes the path returned by `GetProgramExecutableName()` to be un `open`able, breaking zipos.

The culprit is:

https://github.com/jart/cosmopolitan/blob/5ddb5c2adad79407fb800fce47f389611f90a511/libc/intrin/getmainstack.c#L76-L88

This function determines the top of stack for the mappings. When `argc > 0`;  `./ape.elf - /path/to/ape-program argv0`, as `argv0` is placed at a higher address than `/path/to/ape-program`,  `/path/to/ape-program` is included in the stack. When `argc == 0`, the fallback case uses the end of `auxv` as the top of the stack. When `/path/to/ape-program` occurs on a page at a higher address than `auxv`, it is excluded from the stack mapping.

G4Vi:
Fixes #1520

The fix is to fallback to using the end of `__program_executable_name` as the top of the stack when `argv` is empty before fallback to end of auxv if that also doesn't exist.

Includes two tests, one checking `!kisdangerous(__program_executable_name)` when launching with ape loader without arguments, and one checking zipos works when an APE is launched without arguments.

@mrdomino it would be greatly appreciated if you could review this given your previous work with `GetProgramExecutableName` and the APE loader.

### DIFF
--- a/libc/intrin/getmainstack.c
+++ b/libc/intrin/getmainstack.c
@@ -76,7 +76,8 @@ static int __get_length(const char *s) {
 static uintptr_t __get_main_top(int pagesz) {
   uintptr_t top;
   const char *s;
-  if ((s = __get_last(__envp)) || (s = __get_last(__argv))) {
+  if ((s = __get_last(__envp)) || (s = __get_last(__argv)) ||
+      (s = __program_executable_name)) {
     top = (uintptr_t)s + __get_length(s);
   } else {
     unsigned long *xp = __auxv;
--- a/libc/intrin/getmainstack.c
+++ b/libc/intrin/getmainstack.c
@@ -74,16 +74,19 @@ static int __get_length(const char *s) {
 }
 
 static uintptr_t __get_main_top(int pagesz) {
-  uintptr_t top;
+  uintptr_t top, alttop;
   const char *s;
-  if ((s = __get_last(__envp)) || (s = __get_last(__argv)) ||
-      (s = __program_executable_name)) {
+  if ((s = __get_last(__envp)) || (s = __get_last(__argv))) {
     top = (uintptr_t)s + __get_length(s);
   } else {
     unsigned long *xp = __auxv;
     while (*xp)
       xp += 2;
     top = (uintptr_t)xp;
+  }
+  if ((s = __program_executable_name)) {
+    alttop = (uintptr_t)s + __get_length(s);
+    top = MAX(top, alttop);
   }
   return ROUNDUP(top, pagesz);
 }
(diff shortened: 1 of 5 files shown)


mrdomino:
It's been a long time since I touched this code and I don't have time right now to load all the state back in, so this is not a high-confidence review.

My first question is, why is it safe to assume `__program_executable_name` is the top of the stack? I don't know if the variable is consistently related to `sp` in the APE loaders, and if it isn't, then it seems like this could be marking a lot of memory as safe that isn't.

G4Vi:
Thanks for taking a look!

The executable path passed from the ape loader can be in one of two places on the stack. If it's an absolute path, the value passed in to the ape loader is used directly, see `Commandv` in `ape/loader.c`.  (`argv[2]` passed in to the ape loader when executed like `./ape.elf - /path/to/executable`). If it's not an absolute path, `SearchPath` sets it to part of the `ApeLoader` struct `M` which is `alloca`'d before the `argv,env,auxv` arrays so it has a higher address than the end of `auxv`. 

On `x86_64` the executable path is passed in `%rdx` to `ape/launch.S` which does not tamper with it. In `crt.S`, `%rdx` is copied to `__program_executable_name`. It's important to note that `__program_executable_name` is no longer guaranteed to either point to the original stack or NULL once  `__init_program_executable_name` is called (on all archs). `libc/intrin/maps_init.S` has `.init.start 301,_init_maps` and `libc/calls/program_executable_name.S` has `.init.start 305,_init_program_executable_name` so the maps are fortunately initialized first.

On `aarch64`, the executable path is still passed in the third param, in `cosmo` in `libc/runtime/cosmo2.c`,  `__program_executable_name` is set to the third param and `__maps_init` is also called before `__init_program_executable_name`.

mrdomino:
I think this is ok then. Might be good to include a bounds check - reject the pointer unless it's at or above `auxv`, for instance.

G4Vi:
> Might be good to include a bounds check - reject the pointer unless it's at or above auxv, for instance.

Changed in a57238e to be more defensive against ape loader changes, now only uses end of `__program_executable_name` if it exists and is greater than the top calculated from the previous methods.

